PR #26, opened 2025-07-10
feat(config): unify CORS and Vite allowedHosts via ALLOWED_ORIGINS 

LucasRoesler:
Support wildcard hosts in CORS and in the local dev server. 
This is configured via the same ALLOWED_ORIGINS env variable, for simplicity. The allows users to server the local app via NGROK, Tailscale, or other similar tools.


LucasRoesler:
sorry, this was opened prematurely

PR #35, opened 2025-07-11
feat: Add thinking mode selector to chat interface

lvalics:
- Added ThinkingModeSelector component with 5 thinking modes
- Integrated selector into chat header next to permission mode
- Automatically prefixes messages with thinking instructions
- Resets to default mode after sending message

### DIFF
--- a/src/components/ChatInterface.jsx
+++ b/src/components/ChatInterface.jsx
@@ -24,6 +24,7 @@ import ClaudeLogo from './ClaudeLogo.jsx';
 import ClaudeStatus from './ClaudeStatus';
 import { MicButton } from './MicButton.jsx';
 import { api } from '../utils/api';
+import ThinkingModeSelector, { thinkingModes } from './ThinkingModeSelector.jsx';
 
 // Memoized message component to prevent unnecessary re-renders
 const MessageComponent = memo(({ message, index, prevMessage, createDiff, onFileOpen, onShowSettings, autoExpandTools, showRawParameters }) => {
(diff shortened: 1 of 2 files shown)


viper151:
@lvalics thinking mode does not exist in Claude Code. 

Moreover the commit only adds a button in the Frontend

lvalics:
It is exist ... Try to add this words in your promot, actually it is add ONLY to the prompt. Not need any backend.
Ask Claude to make a plan for how to approach a specific problem. We recommend using the word "think" to trigger extended thinking mode, which gives Claude additional computation time to evaluate alternatives more thoroughly. These specific phrases are mapped directly to increasing levels of thinking budget in the system: "think" < "think hard" < "think harder" < "ultrathink." Each level allocates progressively more thinking budget for Claude to use.
https://www.anthropic.com/engineering/claude-code-best-practices

lvalics:
I am using a lot to force Claude Code to get implementation more thinking, overthinking, but consume more credits (if you are on token) but is not consume if you are on MAX or PRO. I am using Claude Code quite a lot, from the first version, it is used by engineers a lot.

lvalics:
image
Just to see how is react.

viper151:
I know it exists but as you mentioned it's only when you mention it within the prompt itself. 
What I missed in your code was this 
    if (selectedThinkingMode && selectedThinkingMode.prefix) {
      messageContent = `${selectedThinkingMode.prefix}: ${input}`;
    }
    
I like it and want to merge, but doesn't it clutter a bit too much the interface? We nw also have the planning mode "pill" above the chat interface input. Think of the mobile interface.

lvalics:
You can add it and see what people think an of course you can move else later if needed. Normally near that button is ok I think.

zuodajiang-1m:
It is a feature in Claudia

zuodajiang-1m:
image

image

It also has a mode selection function

blackmammoth:
Hey @lvalics can you resolve the merge conflicts?

blackmammoth:
Superceded by issue #301

PR #37, opened 2025-07-11
feat: Add file metadata display with view modes

lvalics:
- Added file size, permissions (rwx format), and modified date display
- Implemented three view modes: simple, compact, and detailed
- Added server-side file stats collection in getFileTree
- View mode preference persisted in localStorage
- Detailed view shows table-like layout with column headers
- Compact view shows inline metadata
- Simple view maintains original basic tree structure

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -820,6 +820,14 @@ app.get('*', (req, res) => {
   res.sendFile(path.join(__dirname, '../dist/index.html'));
 });
 
+// Helper function to convert permissions to rwx format
+function permToRwx(perm) {
+  const r = perm & 4 ? 'r' : '-';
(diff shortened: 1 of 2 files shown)


viper151:
@lvalics see my comment above

lvalics:
hi, no comment left. but I see you merged.

viper151:
> hi, no comment left. but I see you merged.

No it's not yet merged as I have a comment on the file permissions. I'm not sure why that is needed

lvalics:
Any developer will be interested if the file are readable, need to have 644 or 600 etc. Also filesize can help. If you have a filelisting, it is better to have details as well.

PR #62, opened 2025-07-14
feat: add ctrl+enter send option & fix IME problen

### LINKED ISSUE #58: Enter key during IME conversion immediately sends message in East Asian languages
When composing messages in East Asian languages like Japanese and Chinese, users rely on Input Method Editors (IME) that require pressing Enter to confirm character conversions. Currently, this Enter keypress immediately submits the message instead of just confirming the conversion, interrupting the writing process before users can complete their intended text. This creates a significant usability barrier for users of these languages.

**Solution**: The input system should detect when Enter is pressed during active IME composition and treat it as conversion confirmation rather than message submission. Only when no IME conversion is active should Enter trigger message sending. Alternatively, please add an option to require a combination like Shift+Enter or ⌘+Enter to submit the message.

https://lingoport.com/i18n-term/cjk-languages-ime/

Image

sarashinanikki:
## Summary

- This PR resolve #58 
- Implement IME composition detection to prevent message sending during East Asian language input conversion
- Add "Send by Ctrl+Enter" option for users who prefer explicit send shortcuts
- Update help text dynamically based on input mode

### IME Composition Handling
- Added IME composition detection using onCompositionStart/onCompositionEnd events
- IME composition always prevents Enter key from sending messages, regardless of user settings
- Ensures East Asian language input conversion works properly without accidental sends

### Send by Ctrl+Enter Option
- Added toggle in Quick Settings Panel for users who prefer Ctrl+Enter sending
- When enabled, plain Enter creates new lines instead of sending messages
- When disabled, plain Enter sends messages (except during IME composition)
- Ctrl+Enter always sends messages in both modes

### Dynamic Help Text
- Updated hint text to reflect current input mode:
  - Default mode: "Press Enter to send"
  - Ctrl+Enter mode: "Ctrl+Enter to send (IME safe)"
- Both desktop and mobile versions updated with appropriate messaging

### Technical Implementation
- State management through App.jsx → MainContent.jsx → ChatInterface.jsx props chain
- QuickSettingsPanel integration with consistent styling and UX
- Proper event handling that respects both IME composition and user preference

## Test Plan

- [x] Test Japanese/Chinese/Korean IME input
- [x] Verify Ctrl+Enter sends messages in both modes
- [x] Verify Shift+Enter creates new lines in both modes
- [x] Test setting persistence across browser sessions
- [x] Verify help text updates correctly when toggling the option

### DIFF
--- a/src/components/ChatInterface.jsx
+++ b/src/components/ChatInterface.jsx
@@ -1016,7 +1016,7 @@ const ImageAttachment = ({ file, onRemove, uploadProgress, error }) => {
 // - onReplaceTemporarySession: Called to replace temporary session ID with real WebSocket session ID
 //
 // This ensures uninterrupted chat experience by pausing sidebar refreshes during conversations.
-function ChatInterface({ selectedProject, selectedSession, ws, sendMessage, messages, onFileOpen, onInputFocusChange, onSessionActive, onSessionInactive, onReplaceTemporarySession, onNavigateToSession, onShowSettings, autoExpandTools, showRawParameters, autoScrollToBottom }) {
+function ChatInterface({ selectedProject, selectedSession, ws, sendMessage, messages, onFileOpen, onInputFocusChange, onSessionActive, onSessionInactive, onReplaceTemporarySession, onNavigateToSession, onShowSettings, autoExpandTools, showRawParameters, autoScrollToBottom, sendByCtrlEnter }) {
   const [input, setInput] = useState(() => {
     if (typeof window !== 'undefined' && selectedProject) {
       return localStorage.getItem(`draft_input_${selectedProject.name}`) || '';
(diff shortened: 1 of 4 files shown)


src/components/ChatInterface.jsx
```diff
@@ -1998,14 +1999,21 @@ function ChatInterface({ selectedProject, selectedSession, ws, sendMessage, mess
     
     // Handle Enter key: Ctrl+Enter (Cmd+Enter on Mac) sends, Shift+Enter creates new line
     if (e.key === 'Enter') {
+      // If we're in composition, don't send message
+      if (isComposing) {
```
odiak: Just checking `e.nativeEvent.isComposing` will be enough.
The `isComposing` state is not necessary.

sarashin65: Oh, you are right. I fixed and It works correctly. Thanks!

PR #63, opened 2025-07-14
feat: Add Tunnelmole integration for remote access

dario-valles:
## Summary
- Adds Tunnelmole integration to expose local Claude Code UI instances to the internet
- Provides easy mobile access through QR codes
- Seamless toggle in the settings panel

## Motivation
Users often need to access their Claude Code UI instance from mobile devices or external networks. This feature provides a secure and easy way to create temporary tunnels without complex network configuration.

## Changes
- **Backend**: 
  - Added Tunnelmole service manager (`server/services/tunnelService.js`)
  - Created API endpoints for tunnel control (`server/routes/tunnel.js`)
  - Updated server to support tunneling in development mode
  
- **Frontend**:
  - Added Network Access section to QuickSettingsPanel
  - Implemented tunnel toggle with real-time status updates
  - Added QR code generation for easy mobile scanning
  - Updated WebSocket connection to use tunnel URLs when active

- **Configuration**:
  - Added `tunnelmole` and `qrcode` npm dependencies
  - Updated Vite config to allow external host access
  - Fixed port configuration for proper development setup

## Screenshots
image

## Testing
1. Run `npm install` to install new dependencies
2. Start the dev server with `npm run dev`
3. Open Quick Settings panel (right sidebar)
4. Toggle "Enable Tunnelmole" in Network Access section
5. Use the displayed URL or scan the QR code to access from external devices

## Security Considerations
- Tunnel URLs are temporary and only active while enabled
- Vite's `allowedHosts: true` is only set in development mode
- All tunnel endpoints are protected by authentication

### DIFF
--- a/server/routes/git.js
+++ b/server/routes/git.js
@@ -444,10 +444,25 @@ router.get('/remote-status', async (req, res) => {
       trackingBranch = stdout.trim();
       remoteName = trackingBranch.split('/')[0]; // Extract remote name (e.g., "origin/main" -> "origin")
     } catch (error) {
-      // No upstream branch configured
+      // No upstream branch configured - but check if we have remotes
+      let hasRemote = false;
+      let remoteName = null;
+      try {
+        const { stdout } = await execAsync('git remote', { cwd: projectPath });
+        const remotes = stdout.trim().split('\n').filter(r => r.trim());
+        if (remotes.length > 0) {
+          hasRemote = true;
+          remoteName = remotes.includes('origin') ? 'origin' : remotes[0];
+        }
+      } catch (remoteError) {
+        // No remotes configured
+      }
+      
       return res.json({ 
-        hasRemote: false, 
+        hasRemote,
+        hasUpstream: false,
         branch,
+        remoteName,
         message: 'No remote tracking branch configured'
       });
     }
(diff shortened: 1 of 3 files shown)


viper151:
@dario-valles I like the idea but why not use Localtunnel instead ? 

https://theboroer.github.io/localtunnel-www/

dario-valles:
"Good question! I wasn't aware of Localtunnel before. Both are indeed open source, which is great. I chose Tunnelmole because:

It offers persistent custom domains (paid tier) so you can keep the same URL across sessions
Has multiple installation options (npm, binary, or curl) vs Localtunnel's npm-only
More actively maintained with recent updates
Can be fully self-hosted (both client and server)

That said, Localtunnel is completely free and simpler if you just need basic tunneling. I'm not affiliated with either service - just picked what worked best for my needs. 
If you want i could change it to use localtunnel

zuodajiang-1m:
There’s also Cloudflare Tunnel, but there are too many options. I think you can run this project in Docker, expose the port, and host it on a server with a public IP.

In a pure environment, tunnels may not be necessary.

blackmammoth:
Hey @dario-valles, this idea is pretty cool! However, there are some errors I found in devTools while testing the PR. In fact, although the URL given by tunnelmole works, I just see a blackscreen. I believe this is because the websocket connections are being rejected. 

1. 
image


2. 
image


In addition, there are some merge conflicts since there was a big refactor.

PR #133, opened 2025-07-31
Fix command escaping in claude-cli.js

WolCarlos:
## Bug Fix

Fixed an issue where commands with special characters were not properly escaped when passed to the Claude CLI.

### Changes
- Added proper quote escaping for commands in `server/claude-cli.js`
- Commands are now wrapped in quotes and internal quotes are escaped with backslashes

### Before
```javascript
args.push('--print', command);
```
### After
```javascript
args.push('--print', `"${command.replace(/"/g, '\\"')}"`);
```


viper151:
@WolCarlos  thanks for this quite useful :)

PR #136, opened 2025-08-05
feat: Add intelligent Claude CLI version detection and management

pbastos:
## Problem

Users often have multiple Claude CLI installations (e.g., from npm global install and Claude's self-updater), causing the app to potentially use outdated versions and fail with cryptic errors.

### Real-world Example
- Old version (v1.0.17) at `/usr/local/bin/claude` from npm
- New version (v1.0.69) at `~/.claude/local/claude` from `claude update`
- Without this fix: App might use the old version depending on PATH order

## Solution

This PR adds intelligent Claude CLI detection that:

### 🔍 Auto-Detection
- Scans multiple common installation locations
- Checks PATH directories
- Detects Claude's self-update location (`~/.claude/local/`)

### ✅ Version Validation
- Gets version of each installation found
- Validates against minimum requirement (v1.0.24+)
- Automatically selects the newest valid version

### 📝 Enhanced Error Messages
Instead of cryptic spawn errors, users now see:
```
⚠️ Found Claude CLI version 1.0.17 at /usr/local/bin/claude, 
but version 1.0.24 or higher is required. Please update Claude CLI.

Using: ~/.claude/local/claude (v1.0.69)
```

### ⚙️ Configuration Support
Users can override with custom path via `.claudeui.json`:
```json
{
  "claudeBinaryPath": "/path/to/specific/claude"
}
```

## Changes

- Added `server/utils/claude-detector.js` - Core detection logic
- Updated `server/claude-cli.js` - Uses detector before spawning
- Updated `server/routes/mcp.js` - MCP routes use detector
- Updated `src/components/ChatInterface.jsx` - Handles warning messages
- Added `.claudeui.json.example` - Configuration example
- Added documentation and test script

## Testing

Tested with multiple Claude installations:
- ✅ Correctly identifies all installations
- ✅ Selects newest valid version (1.0.69 over 1.0.17)
- ✅ Shows clear warnings for outdated versions
- ✅ Handles missing Claude gracefully
- ✅ Configuration override works

## Impact

- **Reliability**: Always uses the best available Claude version
- **Transparency**: Shows which Claude is being used
- **User-friendly**: Clear messages guide users to fix issues
- **Future-proof**: Handles Claude's self-update mechanism

This is especially important because Claude's own `claude update` command installs to a different location than npm, naturally creating multiple installations.

Fixes issues where users get mysterious failures after updating Claude.

### DIFF
--- a/.claudeui.json.example
+++ b/.claudeui.json.example
@@ -1,0 +1,4 @@
+{
+  "claudeBinaryPath": null,
+  "__comment": "Set claudeBinaryPath to use a specific Claude CLI installation, e.g., '/home/user/.claude/local/claude'"
+}
--- a/CLAUDE_VERSION_DETECTION.md
+++ b/CLAUDE_VERSION_DETECTION.md
@@ -1,0 +1,87 @@
+# Claude CLI Version Detection
+
+This feature provides intelligent Claude CLI version detection and management to handle scenarios where users have multiple Claude CLI installations.
+
+## Problem Solved
+
+When users have multiple Claude CLI installations (e.g., one from npm global install and another from `claude update`), the application might use the wrong version, leading to compatibility issues.
+
+### Example Scenario
+- User has Claude v1.0.17 installed at `/usr/local/bin/claude` (from npm)
+- User has Claude v1.0.69 installed at `~/.claude/local/claude` (from claude update)
+- Without this feature, the app might use the old v1.0.17 due to PATH resolution
+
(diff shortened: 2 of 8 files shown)


viper151:
@pbastos  thank you for contributing. Although this makes sense, this is simply a hygiene issue and I would expect over time very few users would need this while this adds some dependencies that would be hard to track and reverse later on.

PR #149, opened 2025-08-13
feat: migrate to Authentik OIDC authentication

teddylabdog:
- Replace JWT-based auth with session-based authentication
- Add Authentik OIDC integration with OAuth2 flow
- Remove local database authentication
- Add development mode bypass for testing
- Enhance security with session regeneration and SameSite cookies
- Update all frontend components to use session cookies
- Add comprehensive integration guide

BREAKING CHANGE: Authentication system completely replaced. Users must configure Authentik or use DEV_AUTH_BYPASS for development.


viper151:
This is a very big breakign change and introduces a dependency to this project which I would like to avoid for now. Perhaps in a later major version we can re-introduce this pr

Issue #140, opened 2025-08-09
server.allowedHosts how to edit?


better319:
Blocked request. This host ("xxxxx.app.cloudstudio.work") is not allowed.
To allow this host, add "xxxx.app.cloudstudio.work" to `server.allowedHosts` in vite.config.js.

<img width="742" height="140" alt="Image" src="https://github.com/user-attachments/assets/869eac92-5d7c-4e53-9bda-c1bce9a245a1" />

`

import { defineConfig, loadEnv } from 'vite'
import react from '@vitejs/plugin-react'

export default defineConfig(({ command, mode }) => {
  // Load env file based on `mode` in the current working directory.
  const env = loadEnv(mode, process.cwd(), '')
  
  
  return {
    plugins: [react()],
    server: {
      port: parseInt(env.VITE_PORT) || 5173,
      proxy: {
        '/api': `http://localhost:${env.PORT || 3001}`,
        '/ws': {
          target: `ws://localhost:${env.PORT || 3001}`,
          ws: true
        },
        '/shell': {
          target: `ws://localhost:${env.PORT || 3002}`,
          ws: true
        }
      }
    },
    build: {
      outDir: 'dist'
    }
  }
})

`

better319:
https://ide.cloud.tencent.com/dashboard/

i use it   online

viper151:
@better319 I'm closing this ticket as we now offer cloudcli..ai from where you can actually create cloud development environments. 
Moreover the comment mentioned in the PR is still valid.

PR #166, opened 2025-08-17
feat: add an environment variable to configure `allowedHosts` in `vite.config.js` to access service from alternative hostnames

roykim98:
Didn't open up a separate issue for this as I thought this was rather straight-forward, but this addresses issues like #140 which will allow users to access the service from alternative hostnames without having to maintain a separate fork and manually edit `vite.config.js`.

This was tested locally and it eliminates the following error message.
```
Blocked request. This host ("[REDACTED]") is not allowed.
To allow this host, add "[REDACTED]" to `server.allowedHosts` in vite.config.js.
```

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,3 +10,6 @@
 PORT=3001
 #Frontend port
 VITE_PORT=5173
+
+# Allowed hosts (if additional hosts are needed). Comma-separated string
+ALLOWED_HOSTS=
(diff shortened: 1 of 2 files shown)


ziozzang:
you can build -> run as production mode. that has no limit of such host allow issue.

viper151:
@roykim98  indeed you can build or use the npm package to use it https://www.npmjs.com/package/@siteboon/claude-code-ui

PR #175, opened 2025-08-24
feat: add AUTH_DISABLED environment variable to disable authentication

iamriajul:
This feature allows users to bypass authentication entirely by setting AUTH_DISABLED=true in their environment variables.

Changes:
- Added AUTH_DISABLED environment variable support
- Updated .env.example with AUTH_DISABLED=false default
- Modified authentication middleware to skip auth when disabled
- Updated WebSocket authentication to handle disabled mode
- Updated frontend components to handle auth disabled state
- Added proper status responses for auth disabled mode

When AUTH_DISABLED=true:
- All API endpoints become accessible without authentication
- WebSocket connections work without tokens
- Frontend skips login/setup forms and provides direct access
- Mock user object (id: 1, username: 'admin') used for compatibility

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,3 +10,12 @@
 PORT=3001
 #Frontend port
 VITE_PORT=5173
+
+# =============================================================================
+# AUTHENTICATION CONFIGURATION
+# =============================================================================
+
+# Disable authentication entirely (default: false)
+# When set to true, all features become accessible without login
+# WARNING: Only use in secure, private environments
+AUTH_DISABLED=false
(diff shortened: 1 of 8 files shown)


iamriajul:
## Why?
This is deployed as Coder App, and Coder generates sub domains for each coder app deployed which secured by default using Coder's authentication mechanism, so we don't need double authentication.
## Background
I installed the Claude Code UI in Coder Workspace as a module, So people can interact with their own workspace's Claude Code on the Go.

blackmammoth:
Hey @iamriajul , thanks for the PR! It's an interesting feature. There has been a lot of refactoring done and, as a result, there are some merge conflicts with your PR. Can you resolve them so that it can be reviewed?

blackmammoth:
PR closed as it's continued on #443

PR #184, opened 2025-08-28
feat: Add Japanese translation

Utakata:
This commit adds Japanese translation to the application.

- A new `i18n.js` file has been created to store all the Japanese translation strings.
- The following components have been modified to use the new translations:
  - `App.jsx`
  - `ChatInterface.jsx`
  - `LoginForm.jsx`
  - `SetupForm.jsx`
  - `Sidebar.jsx`
  - `ToolsSettings.jsx`


viper151:
@Utakata we now have an official translation i18n-feat. have a look at it in case you would like to add the japanese translation there.

PR #197, opened 2025-09-14
Use environment variable for Claude path

johnhenry:
# Description

Add `CLAUDE_CLI_PATH` environment variable to optionally specify path to claude code.

# Note
I repeatedly ran into an error complaining that i was using an outdated version of claude code, but I could not for the life of me track down the issue on my system. This is a very straightforward fix.


server/claude-cli.js
viper151, line 1: Can you also add that in the .env.example?

PR #249, opened 2025-11-30
permission dialog

dima-m711:


odbol:
I wonder if this addresses my feature request #285 ?

PR #275, opened 2025-12-31
fix: navigate to the correct session ID when updating session state

blackmammoth:
TLDR:
- Get codex sessions in windows correctly
- Improve message counting logic
- Fix session navigation. Replace the url with the actual session id after the first complete codex response.

This pull request introduces an improvement to the session management logic in the `ChatInterface` component. The main change ensures a smoother transition when a temporary session is replaced with a real session ID, preventing message loss and keeping the UI in sync.

Session transition enhancements:

* Added a system session change marker (`setIsSystemSessionChange(true)`) to prevent messages from being cleared during session ID updates.
* Ensured navigation to the new session by invoking `onNavigateToSession`, keeping the URL and `selectedSession` up to date.
* Get the correct codex session files in windows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented message loss when sessions are created or finalized—chat history is now preserved when session IDs change and the UI navigates to the correct session context.
  * Improved session matching across project paths so sessions align reliably even with different path formats.
  * More accurate parsing and counting of Codex conversation messages to reflect true user and assistant exchanges.

### DIFF
--- a/src/components/Sidebar.jsx
+++ b/src/components/Sidebar.jsx
@@ -595,9 +595,44 @@ function Sidebar({
         </div>
       </div>
 
-      {/* Search Filter and Actions */}
+      {/* Action Buttons - Desktop only - Always show when not loading */}
+      {!isLoading && !isMobile && (
+        <div className="px-3 md:px-4 py-2 border-b border-border">
+          <div className="flex gap-2">
+            <Button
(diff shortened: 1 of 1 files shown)


blackmammoth:
@viper151 This should work

PR #307, opened 2026-01-18
Fix message flash/disappear on send

EricBlanquer:
## Description
Fixes a race condition where user messages would briefly disappear then reappear when sending new messages.

## Root Cause
The issue was caused by two independent code paths that could overwrite optimistic UI updates:

1. **App.jsx (line 189)**: The file watcher checked `activeSessions.has(selectedSession.id)` but didn't account for temporary `new-session-*` IDs. This caused `externalMessageUpdate` to be incremented prematurely, triggering a message reload.

2. **ChatInterface.jsx (line 3139)**: The useEffect syncing `chatMessages` with `convertedMessages` ran even during `isLoading`, overwriting optimistic UI updates before the server responded.

### App.jsx
- Added check for temporary session IDs in active session detection
- Prevents race condition where `activeSessions` hasn't been updated yet with the real session ID

### ChatInterface.jsx  
- Added `!isLoading` condition to the chatMessages sync useEffect
- Preserves optimistic UI updates during active message sending

## Testing
✅ Tested: Messages now stay visible during the entire send/response cycle without flashing

## Why Both Fixes Are Needed
- The App.jsx fix prevents `externalMessageUpdate` from triggering the useEffect at line 3102, which calls `setChatMessages(converted)` directly (bypassing `isLoading` protection)
- The ChatInterface.jsx fix protects the useEffect at line 3138 that syncs via `convertedMessages`

Without both fixes, messages can still flash via the unprotected code path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Treats newly created/in-flight sessions as active to avoid race-condition updates that could trigger unnecessary reloads.
  * Prevents converted messages from overwriting the optimistic chat UI while data is loading, preserving in-progress message state.

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -209,7 +209,10 @@ function AppContent() {
 
             // Check if this is the currently-selected session
             if (changedSessionId === selectedSession.id) {
-              const isSessionActive = activeSessions.has(selectedSession.id);
+              // Check for active session: either the exact session ID or any temporary new-session-* ID
+              // This prevents race conditions where activeSessions hasn't been updated yet
+              const isSessionActive = activeSessions.has(selectedSession.id) ||
+                Array.from(activeSessions).some(id => id.startsWith('new-session-'));
 
               if (!isSessionActive) {
                 // Session is not active - safe to reload messages
--- a/src/components/ChatInterface.jsx
+++ b/src/components/ChatInterface.jsx
@@ -3181,11 +3181,12 @@ function ChatInterface({ selectedProject, selectedSession, ws, sendMessage, mess
   }, [selectedSession?.id]);
 
   // Update chatMessages when convertedMessages changes
+  // Skip during active sessions to preserve optimistic UI updates
   useEffect(() => {
-    if (sessionMessages.length > 0) {
+    if (sessionMessages.length > 0 && !isLoading) {
       setChatMessages(convertedMessages);
     }
-  }, [convertedMessages, sessionMessages]);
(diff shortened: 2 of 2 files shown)


blackmammoth:
@EricBlanquer can you also send detailed reproduction steps for this (including which LLM you picked for testing)?

EricBlanquer:
@blackmammoth 

## Reproduction steps

### Before the fix (message flash/disappear bug)

1. Start the app and select any project
2. Open a new session or existing session
3. Type a message and send it
4. **Bug**: The message briefly appears, then disappears for ~500ms, then reappears when the server confirms it

This happens because:
- The message is added optimistically to the UI
- A session-change event triggers a reload of messages from the server
- The server hasn't saved the message yet → the reload overwrites the optimistic message
- When the server finally saves, another update restores the message

### After the fix

1. Same steps as above
2. **Expected**: Message appears immediately and stays visible without flashing

### LLM used for testing

Tested with **Claude Sonnet 4** (claude-sonnet-4-20250514) via Claude CLI.

The bug is not LLM-specific - it's a race condition in the frontend between optimistic UI updates and session reload events.


send 1, then send 2 : message 2 has disappeared
image

then reappeared
 
image

blackmammoth:
@EricBlanquer I tried your reproduction and this is what I see on this branch:

There seems to be a lag to see the message after sending a new one.

https://github.com/user-attachments/assets/19186674-3d03-4eb3-9658-c4f11f9914c4

PR #308, opened 2026-01-18
Fix scroll auto-jump and message visibility issues

EricBlanquer:
## Description
Fixes two related scroll issues that impacted user experience when viewing message history.

### 1. Auto-scroll jumping to bottom when loading older messages
**Problem**: When scrolling up to view previous messages, the UI would automatically jump back to the bottom.

**Root cause**: The auto-scroll useEffect (line 4055) triggered on every `chatMessages.length` change, including when pagination loaded older messages. It didn't check if the length change was due to pagination.

**Fix**: Added check for `isLoadingMoreRef.current` to skip auto-scroll during pagination loading.

### 2. Limited message visibility (only 100 messages visible)
**Problem**: Even though pagination loaded 260+ messages, only the last 100 were visible. Users couldn't scroll to see the loaded history.

**Root cause**: `visibleMessageCount` was initialized to 100 and never updated when:
- Loading a session with more than 100 messages
- Pagination loaded additional messages

**Fix**: 
- Initialize `visibleMessageCount` to `Math.max(100, messages.length)` on session load
- Increase `visibleMessageCount` by the number of messages loaded during pagination
- Reset to 100 when clearing session

### ChatInterface.jsx
1. **Line 4059-4061**: Skip auto-scroll if loading older messages (pagination)
2. **Line 3058**: Set visible count for Cursor sessions based on loaded messages
3. **Line 3071**: Set visible count for Claude/Codex sessions based on loaded messages  
4. **Line 2955**: Increase visible count when pagination loads more messages
5. **Line 3089**: Reset visible count to default when clearing session

## Testing
✅ Tested: Can scroll up to view history without auto-jump to bottom
✅ Tested: All loaded messages are now visible (not limited to 100)
✅ Tested: Pagination works seamlessly with the visibility system

## Impact
- Improved UX: Users can freely browse message history
- Better pagination: Loaded messages are actually visible
- Maintains performance: Still uses visibility limiting for very long conversations

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented auto-scroll from interrupting older-message loading so users can page through history without jumpiness.
  * Newly loaded older messages are immediately visible after pagination.
  * Loading state is cleared in a way that avoids premature re-enablement of scrolling or pagination.

* **Improvements**
  * Initial message loads ensure at least 100 messages (or the number available) are visible for better context.
  * Clearing a session now reliably resets the visible message count to 100.

### DIFF
--- a/src/components/ChatInterface.jsx
+++ b/src/components/ChatInterface.jsx
@@ -2977,10 +2977,14 @@ function ChatInterface({ selectedProject, selectedSession, ws, sendMessage, mess
         };
         // Prepend new messages to the existing ones
         setSessionMessages(prev => [...moreMessages, ...prev]);
+        // Increase visible message count to show the newly loaded messages
+        setVisibleMessageCount(prevCount => prevCount + moreMessages.length);
       }
       return true;
     } finally {
-      isLoadingMoreRef.current = false;
+      setTimeout(() => {
+        isLoadingMoreRef.current = false;
+      }, 0);
     }
   }, [hasMoreMessages, isLoadingMoreMessages, selectedSession, selectedProject, loadSessionMessages]);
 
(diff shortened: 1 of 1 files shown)


blackmammoth:
@EricBlanquer Are you still facing this error on `main` branch because a fix had been submitted previously?

PR #311, opened 2026-01-18
Add loading progress indicator

EricBlanquer:
## Summary
- Adds real-time progress feedback when loading projects
- Server broadcasts progress updates via WebSocket
- Sidebar shows current project being loaded with progress bar

before
image

after
image

### DIFF
--- a/server/index.js
+++ b/server/index.js
@@ -80,6 +80,20 @@ import { validateApiKey, authenticateToken, authenticateWebSocket } from './midd
 // File system watcher for projects folder
 let projectsWatcher = null;
 const connectedClients = new Set();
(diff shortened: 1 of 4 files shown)


viper151:
@EricBlanquer wht is the use case for this one ? What is the problem you are trying to solve ? Would love to understand a bit more

EricBlanquer:
> @EricBlanquer wht is the use case for this one ? What is the problem you are trying to solve ? Would love to understand a bit more

Hi @viper151 ,
currently, the loading can takes time without any informations about what is this status

image

with this PR, a progress bar is displayed
image

PR #316, opened 2026-01-20
feat: Add Orchestrator Client Mode for Remote Access via Central Hub

liamhelmer:
## Summary

This PR adds orchestrator client mode to claudecodeui, enabling instances to connect to a central orchestrator server for remote access and management. Users can now access their Claude Code UI instances from anywhere through a unified web interface, with automatic authentication pass-through and real-time status reporting.

## Motivation

When running Claude Code UI on multiple machines (development servers, remote workstations, etc.), users need a way to:

- Access all instances from a single dashboard
- Authenticate once at the orchestrator level
- View real-time status (idle/active/busy) of each instance
- Use the full web interface through a reverse proxy

## Key Features

- **WebSocket-based orchestrator connection** with automatic reconnection and heartbeats
- **HTTP proxy support** allowing full UI access through the orchestrator
- **Auto-authentication pass-through** from orchestrator's GitHub OAuth to local instances
- **Real-time status tracking** (idle → active → busy) reported to orchestrator
- **PWA support through proxy** with service worker URL normalization
- **Graceful degradation** - runs standalone if orchestrator is unavailable

## Architecture

```
┌─────────────────┐     ┌──────────────────────────────────────────────┐
│  claudecodeui   │────▶│         Orchestrator Server                  │
│   instance 1    │     │  (Cloudflare Workers + Durable Objects)      │
└─────────────────┘     │                                              │
                        │  - WebSocket hub per user                    │
┌─────────────────┐     │  - HTTP proxy to instances                   │
│  claudecodeui   │────▶│  - GitHub OAuth authentication               │
│   instance 2    │     │  - Real-time status dashboard                │
└─────────────────┘     │                                              │
                        └──────────────────────────────────────────────┘
     ┌──────────┐                          │
     │ Browser  │◀─────────────────────────┘
     │ (User)   │    (Single entry point for all instances)
     └──────────┘
```

### New Files: `server/orchestrator/`

| File | Purpose |
|------|---------|
| `index.js` | Main orchestrator module - exports all components, provides `initializeOrchestrator()` factory |
| `client.js` | `OrchestratorClient` class - WebSocket client with auto-reconnect, heartbeats, HTTP proxy handling, URL rewriting |
| `protocol.js` | Message protocol definitions for claudecodeui ↔ orchestrator communication |
| `github-auth.js` | GitHub OAuth validation for pass-through authentication (org/team/user restrictions) |
| `status-tracker.js` | Tracks Claude session status (idle/active/busy) and reports changes to orchestrator |
| `proxy.js` | WebSocket proxy utilities for handling user requests routed through orchestrator |

### Modified Files

| File | Changes |
|------|---------|
| `server/cli.js` | Added orchestrator initialization on server startup; environment variable handling |
| `server/middleware/auth.js` | Added debug logging for JWT verification; supports orchestrator-generated tokens |
| `server/database/db.js` | Added `getOrCreateUser()` for auto-creating users from orchestrator OAuth |
| `public/sw.js` | Added `proxyBase` query parameter support for PWA through proxy; URL normalization |
| `src/App.jsx` | Added dynamic React Router `basename` detection for proxy access |
| `.env.example` | Added orchestrator configuration documentation |

## Configuration

```bash
# Orchestrator mode: 'standalone' (default) or 'client'
ORCHESTRATOR_MODE=client

# Orchestrator WebSocket URL
ORCHESTRATOR_URL=wss://orchestrator.example.com/ws/connect

# Authentication token from orchestrator dashboard
ORCHESTRATOR_TOKEN=your-token-here

# Optional: Custom client ID (defaults to hostname-pid)
ORCHESTRATOR_CLIENT_ID=my-dev-machine

# Pass-through auth: restrict which GitHub users can access
ORCHESTRATOR_GITHUB_ORG=your-org
ORCHESTRATOR_GITHUB_USERS=user1,user2
```

### HTTP Proxy URL Rewriting
- HTML: Rewrites `src`, `href`, `action` attributes; injects fetch() patch and auto-auth scripts
- JavaScript: Rewrites known URL prefixes while preserving regex patterns
- Service Worker: Accepts `proxyBase` parameter to normalize cache keys

### Auto-Authentication Flow
1. User authenticates with orchestrator via GitHub OAuth
2. Orchestrator proxies requests with `X-Orchestrator-User-Id` headers
3. claudecodeui generates JWT for orchestrator user (creating user if needed)
4. Token injected into HTML response, stored in localStorage

### Status Tracking
Reports `idle` → `active` → `busy` status to orchestrator based on WebSocket connections and Claude generation state.

## Breaking Changes

None - orchestrator mode is opt-in via environment variables. Existing installations continue to work in standalone mode.

## Related

- Orchestrator server implementation: [ai-orchestrator](https://github.com/liamhelmer/ai-orchestrator)

---

🤖 Generated with [Claude Code](https://claude.com/code)

## Release Notes

* **New Features**
  * Added orchestrator client mode for connecting to central orchestrator servers with WebSocket support
  * Added GitHub organization, team, and user-based access control for orchestrator authentication
  * Added proxy base routing support for improved nested deployment compatibility

* **Documentation**
  * Expanded CLI status output and help text with orchestrator and authentication configuration details

---

## Prompt History

<details>
<summary>📝 Prompt - 2026-01-20 22:06:24 UTC</summary>

```
Fix all of the following. If any would break new functionality, ask before implementing the fix. Assign an agent to each fix for clarity: In `@server/database/db.js`:
- Around line 229-268: The getOrCreateOrchestratorUser function currently
ignores githubId and can bind an orchestrator user to any local account with the
same username; update this logic to (1) validate githubUsername is
non-empty/defined before DB ops, (2) prefer finding users by a persisted
github_id column (add/ensure users.github_id exists), (3) if a user is found by
username but github_id differs, reject or namespace the orchestrator user
instead of reusing the local account, and (4) when creating a new orchestrator
user insert the github_id (and a namespaced username if you choose that
approach) and set last_login; update the SELECT, INSERT and UPDATE statements in
getOrCreateOrchestratorUser to use github_id checks and handle mismatch cases
explicitly.

In `@server/middleware/auth.js`:
- Around line 9-112: The middleware logs sensitive token fragments, decoded
payloads, and secret hashes unguarded (see secretHash, authenticateToken,
generateToken), which must be disabled in production; wrap all
console.log/console.error calls that print token/payload/secret information
behind an explicit AUTH_DEBUG env check (e.g., process.env.AUTH_DEBUG ===
"true") so logs only run when AUTH_DEBUG is enabled, leaving non-sensitive error
handling intact and keeping validateApiKey, authenticateToken, and generateToken
logic unchanged except for gating their debug logging.

In `@server/orchestrator/client.js`:
- Around line 255-264: In sendPing(), clear any existing heartbeat timeout
before setting a new one to prevent stale timers from firing; call
clearTimeout(this.heartbeatTimeoutTimer) (and reset it to null) right before
assigning this.heartbeatTimeoutTimer = setTimeout(...), so overlapping pings
won’t cause a healthy connection to be terminated in sendPing(), keeping the
existing DEFAULTS.heartbeatTimeout and the ws?.terminate() behavior unchanged.
- Around line 501-516: Remove logging of token material: keep the token format
validation for orchestratorToken and still set
fetchOptions.headers["authorization"] = `Bearer ${orchestratorToken}` when
valid, but change the console.log to omit the raw token (do not use
orchestratorToken.substring). Instead log only non-sensitive context such as
orchestratorUsername or a fixed placeholder (e.g., "[REDACTED]" or masked
indicator) so no bearer token characters are emitted; update the message near
the existing token-format check and header assignment in
server/orchestrator/client.js that references orchestratorToken,
orchestratorUsername, and fetchOptions.headers.
- Around line 734-769: The issue is that '.' and other regex metacharacters in
entries of urlPrefixes (e.g., "socket.io", "manifest.json") are being
interpreted by RegExp, causing unintended matches; in rewriteJsUrls(js,
proxyBase) escape each prefix before constructing the RegExp (e.g., compute an
escapedPrefix by replacing /[.*+?^${}()|[\]\\]/g with '\\[Pasted text #1 +107 lines]') and use
escapedPrefix in the three new RegExp(...) calls so the patterns match the
literal prefix characters.
- Around line 525-539: The current code always calls response.text() which
corrupts binary payloads; change the logic around the response handling in
server/orchestrator/client.js (the block that builds responseHeaders,
contentType, and responseBody from response) to detect binary vs text via
contentType (inspect the contentType variable from the response.headers loop),
and for binary types (non-text/* and not application/json or lacking utf-8) read
the body with response.arrayBuffer(), base64-encode it, set a sentinel header
(e.g. add ["x-orch-encoding","base64"] into responseHeaders) and set
responseBody to the base64 string; for text types continue to use
response.text() and keep no encoding header; apply the same change to the second
similar block later in the file where response.text() is used.

In `@server/orchestrator/github-auth.js`:
- Around line 32-47: The githubFetch function currently can hang indefinitely;
modify githubFetch to use an AbortController with an 8 second timeout: create an
AbortController, pass its signal to fetch, set a timeout (8000ms) that calls
controller.abort(), and clear that timeout after the fetch completes (both
success and error) to avoid leaks; ensure the thrown error when aborted is
surfaced as a fetch failure so the existing response.ok handling still applies
(update references to the controller and signal usage inside githubFetch).

In `@server/orchestrator/index.js`:
- Around line 67-69: The function createOrchestratorClientFromEnv uses
require("./client.js") which fails in this ESM module; replace that with either
a static top-level import "import { OrchestratorClient } from './client.js'" or
switch to the existing dynamic-import pattern (use await import("./client.js")
and destructure OrchestratorClient) inside createOrchestratorClientFromEnv and
make the function async if you choose dynamic import so the destructuring works
correctly.

In `@server/orchestrator/protocol.js`:
- Around line 133-139: The createResponseCompleteMessage function currently
omits valid falsy values because it spreads ...(data && { data }); change the
conditional to only exclude when data is null or undefined (e.g., use a
conditional like data !== null && data !== undefined) so that empty string, 0,
false are preserved; update createResponseCompleteMessage to build the returned
object for OutboundMessageTypes.RESPONSE_COMPLETE and include the data property
when data !== null && data !== undefined (leave it out when null/undefined).

In `@server/orchestrator/proxy.js`:
- Around line 215-285: handleUserRequest may destructure payload as undefined
and then do payload.user = ..., causing a crash when githubAuth is configured;
fix by ensuring payload is initialized before mutation: change the
destructuring/assignment so payload is a defined object (e.g., let payload =
message.payload || {}) prior to using githubAuth.validate and assigning
payload.user, and then pass that initialized payload into
OrchestratorProxySocket and handlers.handleChatMessage so subsequent code
(OrchestratorProxySocket, OrchestratorProxyWriter, handlers.handleChatMessage)
can safely read/write payload without throwing.

In `@src/App.jsx`:
- Around line 109-140: The mediaQuery listener leaks because
window.matchMedia("(display-mode: standalone)") is invoked multiple times;
capture the MediaQueryList once (e.g., const mql = window.matchMedia(...)) in
the useEffect, add the "change" listener to mql and remove it from the same mql
in the cleanup, and stop calling matchMedia inside checkPWA. Also move the
touchstart listener setup out of checkPWA (or guard so it only attaches once),
provide a real handler function instead of {} (e.g., handleTouchStart) and
remove that same handler in the useEffect cleanup; update references to
checkPWA, mql, and handleTouchStart accordingly so listeners are added/removed
from the same objects.
```

</details>

<details>
<summary>📝 Prompt - 2026-01-20 22:24:27 UTC</summary>

```
close both PRs from this branch without merging. Force push the branch. Then change the origin to https://github.com/Epiphytic/claudecodeui and push all changes there. Then make new PRs to https://github.com/Epiphytic/claudecodeui and https://github.com/siteboon/claudecodeui/
```

</details>


liamhelmer:
Closing to recreate PR with updated changes

PR #339, opened 2026-01-24
fix: use git root for session tracking to prevent session loss

lukaskraic:
## Summary

When Claude changes directories during work (e.g., `cd` into a subdirectory), sessions were being lost because the UI tracked sessions by the current working directory rather than the git repository root.

This fix detects the git repository root and uses it consistently across all components:

- **Chat tab**: Uses git root for `cwd` option in Claude SDK calls
- **Terminal tab**: Uses git root for PTY session key and resume commands  
- **Source Control tab**: `validateGitRepository` now returns git root instead of throwing an error, allowing all git operations from subdirectories

## Problem

1. User opens project at `/home/user/myproject`
2. Claude navigates to `/home/user/myproject/src/components` during work
3. Session is lost because the UI looks for session at the new path
4. Source Control shows error: "This directory is inside a git repository at /home/user/myproject, but git operations should be run from the repository root"

## Solution

Added `getGitRoot()` helper function that detects the git repository root using `git rev-parse --show-toplevel`. This root is then used for:

1. Session key generation in terminal
2. Working directory for Claude SDK calls
3. All git operations in Source Control

## Test plan

- [ ] Open a git project in ClaudeCodeUI
- [ ] Let Claude change to a subdirectory during work
- [ ] Verify chat session continues without interruption
- [ ] Verify terminal session persists
- [ ] Verify Source Control tab shows correct status and allows commits
- [ ] Test with non-git directories (should fall back to original behavior)

## Files changed

- `server/index.js` - Added `getGitRoot()` helper, patched chat and terminal handlers
- `server/routes/git.js` - Modified `validateGitRepository()` to return git root

## Summary by CodeRabbit

* **New Features**
  * Automatically detect and use repository root for source-control and session scoping.

* **Bug Fixes**
  * Improved session continuity when tools change directories; session keys now stable across directory moves.
  * More reliable shell resume across platforms with repository-root fallbacks.
  * Buffered shell output now scrolls to the bottom when resumed.
  * Session ID sanitization with user-facing warnings for invalid IDs.

* **Improvements**
  * Path-validation to prevent directory traversal when reading/creating files.
  * Safer execution of Git operations and clearer root-based logging.


blackmammoth:
Hey @lukaskraic , thanks for the PR! Can you resolve the merge conflicts?

blackmammoth:
Hello @lukaskraic, I tested this PR with the following reproduction step and the original branch in `main` is working as expected. Do you still see a need for this PR?

1. created a `src/components` subdirectory in `test-folder` manually
2. created a project in 'test-folder' in CloudCLI UI
3. Created a "New session" in the folder
4. Navigated to `Shell` tab
5. Sent a message like: "echo ROOT_MARKER"
6. Navigated to Chat tab and then sent "cd src/components && pwd"
7. The command successfully showed me the working directory in the Chat UI.
8. I returned to "Shell" tab to see that the session was still continuing in the original "test-folder" so no session info was lost.

PR #365, opened 2026-02-02
fix: correct version comparison to only show updates when newer versi…

xukunfeng0496:
…on available

Previously the update notification would show whenever the version was different, even if the current version was newer than the latest release. Now it properly compares semantic versions and only shows the update dialog when a newer version is actually available.

Also adds missing @codemirror/view and @codemirror/state dependencies.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed version comparison logic to accurately determine when updates are available using proper semantic version comparison instead of simple string matching, ensuring users receive correct update notifications.

* **Chores**
  * Added project dependencies to support future enhancements.


viper151:
this serves as a reminder actually to complete the release pipeline :)

PR #375, opened 2026-02-09
fix: claude code login issues

blackmammoth:
1. Now, the browser opens in a new tab automatically when logging using claude code
2. Clicking "C" to copy works
3. I have removed the "x-term" link selector since it didn't select the whole link
4. Users can copy the authentication url whether they are in mobile or pc
5. Resolved double copy pasting in shell

# Notes
- The `/exit` was removed because claude was terminating processes unnecessarily. For e.g. as shown in the following image, it exited the process before I even had the chance to do a `/login` command.
image

## Summary by CodeRabbit

* **New Features**
  * Smarter detection and display of authentication URLs from terminal/chat output, with deduplication and auto-open cues.
  * UI shows a login URL panel with Open and Copy actions; supports copying via clipboard (with fallback).
  * Keyboard shortcut (Ctrl/Cmd+C) to quickly copy auth links in minimal/plain-shell mode.

* **Changes**
  * Login modal remains open after successful completion so users can review output before closing.

### DIFF
--- a/src/components/LoginModal.jsx
+++ b/src/components/LoginModal.jsx
@@ -30,13 +30,13 @@ function LoginModal({
 
     switch (provider) {
       case 'claude':
-        return isAuthenticated ? 'claude setup-token --dangerously-skip-permissions' : 'claude --dangerously-skip-permissions';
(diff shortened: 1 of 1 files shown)


blackmammoth:
I have reverted the `/exit` change since the user doesn't actually need to login if they are using API keys.

PR #407, opened 2026-02-22
fix(database): replace better-sqlite3 with built-in node:sqlite

### LINKED ISSUE #406: npm install fails on Node.js 25 due to better-sqlite3 native compilation
Cloud CLI fails to install on Node.js 25 (current release) because `better-sqlite3` does not support Node 25. The package only supports Node 20, 22, 23, and 24, and there are no prebuilt binaries for Node 25. This forces a native compilation via `node-gyp`, which fails due to removed V8 APIs in Node 25.

The upstream issue is tracked at: https://github.com/WiseLibs/better-sqlite3/issues/1411

### Steps to Reproduce

1. Use Node.js v25.x (e.g. `v25.2.1`)
2. Clone the repo: `git clone https://github.com/siteboon/claudecodeui.git`
3. Run `npm install`
4. `better-sqlite3` fails to build with errors like:
   - `No prebuilt binaries found`
   - `gyp ERR! find VS could not find a version of Visual Studio 2017 or newer` (Windows)
   - `fatal error: util/macros.cpp: No such file or directory` (Linux)

### Environment

- **OS:** Windows 11
- **Node.js:** v25.2.1
- **npm:** 11.x

### Analysis

Only `server/database/db.js` uses `better-sqlite3`. The other SQLite dependency (`sqlite3` + `sqlite` used in `cursor.js` and `projects.js`) uses Node-API and works fine on Node 25.

### Suggested Fix

Replace `better-sqlite3` with Node's built-in `node:sqlite` module (`DatabaseSync`). The API is nearly identical (`.prepare()`, `.exec()`, `.run()`, `.get()`, `.all()`), requires zero additional dependencies, and works on Node 22.5+.

This would:
- Remove the native compilation requirement entirely
- Support all current and future Node versions
- Reduce the dependency count

PaloSP:
## Summary
- Replace `better-sqlite3` native dependency with Node.js built-in `node:sqlite` (`DatabaseSync`)
- Remove `better-sqlite3` from `package.json`
- No other code changes needed — the APIs are compatible

## Motivation
`npm install` fails on Node.js 25+ because `better-sqlite3` has no prebuilt binaries for v25 and requires Visual Studio C++ build tools for native compilation. The built-in `node:sqlite` module eliminates this entirely.

Closes #406

## Test plan
- [x] `npm install` completes without errors
- [x] Server starts and database initializes
- [x] Login/registration flow works

## Summary by CodeRabbit

* **Chores**
  * Updated the underlying database library implementation to improve dependency management and system compatibility.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,7 +31,6 @@
         "@xterm/addon-webgl": "^0.18.0",
         "@xterm/xterm": "^5.5.0",
         "bcrypt": "^6.0.0",
-        "better-sqlite3": "^12.2.0",
         "chokidar": "^4.0.3",
         "class-variance-authority": "^0.7.1",
         "clsx": "^2.1.1",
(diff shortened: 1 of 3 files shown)


PaloSP:
@siteboon Good point from CodeRabbit about the Node.js version. Two options:

1. Set `engines` to `>=23.4.0` (where `node:sqlite` became stable, no flag needed)
2. Set `engines` to `>=22.5.0` and add `--experimental-sqlite` to the start scripts to support Node 22 LTS

Which would you prefer?

blackmammoth:
@PaloSP thanks, I verified the issue with the amazingly detailed PR description you provided. However, I think it's better to update `better-sqlite3` to the latest version instead since it has support for Node 25

blackmammoth:
#445

PR #461, opened 2026-03-01
fix: prevent React 18 batching from losing messages during session sync

shikihane:
## Summary

- Prevent `sessionMessages` from overwriting user's just-sent message during React 18 batch updates
- Add guards so `setChatMessages` only fires when `sessionMessages` actually grew (new server data), not during re-renders
- Track `lastLoadedSessionId` to skip redundant session reloads

## Problem

When a user sends a message, React 18 batches state updates. The `sessionMessages` effect would fire and overwrite `chatMessages` with stale server data, causing the user's message to briefly disappear. In some cases, the assistant response was lost in the same batch.

## Changes

- `useChatSessionState.ts`: Add `lastLoadedSessionIdRef` to skip redundant loads; add length-based guards and active-processing detection to the `sessionMessages → chatMessages` sync effect

## Test plan

- [ ] Send a message and verify it doesn't briefly disappear
- [ ] Verify assistant responses appear correctly after sending
- [ ] Switch between sessions and verify messages load properly
- [ ] Verify `npm run build` passes

## Summary by CodeRabbit

* **Refactor**
  * Cache last-loaded session to skip redundant reloads, speeding session switches and reducing UI flicker.
  * Improve message synchronization: update local chat view only when messages truly change (initial load, growth, or clear), avoiding unnecessary overwrites.
  * Reset cache on session deselect and add guards around loading state and local persistence for more stable save/restore behavior.
  * No public APIs or signatures were changed.

### DIFF
--- a/src/components/chat/hooks/useChatSessionState.ts
+++ b/src/components/chat/hooks/useChatSessionState.ts
@@ -93,7 +93,7 @@ export function useChatSessionState({
   const scrollPositionRef = useRef({ height: 0, top: 0 });
   const loadAllFinishedTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
   const loadAllOverlayTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
-  const lastLoadedSessionIdRef = useRef<string | null>(null);
+  const lastLoadedSessionKeyRef = useRef<string | null>(null);
 
   const createDiff = useMemo<DiffCalculator>(() => createCachedDiffCalculator(), []);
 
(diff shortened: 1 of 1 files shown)


shikihane:
@coderabbitai review

blackmammoth:
Hey @shikihane, thanks for this PR!!! The issue had been bugging me as hell and was a priority in my todo list! 

Also, if you’d like to join project discussion or community chat, we have a Discord here: [link](https://discord.gg/buxwujPNRE). No pressure at all — just sharing in case it’s useful.

PR #465, opened 2026-03-01
fix: detect router basename for path-prefixed deployments

### LINKED ISSUE #464: Blank page in path-prefixed deployments because Router basename is never initialized
## Summary
When CloudCLI UI is deployed under a path prefix such as `/ai` behind a reverse proxy, authentication can succeed but the app renders a blank page after login or when opening existing session links.

## Reproduction
1. Deploy `claudecodeui` behind a reverse proxy under a path prefix, for example `/ai`.
2. Rewrite static asset and API paths so the HTML loads correctly under that prefix.
3. Open `/ai/` and log in.
4. Open an existing session link such as `/ai/session/<id>` or let the app render the post-login routed view.

## Actual result
The login/setup screens render, but the routed application view becomes blank because React Router is still using `basename=""`.

## Why this happens
`src/App.tsx` currently uses:

```tsx
<Router basename={window.__ROUTER_BASENAME__ || ''}>
```

but `window.__ROUTER_BASENAME__` is never initialized by the app. In prefixed deployments the app HTML and assets can be rewritten successfully by the proxy, while the router still thinks it is mounted at `/`.

That makes URLs like `/ai/session/:sessionId` fail to match `/session/:sessionId`.

## Proposed fix
Detect the router basename from already-available runtime hints:
- `manifest` href
- module script src
- icon href

Those URLs already contain the deployed prefix after proxy rewriting or non-root hosting, so the router can derive the correct basename without requiring an extra injected global.

## Notes
I have a small patch ready that keeps the existing `window.__ROUTER_BASENAME__` override if someone explicitly provides it, and otherwise auto-detects the prefix.

JohnGenri:
## Summary
Fix blank-page routing failures when CloudCLI UI is deployed behind a reverse proxy under a path prefix such as `/ai`.

Closes #464.

## Problem
`src/App.tsx` currently uses `window.__ROUTER_BASENAME__ || ''`, but that global is not initialized by the app.

As a result, path-prefixed deployments can successfully rewrite HTML, assets, and API calls, while React Router still assumes it is mounted at `/`.

That breaks routed views after login and direct links like:
- `/ai/session/:sessionId`

## Solution
Keep the explicit `window.__ROUTER_BASENAME__` override when present, but otherwise derive the basename from runtime URLs that already reflect the deployed prefix:
- manifest href
- module script src
- icon href

This makes prefixed deployments work without requiring an additional server-side bootstrap script.

## Validation
Built locally with Vite after the change and verified that the generated app bundle was updated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Routing now dynamically determines the app base path: it honors an explicit override when present, otherwise inspects page-provided hints (manifest, script and icon links) and ignores invalid values. If nothing reliable is found, an empty base is used. This improves routing reliability across varied hosting setups and prevents incorrect base-path resolution at runtime.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -9,6 +9,9 @@ import { PluginsProvider } from './contexts/PluginsContext';
 import AppContent from './components/app/AppContent';
 import i18n from './i18n/config.js';
 
+/**
+ * Detect the router basename from explicit runtime config or deployment hints.
+ */
 function detectRouterBasename() {
   const explicitBasename = typeof window !== 'undefined' ? window.__ROUTER_BASENAME__ || '' : '';
   if (explicitBasename) {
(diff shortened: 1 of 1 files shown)


JohnGenri:
Validation note from the environment where this was reproduced:

- Deployment mode: reverse proxy under `/ai`
- Before this patch: login/setup screen rendered, but the post-login routed UI could go blank and direct links like `/ai/session/<id>` failed
- After this patch: the router picks up the effective prefix and the routed UI works correctly under `/ai`

So the fix is not theoretical; it resolves the concrete path-prefixed deployment failure that triggered issue #464.

JohnGenri:
Addressed both CodeRabbit findings in a follow-up commit:

- switched URL resolution from `window.location.origin` to `document.baseURI || window.location.href`
- scan all runtime path hints and keep the longest detected normalized prefix instead of returning the first match

This makes basename detection more robust for path-prefixed deployments where some hints may resolve to root while others preserve the actual prefix.

JohnGenri:
Validation note after refreshing this branch onto current `main`:

- Issue `#464` is still reproducible in a real path-prefixed deployment behind `nginx` under `/ai`.
- The HTML shell, assets, API, and login flow work there, but the routed app view can go blank after login, and direct URLs like `/ai/session/<id>` fail because the client router still ends up with `basename=""`.
- This branch is now rebased onto the current upstream `main` and still fixes that specific routing failure by deriving the effective basename from runtime deployment hints when `window.__ROUTER_BASENAME__` is not initialized.

So this is still an active bug, and the PR remains intentionally scoped to the client-side router basename detection only.

JohnGenri:
Addressed the remaining CodeRabbit edge case in `111431c`.

- broadened icon hint detection beyond `favicon.svg|png` to cover `favicon.ico`, `apple-touch-icon`, `mask-icon`, and other icon-style filenames
- scan all relevant icon link tags instead of a single `rel="icon"`
- normalize trailing asset-style directories so icon URLs like `/ai/assets/icons/icon-192.png` still resolve to basename `/ai`

The branch remains intentionally scoped to runtime basename detection for issue `#464`.

patrickstorm:
Also interested in this PR being merged

JohnGenri:
Rebased this PR onto current upstream main (v1.29.5 / 25b00b5). The change remains scoped to src/App.tsx and still preserves the explicit window.__ROUTER_BASENAME__ override while detecting path-prefixed deployments from runtime asset hints.\n\nValidation run after rebase:\n\n- npm run typecheck: passed\n- npm run lint: passed with existing warnings, no errors\n- npm run build: passed\n\nThis still addresses issue #464 for deployments served under a path prefix such as /ai/.

JohnGenri:
Just a small follow-up: this PR has now been rebased onto the latest main (v1.29.5), and the change is still scoped to src/App.tsx.

I re-tested the path-prefixed deployment case under /ai/:
- without basename detection, routed views such as /ai/session/:id can fail because Router still receives basename=""
- with this patch, the app derives the effective prefix from same-origin runtime asset hints and keeps direct routed URLs working

Validation after rebase:
- npm run typecheck: passed
- npm run lint: passed with existing warnings, no errors
- npm run build: passed

Happy to adjust the approach if maintainers prefer an explicit server-provided basename instead of runtime detection.

JohnGenri:
Rebased this PR onto current upstream main (`v1.30.0` / `f6200e3`). The branch still changes only `src/App.tsx` and keeps the fix scoped to runtime router basename detection for path-prefixed deployments.

Validation after the rebase:

- `npm run lint`: passed with existing warnings, no errors
- `npm run build:client`: passed with existing upstream build warnings
- `npm run typecheck`: currently blocked by upstream server-side typing errors on `v1.30.0` (`cross-spawn`, `express`, `better-sqlite3` declarations / implicit any), unrelated to this client-only PR

The `/ai/` path-prefix case from issue #464 is still valid: without an initialized router basename, direct routed URLs such as `/ai/session/:id` can fail even when the shell/assets/login flow load correctly.

JohnGenri:
Follow-up after updating the real production deployment:

- The server where this issue was reproduced is now running `@cloudcli-ai/cloudcli` `v1.30.0`.
- This PR branch is rebased onto the same current upstream version (`v1.30.0` / `f6200e3`).
- The deployed server includes this router basename fix and the `/ai/` path-prefixed deployment is working with the current codebase.
- The PR diff remains scoped to `src/App.tsx` only.

blackmammoth:
Hey @JohnGenri, thanks for the PR! I found a bug with the implementation for root deployments and submitted PR 815 which would resolve the issue.

 @patrickstorm and @michal-bouska, you can track the progress there.

Issue #14, opened 2025-07-08
Hyphens in folder name make the files list not show up


paraddox:
<img width="1999" height="1483" alt="Image" src="https://github.com/user-attachments/assets/e6543de7-6242-464d-84ce-067ce8bc40c2" />
The project list now shows proper names with hyphens but in the backend the file paths still have the hyphens replaced with /

viper151:
@paraddox  Thank you so much for the error logs.  Should be fixed now. There was the same issue on the files as well.

PR #472, opened 2026-03-02
fix: resolve project paths from sessions-index.json and persist to config

### LINKED ISSUE #115: [BUG] Git Path Resolution Issue
## Summary
The git routes are failing to resolve project paths correctly, causing "Project path not found" errors when trying to access git functionality.

## Error Details
```
Git status error: Error: Project path not found: /Volumes/MasterData/Project/Sandbox/claudecodeui
```

### 1. Path Encoding/Decoding Mismatch
- **Expected path**: `/Volumes/MasterData/Project-Sandbox/claudecodeui`
- **Actual resolved path**: `/Volumes/MasterData/Project/Sandbox/claudecodeui`

The issue occurs in the path encoding/decoding process:
- Project name: `-Volumes-MasterData-Project-Sandbox-claudecodeui`
- When decoded: `projectName.replace(/-/g, '/')` → `/Volumes/MasterData/Project/Sandbox/claudecodeui`
- **Problem**: The original path contains `Project-Sandbox` but gets decoded to `Project/Sandbox`

### 2. Path Construction Logic
The bug is in the `extractProjectDirectory` function in `server/projects.js`:

```javascript
// Line 85: Fallback logic
extractedPath = projectName.replace(/-/g, '/');
```

This simple dash-to-slash replacement doesn't handle cases where the original path contains hyphens that should remain as hyphens.

### 3. File System Reality
- **Actual directory**: `/Volumes/MasterData/Project-Sandbox/claudecodeui`
- **Decoded path**: `/Volumes/MasterData/Project/Sandbox/claudecodeui`
- **Result**: Directory doesn't exist, causing `ENOENT` error

## Impact
- Git status endpoint fails
- Git branches endpoint fails  
- Git remote status endpoint fails
- All git-related functionality is broken for projects with hyphens in their paths

## Affected Files
- `server/routes/git.js` - Git route handlers
- `server/projects.js` - Project path extraction logic
- `server/index.js` - File tree endpoint (line 409)

### Option 1: Fix Path Encoding/Decoding
Implement proper path encoding that preserves hyphens in directory names:
- Use a different separator (e.g., `_` or `%2D`)
- Implement URL-safe encoding/decoding

### Option 2: Improve Fallback Logic
Enhance the `extractProjectDirectory` function to:
- Try multiple path variations
- Check if decoded path exists before using it
- Fall back to original project name if path doesn't exist

### Option 3: Store Original Paths
Modify the project creation process to store the original absolute path and use that instead of relying on encoding/decoding.

## Priority
**High** - This breaks core git functionality for projects with hyphens in their paths.

## Steps to Reproduce
1. Create a project in a directory with hyphens (e.g., `Project-Sandbox`)
2. Try to access git status in the UI
3. Observe "Project path not found" error

## Environment
- OS: macOS (darwin 24.5.0)
- Project path: `/Volumes/MasterData/Project-Sandbox/claudecodeui`
- Encoded project name: `-Volumes-MasterData-Project-Sandbox-claudecodeui`

PabloLION:
## Summary

- Read `projectPath` from `sessions-index.json` when `.jsonl` files are absent
- Persist successfully resolved paths to `~/.claude/project-config.json` so they survive server restarts
- Serialize config writes through a promise queue to prevent race conditions

## Problem

`extractProjectDirectory()` resolves project paths by reading `cwd` from `.jsonl` session files, but:

1. **Misses `sessions-index.json`** — projects that have a session index but no `.jsonl` files fall through to the lossy decode immediately
2. **Never persists results** — on server restart, it re-parses everything. When parsing fails, it falls back to `projectName.replace(/-/g, '/')` which is lossy for paths containing hyphens

Example: `/Applications/Auto-Claude.app/Contents/Resources/backend` encodes as `-Applications-Auto-Claude-app-Contents-Resources-backend`, but decodes to `/Applications/Auto/Claude/app/Contents/Resources/backend` — a path that doesn't exist. This causes `posix_spawnp failed` errors in the shell terminal.

## Approach

**Single-file fix in `server/projects.js`** — no caller changes needed:

1. Before falling back to lossy decode, check `sessions-index.json` for `projectPath` entries
2. New `persistOriginalPath()` helper writes resolved paths to `project-config.json` (fire-and-forget, serialized via promise queue)
3. `fromJsonl` flag tracks whether the path came from a reliable source (JSONL `cwd` or session index)
4. After successful extraction → persist to config → future lookups use config directly (line 288)

## Test plan

- [x] Start production server, open a project with hyphens in its path (`Auto-Claude.app`)
- [x] Verify the terminal resolves the correct path instead of the lossy decode
- [x] Verify `~/.claude/project-config.json` gets an `originalPath` entry after first load
- [ ] Restart server — project should resolve instantly from config without re-parsing

## Related issues

Fixes #115, related to #8, #14

### DIFF
--- a/.github/workflows/discord-release.yml
+++ b/.github/workflows/discord-release.yml
@@ -1,0 +1,22 @@
+name: Discord Release Notification
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  github-releases-to-discord:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Github Releases To Discord
+        uses: SethCohen/github-releases-to-discord@v1.19.0
+        with:
+          webhook_url: ${{ secrets.DISCORD_WEBHOOK_URL }}
+          color: "2105893"
+          username: "Release Changelog"
+          avatar_url: "https://cdn.discordapp.com/avatars/487431320314576937/bd64361e4ba6313d561d54e78c9e7171.png"
+          content: "||@everyone||"
+          footer_title: "Changelog"
+          reduce_headings: true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,24 @@
 All notable changes to CloudCLI UI will be documented in this file.
 
 
+## [1.22.0](https://github.com/siteboon/claudecodeui/compare/v1.21.0...v1.22.0) (2026-03-03)
+
+### New Features
+
+* add community button in the app ([84d4634](https://github.com/siteboon/claudecodeui/commit/84d4634735f9ee13ac1c20faa0e7e31f1b77cae8))
+* Advanced file editor and file tree improvements ([#444](https://github.com/siteboon/claudecodeui/issues/444)) ([9768958](https://github.com/siteboon/claudecodeui/commit/97689588aa2e8240ba4373da5f42ab444c772e72))
+* update document title based on selected project ([#448](https://github.com/siteboon/claudecodeui/issues/448)) ([9e22f42](https://github.com/siteboon/claudecodeui/commit/9e22f42a3d3a781f448ddac9d133292fe103bb8c))
+
+### Bug Fixes
+
+* **claude:** correct project encoded path ([#451](https://github.com/siteboon/claudecodeui/issues/451)) ([9c0e864](https://github.com/siteboon/claudecodeui/commit/9c0e864532dcc5ce7ee890d3b4db722872db2b54)), closes [#447](https://github.com/siteboon/claudecodeui/issues/447)
+* **claude:** move model usage log to result message only ([#454](https://github.com/siteboon/claudecodeui/issues/454)) ([506d431](https://github.com/siteboon/claudecodeui/commit/506d43144b3ec3155c3e589e7e803862c4a8f83a))
+* missing translation label ([855e22f](https://github.com/siteboon/claudecodeui/commit/855e22f9176a71daa51de716370af7f19d55bfb4))
+
+### Maintenance
+
+* add Gemini-CLI support to README ([#453](https://github.com/siteboon/claudecodeui/issues/453)) ([503c384](https://github.com/siteboon/claudecodeui/commit/503c3846850fb843781979b0c0e10a24b07e1a4b))
+
 ## [1.21.0](https://github.com/siteboon/claudecodeui/compare/v1.20.1...v1.21.0) (2026-02-27)
 
 ### New Features
--- a/README.md
+++ b/README.md
@@ -1,12 +1,20 @@
 <div align="center">
-  <img src="public/logo.svg" alt="Claude Code UI" width="64" height="64">
+  <img src="public/logo.svg" alt="CloudCLI UI" width="64" height="64">
   <h1>Cloud CLI (aka Claude Code UI)</h1>
 </div>
 
 
-A desktop and mobile UI for [Claude Code](https://docs.anthropic.com/en/docs/claude-code), [Cursor CLI](https://docs.cursor.com/en/cli/overview) and [Codex](https://developers.openai.com/codex). You can use it locally or remotely to view your active projects and sessions in Claude Code, Cursor, or Codex and make changes to them from everywhere (mobile or desktop). This gives you a proper interface that works everywhere. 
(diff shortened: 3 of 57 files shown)


PabloLION:
Addressed CodeRabbit feedback in the force-pushed commit:

**Race condition (Major)** — Fixed. Writes are now serialized through a promise queue (`persistQueue`). Each write waits for the previous one to complete before doing read-modify-write.

**ENOENT check (Minor)** — Fixed. All 4 `catch` blocks now check `err?.code === 'ENOENT'` before logging the lossy decode warning. Permission errors and other failures are silently ignored.

**Validate JSONL path before persisting (Nitpick)** — Intentionally not adopted. The `cwd` from JSONL files is what Claude Code actually used as its working directory — it's the most reliable source of truth for the original path. If the directory was later deleted or moved, we still want to remember the correct path (e.g., the user may restore it). Validating would defeat the purpose of persisting paths that survive server restarts.

PabloLION:
Addressed round 2 CodeRabbit feedback:

**Extract warning helper (Nitpick)** — Done. Created `warnIfLossyDecodeMissing()` and replaced all 4 inline blocks with calls to it.

**Validate JSONL path before persisting (Major)** — Accepted. Now checks `fs.access()` before calling `persistOriginalPath()`. Stale paths from JSONL won't become sticky in config.

**Serialize ALL config mutations (Major)** — Not adopted. This would require refactoring `renameProject`, `addProjectManually`, `deleteProject`, and other pre-existing config writers — a much larger change that's out of scope for this fix. The race condition between `persistOriginalPath` and those functions existed before this PR (they all do unserialized read-modify-write). Our queue serializes the new code path we're adding. Fixing the broader serialization issue should be a separate PR.

Requesting re-review: @coderabbitai review

blackmammoth:
Hey @PabloLION, thanks for the PR! However, I couldn't replicate the issue. I have used both Windows and Linux for testing.
Reproduction steps I followed:
1. Created a workspace called `project-path`
2. Created a session in that workspace
3. Moved to "Shell" tab.
4. Observe which path it was following. 

After following the above steps, I see that the "Shell" tab is using the correct `project-path` as well.

PabloLION:
> Hey @PabloLION, thanks for the PR! However, I couldn't replicate the issue. I have used both Windows and Linux for testing.
> 
> Reproduction steps I followed:
> 
> 1. Created a workspace called `project-path`
> 
> 2. Created a session in that workspace
> 
> 3. Moved to "Shell" tab.
> 
> 4. Observe which path it was following. 
> 
> 
> 
> After following the above steps, I see that the "Shell" tab is using the correct `project-path` as well.

Not sure. It happened on my Mac and it happens in the Claude code login backend. Anyway after this fix it's still problematic because the path doesn't exists.

blackmammoth:
Could it be due to another issue? If you share a detailed reproduction step, maybe we can figure it out.

> > Hey @PabloLION, thanks for the PR! However, I couldn't replicate the issue. I have used both Windows and Linux for testing.
> > Reproduction steps I followed:
> > 
> > 1. Created a workspace called `project-path`
> > 2. Created a session in that workspace
> > 3. Moved to "Shell" tab.
> > 4. Observe which path it was following.
> > 
> > After following the above steps, I see that the "Shell" tab is using the correct `project-path` as well.
> 
> Not sure. It happened on my Mac and it happens in the Claude code login backend. Anyway after this fix it's still problematic because the path doesn't exists.

PabloLION:
> Could it be due to another issue? If you share a detailed reproduction step, maybe we can figure it out.
> 
> 
> 
> > > Hey @PabloLION, thanks for the PR! However, I couldn't replicate the issue. I have used both Windows and Linux for testing.
> 
> > > Reproduction steps I followed:
> 
> > > 
> 
> > > 1. Created a workspace called `project-path`
> 
> > > 2. Created a session in that workspace
> 
> > > 3. Moved to "Shell" tab.
> 
> > > 4. Observe which path it was following.
> 
> > > 
> 
> > > After following the above steps, I see that the "Shell" tab is using the correct `project-path` as well.
> 
> > 
> 
> > Not sure. It happened on my Mac and it happens in the Claude code login backend. Anyway after this fix it's still problematic because the path doesn't exists.
> 
> 
> 
> 

Yeah, I guess it's because the way I started it. I saw a path in the application folder. But I only installed with npm / clone + build. I guess there were two or more issues. The first one is what I'm trying to fix with this PR and the second one I don't know how to solve it. We can discuss more if closing this PR is not your final decision.

PR #474, opened 2026-03-03
 build(docker): add dev-friendly container runtime with agent tooling and host mounts

yusing:
What does it do?

This PR introduces Docker support from scratch for the project by adding:

- A new multi-stage Dockerfile for:
    - dependency installation (deps-prod, deps)
    - application build (builder)
    - runtime image (runner)
- A new compose.yml service definition (claude-code-ui) with:
    - image build config
    - port mapping (3001:3001)
    - runtime environment and user mapping
    - bind mounts for app data, local agent configs, project workspace, and agent binaries
- A new .dockerignore to reduce build context size and exclude local/dev artifacts

Prove of work:

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,28 @@
-FROM node:22-bookworm-slim AS node
+FROM node:24-trixie-slim AS node
 
-# required for node-pty
+# build-essential and python3 are required for node-pty
+# others are for development inside the container
 RUN apt-get update && apt-get install -y --no-install-recommends \
     build-essential \
     python3 \
+    golang \
+    git \
+    ca-certificates \
+    libglib2.0-0 \
+    bash \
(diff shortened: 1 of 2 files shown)


viper151:
@yusing thank you very much for working on this. We've thought about Docker support in the past and while it makes sense for hosted/platform deployments which [CloudCLI](https://cloudcli.ai) already offers, Claude Code UI is primarily a local dev tool that needs deep access to the user's environment: their shell config, SSH keys, git credentials, CLI auth tokens, and dotfiles like ~/.claude/projects/. Containerizing it would require so many bind mounts that it effectively defeats the purpose of isolation. If you're looking for a containerized experience, we'd recommend checking out [CloudCLI](https://cloudcli.ai) which is built for exactly that.

PR #501, opened 2026-03-08
fix: replace getDatabase with better-sqlite3 db in getGithubTokenById

b0x42:
## Bug

When creating a project using a stored GitHub token, the following error is thrown:

```
getDatabase is not a function
```

## Root Cause

`getGithubTokenById` in `server/routes/projects.js` imports `getDatabase` from `database/db.js`:

```js
const { getDatabase } = await import('../database/db.js');
const db = await getDatabase();
const credential = await db.get(sql, [params]);
```

However, `db.js` uses `better-sqlite3` and does **not** export a `getDatabase` function — it exports the `db` instance directly. The code also uses the async `sqlite` package API (`db.get(sql, array)`) which is incompatible with `better-sqlite3`'s synchronous API.

## Fix

Import `db` directly and use the correct `better-sqlite3` API:

```js
const { db } = await import('../database/db.js');
const credential = db.prepare(sql).get(...params);
```

## How to Reproduce

1. Set up a self-hosted instance
2. Add a GitHub token via the UI
3. Create a new project using that stored token
4. Observe `getDatabase is not a function` error on the confirm step

## Summary by CodeRabbit

* **Refactor**
  * Improved internal code structure for enhanced maintainability and consistency.

### DIFF
--- a/server/routes/projects.js
+++ b/server/routes/projects.js
@@ -311,13 +311,11 @@ router.post('/create-workspace', async (req, res) => {
  * Helper function to get GitHub token from database
  */
 async function getGithubTokenById(tokenId, userId) {
-  const { getDatabase } = await import('../database/db.js');
-  const db = await getDatabase();
+  const { db } = await import('../database/db.js');
 
-  const credential = await db.get(
-    'SELECT * FROM user_credentials WHERE id = ? AND user_id = ? AND credential_type = ? AND is_active = 1',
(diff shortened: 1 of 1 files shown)


viper151:
@b0x42 This is a nice find. Would love to see sqlite3 being replaced across with better sqlite3 as we now seem to have both .

PR #519, opened 2026-03-09
feat: add copy as text or markdown feature for assistant messages

blackmammoth:
If the message is a code command or file editor view, the copy icon won't be shown. For other assistant messages, the copy icon will be shown.

To see the icon, open any session and look below the messages.

## Summary by CodeRabbit

* **New Features**
  * Enhanced copy-to-clipboard control for messages with two formats (Markdown and plain text) and visual success feedback.
  * Format selection dropdown for assistant responses; user messages default to plain-text copy.
  * Conditional copy controls: certain tool/assistant responses can suppress copy UI to prevent unintended copying.
  * Improved accessibility and keyboard/outside-click handling for the copy UI.

### DIFF
--- a/shared/modelConstants.js
+++ b/shared/modelConstants.js
@@ -13,14 +13,14 @@
 export const CLAUDE_MODELS = {
   // Models in SDK format (what the actual SDK accepts)
   OPTIONS: [
-    { value: 'sonnet', label: 'Sonnet' },
-    { value: 'opus', label: 'Opus' },
-    { value: 'haiku', label: 'Haiku' },
-    { value: 'opusplan', label: 'Opus Plan' },
-    { value: 'sonnet[1m]', label: 'Sonnet [1M]' }
+    { value: "sonnet", label: "Sonnet" },
+    { value: "opus", label: "Opus" },
+    { value: "haiku", label: "Haiku" },
+    { value: "opusplan", label: "Opus Plan" },
+    { value: "sonnet[1m]", label: "Sonnet [1M]" },
   ],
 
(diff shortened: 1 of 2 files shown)


blackmammoth:
#493

src/components/chat/view/subcomponents/MessageComponent.tsx
```diff
@@ -49,18 +50,32 @@ const MessageComponent = memo(({ message, prevMessage, createDiff, onFileOpen, o
...
+  const userCopyContent = String(message.content || '');
+  const formattedMessageContent = useMemo(
+    () => formatUsageLimitText(String(message.content || '')),
+    [message.content]
+  );
+  const assistantCopyContent = message.isToolUse
+    ? String(message.displayText || message.content || '')
+    : formattedMessageContent;
+  const isCommandOrFileEditToolResponse = Boolean(
+    message.isToolUse && COPY_HIDDEN_TOOL_NAMES.has(String(message.toolName || ''))
```
viper151, line 66: @blackmammoth  would it make  more sense to have this as a tool parameter that is passed though the toolrendered instead of having it on the main view ?

viper151 (changes requested):
@blackmammoth see the inline comments

src/components/chat/view/subcomponents/MessageComponent.tsx
blackmammoth: I think we are passing many props. Since this is just used here, it's better to localize the variable usage. If we find out that it can be useful in other places later on, we can move it out

PR #521, opened 2026-03-09
feat: Browser autofill support for login form

b0x42:
## Problem

Password managers (1Password, Bitwarden, etc.) were not offering to fill credentials on the login and setup forms. This is because browsers and password manager extensions rely on two standard HTML attributes to detect and autofill credential fields:

- `name` — identifies the field in form submissions and is used by password managers to recognize field types
- `autocomplete` — the WHATWG standard hint that tells the browser/extension what type of data a field expects

The `AuthInputField` component was rendering `<input>` elements with neither attribute, so password managers could not identify the fields and did not offer autofill.

## Solution

- **`AuthInputField.tsx`** — added optional `name` and `autoComplete` props; `name` falls back to the `id` value if not provided
- **`LoginForm.tsx`** — passes `autoComplete="username"` and `autoComplete="current-password"` to the respective fields
- **`SetupForm.tsx`** — passes explicit `name` props (`name="username"`, `name="password"`, `name="confirmPassword"`) alongside `autoComplete="username"` and `autoComplete="new-password"` (signals a registration form so password managers offer to save new credentials); explicit `name` props make the password-manager contract self-contained and resilient to future `id` changes

## Documentation

Added JSDoc comments to `AuthInputField`, `LoginForm`, `SetupForm`, and the `validateSetupForm` helper to meet the project's docstring coverage threshold.

## Test plan

- [x] Open the login page with a password manager extension active (e.g. Bitwarden, 1Password)
- [x] Click the username or password field — the password manager should offer to fill saved credentials
- [x] Use the extension keyboard shortcut to trigger autofill — it should work
- [x] Open the setup/registration page and verify the password manager offers to save the new credentials after submission

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1909,6 +1909,7 @@
       "cpu": [
         "arm64"
       ],
+      "dev": true,
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -1925,6 +1926,7 @@
       "cpu": [
         "x64"
       ],
+      "dev": true,
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
--- a/server/routes/git.js
+++ b/server/routes/git.js
@@ -651,26 +651,28 @@ router.get('/branches', async (req, res) => {
     
     // Get all branches
     const { stdout } = await spawnAsync('git', ['branch', '-a'], { cwd: projectPath });
-    
-    // Parse branches
-    const branches = stdout
+
+    const rawLines = stdout
       .split('\n')
-      .map(branch => branch.trim())
-      .filter(branch => branch && !branch.includes('->')) // Remove empty lines and HEAD pointer
-      .map(branch => {
-        // Remove asterisk from current branch
-        if (branch.startsWith('* ')) {
-          return branch.substring(2);
-        }
-        // Remove remotes/ prefix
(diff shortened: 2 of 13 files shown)


b0x42:
@coderabbitai review again

b0x42:
@coderabbitai review again. I fixed "Prefer explicit name props on the credential inputs.", but fixing "Add a render-level regression test for these autofill attributes." would require to setup a testing framework. More details: 

> The project has no test framework installed — no vitest, jest, or @testing-library/react in package.json, and no existing  
>   test files anywhere in src/. There is nowhere to place this test and no runner to execute it.                              
>                                                                                                                              
>   The finding cannot be applied as-is. To do it properly the project would need:                                             
>   1. A test framework added (vitest + @testing-library/react + jsdom)                                                        
>   2. Vite/TS config updated to include test globals                                                                          
>   3. A test script added to package.json                                                                                     
>                                                                                                                            
>   That's non-trivial infrastructure work that goes well beyond the scope of this PR. I'd recommend either:
>   - Skip this finding — the props are explicitly set and lint/TypeScript already catch prop-type mismatches at the call site.
>   - Open a separate issue to add a test harness to the project and cover auth components as a first batch.

b0x42:
Ready for Review cc @viper151

PR #560, opened 2026-03-19
feat: clean conversation view with noise stripping and tool grouping

rtudo:
## Summary

- Adds a **clean view** toggle (default: on) in Quick Settings that collapses noise in conversation sessions
- **Strips noise tags** (system-reminders, command tags) from user messages at the view layer, preserving raw data for raw mode
- **Groups consecutive read-only tools** (Read, Grep, Glob, etc.) into collapsible "Research" summary bars with expand/collapse
- **Renders skill definition loads** as compact chips instead of full definition text
- Threads `cleanView` preference through MainContent → ChatInterface → ChatMessagesPane

## Architecture

- Skill detection at the **data layer** (`messageTransforms.ts`) — lightweight flag, no data destruction
- Noise stripping + tool grouping at the **view layer** (`cleanViewFilters.ts`, `cleanViewGrouping.ts`) — only when cleanView is enabled
- Two new components: `SkillLoadChip` (pill-shaped skill indicator) and `CollapsedToolGroup` (expandable tool summary)
- Memoized pipeline in `ChatMessagesPane` with `useMemo` keyed on `[visibleMessages, cleanView]`

## Test plan

- [ ] Toggle clean view on/off in Quick Settings panel — verify toggle works and persists
- [ ] Open a session with skill loads — verify they appear as compact chips
- [ ] Open a session with system-reminder tags — verify they are stripped in clean mode, visible in raw mode
- [ ] Open a session with consecutive Read/Grep/Glob calls — verify they collapse into a summary bar
- [ ] Click a collapsed group — verify it expands to show individual tools
- [ ] Verify Edit/Write/Bash tools remain fully visible (not collapsed)
- [ ] Verify `npx tsc --noEmit` and `npm run build` pass

## Summary by CodeRabbit

* **New Features**
  * Clean View toggle (default on): collapses consecutive read-only tool calls, shows compact skill-load chips, and strips view-layer noise while preserving raw mode; persisted via quick settings.
  * New slash commands: /compact, /export, /context.
  * Provider integration toggles (Cursor, Codex, TaskMaster) in Settings.

* **Performance**
  * Improved scroll responsiveness via throttled scroll updates.

* **Documentation**
  * Added Clean View design/plan and Session Context Tracking spec.

* **Chores**
  * Server auto-refresh/watch disabled; project refresh now manual or via API.

### DIFF
--- a/docs/superpowers/plans/2026-03-19-clean-conversation-view.md
+++ b/docs/superpowers/plans/2026-03-19-clean-conversation-view.md
@@ -1,0 +1,829 @@
+# Clean Conversation View Implementation Plan
+
+> **For agentic workers:** REQUIRED: Use superpowers:subagent-driven-development (if subagents available) or superpowers:executing-plans to implement this plan. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add a clean/raw view toggle to the chat session viewer that collapses noise (skill definitions, system reminders, read-only tools) while preserving code-change tool visibility.
+
+**Architecture:** The feature threads a `cleanView` boolean through the existing preference system (`useUiPreferences`). Noise stripping and skill detection happen at the **view layer** (in `ChatMessagesPane`), not at the data transform layer — this preserves raw data for raw mode. A grouping layer in `ChatMessagesPane` collapses consecutive read-only tools. Two small new components render the collapsed group and skill chip.
+
+**Tech Stack:** React 18, TypeScript, Tailwind CSS, Lucide icons
+
+**Spec:** `docs/superpowers/specs/2026-03-18-clean-conversation-view-design.md`
+
+---
+
+## File Structure
+
+| File | Action | Responsibility |
+|------|--------|---------------|
+| `src/hooks/useUiPreferences.ts` | Modify | Add `cleanView` preference (default: `true`) |
+| `src/components/chat/types/types.ts` | Modify | Add `isSkillLoad`, `skillName` to `ChatMessage` |
+| `src/components/chat/utils/messageTransforms.ts` | Modify | Skill detection (always, lightweight — sets `isSkillLoad` flag) |
+| `src/components/chat/utils/cleanViewGrouping.ts` | Create | Group consecutive read-only tools for clean view |
+| `src/components/chat/view/subcomponents/SkillLoadChip.tsx` | Create | Compact skill indicator chip |
+| `src/components/chat/view/subcomponents/CollapsedToolGroup.tsx` | Create | Expandable summary for grouped read-only tools |
+| `src/components/chat/utils/cleanViewFilters.ts` | Create | Noise stripping functions (applied at view layer) |
+| `src/components/chat/view/subcomponents/ChatMessagesPane.tsx` | Modify | Wire `cleanView` prop, apply noise stripping + grouping, render groups |
+| `src/components/chat/view/ChatInterface.tsx` | Modify | Pass `cleanView` through |
+| `src/components/main-content/view/MainContent.tsx` | Modify | Extract `cleanView` from preferences, pass to ChatInterface |
+| `src/components/quick-settings-panel/constants.ts` | Modify | Add toggle entry for `cleanView` |
+| `src/components/quick-settings-panel/types.ts` | Modify | Add `cleanView` to preference key union |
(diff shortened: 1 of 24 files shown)


rtudo (commented):
## Code Review: PR #560 — Clean Conversation View

### Overall

Good feature — clean view makes session audit much more usable. Architecture is sound: skill detection at data layer (lightweight flag, no data destruction), noise stripping + grouping at view layer (preserving raw mode). The plan-driven approach shows.

However, this PR bundles **5 unrelated features** into one. Consider splitting for easier review/bisection in future:
1. Clean conversation view (the main feature)
2. Chokidar watcher removal (CPU fix)
3. New slash commands (`/compact`, `/export`, `/context`)
4. Provider enable/disable system
5. Scroll perf optimization (rAF throttle)

---

### Issues

**1. `CollapsedToolGroup` — missing `useMemo` on `summarizeToolGroup`** (Minor)
`CollapsedToolGroup.tsx:34` calls `summarizeToolGroup(tools)` on every render. Since this iterates tools and builds strings, wrap it:
```tsx
const summary = useMemo(() => summarizeToolGroup(tools), [tools]);
```

**2. `onGrantToolPermission` type mismatch** (Minor)
`CollapsedToolGroup.tsx:13` declares:
```ts
onGrantToolPermission?: (suggestion: { entry: string; toolName: string }) => { success: boolean };
```
But `MessageComponent.tsx:29` expects:
```ts
onGrantToolPermission?: (suggestion: ClaudePermissionSuggestion) => PermissionGrantResult | null | undefined;
```
This works structurally but is inconsistent. Import and use the same types as `MessageComponent` for correctness.

**3. `api.js` — exports before imports** (Minor)
`src/utils/api.js:1-35` places provider utility exports *before* `import { IS_PLATFORM }` at line 37. While JS hoisting makes this work, it's unconventional and will confuse contributors. Move the provider utilities below the imports block.

**4. Default-disabled providers is a breaking change for existing users** (Important)
`api.js:4`: `DEFAULT_DISABLED = ['cursor', 'codex', 'taskmaster']` means **new installs** silently disable Cursor/Codex/TaskMaster. Existing users who haven't set `disabled-providers` in localStorage will also get these disabled on upgrade. Consider:
- Either default to empty (all enabled) and let users opt-out
- Or add a migration: if `disabled-providers` key doesn't exist in localStorage, don't apply defaults

**5. `prevMessage` lookup is O(n) per item in render** (Minor)
`ChatMessagesPane.tsx:289-296` walks backwards through `cleanViewItems` for each message item. For large sessions this is O(n²). Pre-compute the `prevMessage` map during the `useMemo` pass instead.

**6. `setupProjectsWatcher` is now a no-op** (Cleanup)
`server/index.js:93-94` — the function just logs and returns. Either remove it entirely and its callsite, or document that it's kept as a placeholder for future reimplementation.

**7. No unit tests for pure utility functions** (Important)
`cleanViewFilters.ts`, `cleanViewGrouping.ts`, and skill detection in `messageTransforms.ts` are pure functions — ideal test targets. The grouping logic (group breaks, edge cases, single-tool groups) and noise stripping (partial removal, full removal, nested tags) would benefit from tests, especially since bugs here silently hide user content.

---

### Strengths

- **View-layer filtering is the right call** — raw mode gets unmodified data, clean mode transforms at render time
- **Memoized pipeline** in `ChatMessagesPane` with correct deps `[visibleMessages, cleanView]`
- **Pluralization fix** (final commit) — explicit singular/plural pairs instead of naive suffix appending
- **Scroll rAF throttle** with proper cleanup on unmount
- **Skill detection** is lightweight (string prefix check + regex) and doesn't destroy original content

### Verdict

The clean view feature itself is solid and well-architected. The main concerns are the type mismatch (#2), the default-disabled providers breaking change (#4), and the lack of tests (#7). Items #1, #3, #5, #6 are minor cleanup.

rtudo (commented):
**Update:** Items #1 and #2 from my review have been addressed:

- ✅ `summarizeToolGroup` now wrapped in `useMemo` in `CollapsedToolGroup`
- ✅ `onGrantToolPermission` type now uses shared `ClaudePermissionSuggestion` / `PermissionGrantResult` types, matching `MessageComponent`

Remaining items: #3 (exports before imports in api.js), #4 (default-disabled providers breaking change), #5 (prevMessage O(n) lookup), #6 (no-op setupProjectsWatcher), #7 (no unit tests).

PR #604, opened 2026-03-29
fix: line breaks for adding new environment variables

CSchulz:
Fixes #600 

This implementation adds an empty entry for a line break.

```
API_KEY=1234
=
```

Known issues:
If you paste env variables with a line break at the end, it would create a new line.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved environment variable parsing in Claude MCP settings to more reliably handle edge cases involving trailing whitespace and duplicate variable names.


blackmammoth:
Fixed in #666

PR #615, opened 2026-04-04
fix: stop intercepting plain c in auth shell

CoderLuii:
## Summary
- Keeps plain `c` and uppercase `C` as terminal input while an auth URL is active.
- Moves the auth URL copy shortcut to `Ctrl/Cmd+Shift+C`.
- Adds the visible hint requested in review for both desktop and mobile minimal auth flows.

## Context
This fixes the same failure seen downstream in CoderLuii/HolyClaude#32: while the auth URL was active, the shell copied the URL on plain lowercase `c` and swallowed the typed character. Claude can still print `c to copy` in its own prompt, so the UI now says what CloudCLI actually supports.

## Verification
- `npm ci`
- `npm run typecheck`
- `npm run lint` exited 0 with existing warnings.
- `npm run build` exited 0 with existing Vite/Browserslist/CSS warnings.
- Source check covered plain `c`, uppercase `C`, `Ctrl/Cmd+Shift+C`, selected-text copy, paste, and the desktop/mobile helper text.


CoderLuii:
Superseded by the June 15 update below. The branch has been rebased, the PR body now reflects the current shortcut, and the visible auth-copy guidance is included.

blackmammoth:
Hey @CoderLuii, this is great! However, we should include some description that the user should use CTRL/CMD+SHIFT+C for copying especially since claude itself expresses use `c to copy` (look at the red circled part of the image)

image

CoderLuii:
@blackmammoth you’re right. Sorry for the noisy earlier update, and good catch on the missing explanation.

This came from a HolyClaude report where plain lowercase `c` was being swallowed while the auth URL was active. The important part is that normal typing should win, so plain `c` and uppercase `C` now stay as terminal input.

I updated the shortcut to `Ctrl/Cmd+Shift+C`, kept normal selected-text copy working, and added visible text for the mismatch with Claude’s `c to copy` prompt.

I also cleaned up the PR body and rechecked the auth prompt path before taking this out of draft.

PR #616, opened 2026-04-04
fix: dedupe realtime messages before render

CoderLuii:
## Summary
- Upstreams the HolyClaude #29 realtime chat dedupe fix into the session store.
- Realtime rows now merge by stable row/sequence identity instead of appending blindly.
- Thinking rows without row/sequence identity also use a narrow same-session, same-text, short-window guard.
- Server-backed realtime rows are skipped before they can count against the capped realtime buffer.
- The dedupe logic avoids broad `requestId` or `toolId` matching, so separate tool and permission events stay separate.

## Context
This came from CoderLuii/HolyClaude#29, reported by @mikkelnl. The UI rendered duplicate sent user rows and duplicate Thinking blocks while Claude answered once.

HolyClaude shipped the downstream image fix in v1.1.9. This PR upstreams the same class of session-store fix into claudecodeui.

## Verification
I refreshed dependencies with `npm ci`, then ran typecheck, lint, build, and the targeted realtime/session-store check.

Commands:
- `npm ci`
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `node --import tsx --test src/stores/sessionMessageDedupe.test.ts`
- `git diff --check`

The targeted check covers duplicate user rows, duplicate Thinking rows with and without row/sequence identity, older same-text Thinking rows staying separate, different-sequence Thinking rows staying separate, server-backed realtime rows, capped-buffer behavior, and distinct tool/permission events staying separate.

Not live-tested: full current-main UI reproduction.


CoderLuii:
Superseded by the refreshed PR update posted on June 15, 2026.

The old branch head in this comment was `7e75eb9`. The PR has since been rebased against current main and replaced by `480e7d4`.

Current update: https://github.com/siteboon/claudecodeui/pull/616#issuecomment-4711066243

blackmammoth:
Hey @CoderLuii, there has been a ton of changes since this PR. Can you check if the PR is still relevant? If so, can you share a detailed reproduction step for the issue this PR attempts to fix.

CoderLuii:
@blackmammoth yes, this is still relevant, and I refreshed the PR against current main.

Sorry for the earlier update being too thin and for leaving the noisy release-notes block in the PR body. I should have included the reproduction trail when I refreshed the branch.

This PR came from CoderLuii/HolyClaude#29, reported by @mikkelnl. The report was not that Claude answered twice. The UI rendered the same flow twice: each sent user message appeared twice with about a one-second timestamp gap, and the Thinking state appeared twice as two response blocks. The actual Claude response appeared once.

The reported setup was HolyClaude full/latest on a Raspberry Pi 5 ARM64 host, clean data directory, direct local IP access, Safari and Chrome both reproducing it. The reproduction was: deploy the image, open the web UI, create a CloudCLI account/session, send any message, then observe duplicate user rows and duplicate Thinking rows.

HolyClaude shipped the downstream fix in v1.1.9 because users were hitting it in the image. That release vendored a patched CloudCLI build and called out the realtime chat dedupe fix in the release notes. This PR now upstreams that same class of fix into claudecodeui.

What changed in this refresh:

I rebased the branch onto current main and removed the noisy release-notes block from the PR body.

The fix stays in the session store. Realtime messages now upsert against stable identities, and messages already represented by server history are skipped before they can count against the realtime buffer.

I also covered the weaker Thinking shape from the HolyClaude report: same-session, same-text Thinking rows about a second apart now collapse even when neither row has `rowid` or `sequence`. That fallback does not run when either row has `rowid` or `sequence`, and it stays away from broad `requestId` or `toolId` matching, so separate tool and permission events do not get collapsed.

I refreshed dependencies with `npm ci`, then ran typecheck, lint, build, and the targeted realtime/session-store check. The commands completed successfully.

The targeted check covers duplicate user rows, duplicate Thinking rows with and without row/sequence identity, older same-text Thinking rows staying separate, different-sequence Thinking rows staying separate, server-backed realtime rows being skipped before they count against the buffer, capped-buffer behavior, and distinct tool/permission events staying separate.

Current commit: `480e7d4`

blackmammoth:
Hey @CoderLuii, this should be fixed with the latest releases. Lmk if the issue is still recurrent tho.

Issue #576, opened 2026-03-23
"Delete Project" destroys Claude Code memory & session data — misleading UX causes permanent data loss


WhymustIhaveaname:
## Bug Summary

The "Delete Project" button in the sidebar **permanently deletes all Claude Code project data** from `~/.claude/projects/{projectName}/`, including:

- **Project memories** (`memory/MEMORY.md` and all linked memory files)
- **Session history** (all `.jsonl` conversation logs)
- **Subagent data**

Users reasonably expect this button to **remove the project from the UI list**, not to `rm -rf` their actual Claude Code data. This is a critical UX mismatch.

## Steps to Reproduce

1. Use Claude Code CLI to build up project memories over multiple sessions
2. Open claudecodeui, see the project in the sidebar
3. Click the trash icon (Delete Project) — confirm when prompted
4. Open a new Claude Code session in the same project directory
5. All project memories are gone. Permanently.

## Root Cause

`server/projects.js` line ~1200:

```js
await fs.rm(projectDir, { recursive: true, force: true });
// projectDir = ~/.claude/projects/{projectName}
```

This recursively deletes the **entire** project directory, which is shared infrastructure owned by Claude Code CLI — not by this UI.

## Impact

- **Permanent data loss** — Claude Code memories are not backed up and cannot be recovered
- **Silent destruction** — the confirmation dialog mentions "sessions" but not "memories", giving no indication that CLI-created data will be destroyed
- **Cross-tool damage** — this UI is deleting data created and managed by a different tool (Claude Code CLI)

## What Happened to Me

I had project memories documenting Telegram plugin startup procedures, known bugs, and deployment configurations. I clicked "Delete Project" thinking it would just clean up the UI. Everything was permanently wiped. I had to forensically reconstruct memories from session `.jsonl` logs in an unrelated project directory.

## Suggested Fix

**Option A (minimal):** Only delete the UI's own metadata (the entry in `projects.json`). Leave `~/.claude/projects/` untouched.

**Option B (safe delete):** If deleting session data is intended, at minimum:
- Clearly warn: *"This will permanently delete all Claude Code project memories and session history from disk"*
- Separate "Remove from UI" vs "Delete all data"
- Never delete the `memory/` subdirectory — it belongs to the user's Claude Code workflow

**Option C (ideal):** Move to trash / soft-delete instead of `rm -rf`.

## Environment

- claudecodeui: latest (as of 2026-03-23)
- Claude Code CLI: 2.1.81
[...]

PR #620, opened 2026-04-05
fix(projects): allow removing manually added projects

Steve-Rye:
## Summary

This PR adds a safe `Remove` flow for manually added projects so they can be removed from the sidebar without deleting the underlying project directory or related session data.

## What changed

- add `DELETE /api/projects/:projectName/remove` for eligible manually added projects
- keep `Delete Project` behavior unchanged for projects with local Claude history
- show `Remove` instead of `Delete` for manually added sidebar-only projects
- add separate confirmation copy that explicitly states files and sessions are not deleted
- add a focused backend regression test for `removeProjectFromList`
- add locale strings for the new sidebar action and confirmation copy

## Why

Users can add existing workspaces to the project list, but the reverse action was missing. The only available action was `Delete Project`, which is much heavier and can delete local data. For manually added projects that only exist as sidebar entries, users need a way to remove them from the list without treating them as disposable data.

## Scope

This PR intentionally only adds `Remove` for projects that are:

- manually added via config
- not backed by a local Claude project directory in `~/.claude/projects`

Projects with local Claude history still use the existing delete flow because they are auto-discovered again from disk and cannot be truly hidden by removing config alone.

## Testing

- [x] `node --test server/projects.remove-project.test.mjs`
- [x] `npm run typecheck`
- [x] `npm run build`

## Related

- Related to #576

## Summary by CodeRabbit

* **New Features**
  * Added ability to remove manually added projects from the sidebar without deleting local history or sessions.
  * Distinct UI and messaging for removing projects versus full deletion.

* **Tests**
  * Added test coverage for project removal functionality.


blackmammoth:
Hey @Steve-Rye, this has been implemented with the latest release.

PR #633, opened 2026-04-10
fix: preserve Shell tab scroll position across periodic refresh

CoderLuii:
## Summary

- Preserve the Shell terminal viewport when the existing focus callback runs.
- Keep focus scheduling, connection handling, refresh behavior, and auto-connect unchanged.
- Move the downstream HolyClaude #35 vendored-bundle patch into source.

## Context

CoderLuii/HolyClaude#35 reported the Shell tab losing scroll position during the periodic browser refresh in the HolyClaude full image on Linux. The downstream reproduction was: open HolyClaude, go to the Shell tab, scroll inside a long shell session, wait for the periodic refresh, and the viewport jumps away from the line being read.

HolyClaude v1.2.1 fixed that downstream by patching the vendored CloudCLI bundle to preserve `buffer.active.viewportY` around the Shell `focus()` call. Current CloudCLI source still focuses the terminal without preserving the viewport, so this PR applies the same behavior in `src/components/shell/view/Shell.tsx`.

## Testing

- `git diff --check`
- `npm ci` completed; npm audit reported 39 existing advisories
- `npm run typecheck`
- `npm run build` passed with existing Browserslist/CSS/chunk-size warnings
- `npm run lint` passed with 0 errors and existing repo-wide warnings
- Built app smoke: login completed with a disposable local account, a test project opened, and the Shell tab mounted with xterm focused

### DIFF
--- a/src/components/shell/view/Shell.tsx
+++ b/src/components/shell/view/Shell.tsx
@@ -160,7 +160,11 @@ export default function Shell({
     }
 
     const focusTerminal = () => {
-      terminalRef.current?.focus();
+      const term = terminalRef.current;
+      if (!term) return;
+      const savedViewport = term.buffer.active.viewportY;
+      term.focus();
+      term.scrollToLine(savedViewport);
     };
 
     const animationFrameId = window.requestAnimationFrame(focusTerminal);
--- a/src/components/shell/view/Shell.tsx
+++ b/src/components/shell/view/Shell.tsx
@@ -165,7 +165,11 @@ export default function Shell({
     }
 
     const focusTerminal = () => {
-      terminalRef.current?.focus();
+      const term = terminalRef.current;
+      if (!term) return;
+      const savedViewport = term.buffer.active.viewportY;
+      term.focus();
+      term.scrollToLine(savedViewport);
     };
 
     const animationFrameId = window.requestAnimationFrame(focusTerminal);


CoderLuii:
Superseded by the refreshed PR branch and the newer maintainer reply below. The branch has now been rebased onto current `main`, the PR body was cleaned up, and the latest verification is recorded below.

blackmammoth (commented):
Hey @CoderLuii, thanks for the PR. Can you send me a detailed reproduction step for this issue? I tried this both on Windows and Android (using Chrome) and I didn't notice a scroll issue with the shell.

CoderLuii:
@blackmammoth thanks for checking this. I traced the report back to CoderLuii/HolyClaude#35. That downstream report was for the HolyClaude full image on Linux: open the Shell tab, scroll inside a long shell session, wait for the periodic browser refresh, and the terminal viewport jumps away from the line being read.

HolyClaude v1.2.1 fixed it downstream by patching the vendored CloudCLI bundle to preserve `buffer.active.viewportY` around the Shell `focus()` call. This PR applies the same fix at source in `Shell.tsx`, so the terminal can still regain focus without forcing the viewport away from the line the user was reading.

I refreshed the branch against current `main`, cleaned up the PR body, and re-ran `git diff --check`, `npm ci`, `npm run typecheck`, `npm run build`, and `npm run lint`. I also opened the built app locally and confirmed the Shell tab mounts with the xterm input focused. The change is intentionally small: it does not change connection handling, refresh behavior, or auto-connect logic. It only preserves the terminal viewport around the existing focus call.

Ready for another look.

blackmammoth:
@CoderLuii tested on a linux environment as well and there appears to be no issues. As a result, I'm closing this.

PR #636, opened 2026-04-10
refactor: removed the token calculator logic

blackmammoth:
The information given was not accurate for the model providers. We can add it later on when we can get accurate responses from the sdks/spawned processes.

## Summary by CodeRabbit

* **Removed Features**
  * Token usage tracking and visual indicators have been removed from the chat UI and composer.
  * The built-in /cost command has been removed.
  * Non-streaming API responses and session/history endpoints no longer include token usage or token-budget data; streaming messages no longer emit token-usage/status updates.
* **Documentation**
  * Public API docs updated to omit token accounting references.


viper151:
@blackmammoth  this removes more than it should actually. The tokens being returned from at least claude code are accurate. the visualization we have indeed is not relevant anymore. But the tokens are coming straight out of the sdk.

PR #638, opened 2026-04-10
Fix/replace all spawn with cross spawn

blackmammoth:
## Summary by CodeRabbit

* **Chores**
  * Standardized subprocess launching across server modules to use a single, consistent implementation for improved cross-platform reliability and more predictable behavior when running CLI tools and background processes.


viper151 (changes requested):
@blackmammoth  can you look the coderabbit comments?

PR #643, opened 2026-04-11
feat: auto-name new sessions and support API-key renames

walksoda:
## Summary

Today, newly created Claude sessions are labelled by a truncated slice of the first user message. Once the sidebar fills up with dozens of sessions that all start with "help me fix...", it gets hard to tell them apart, and users end up either hand-renaming every session they want to revisit or scrolling through low-signal titles.

This PR auto-generates a short, descriptive title for each new session from the first user/assistant exchange, so session lists stay scannable without manual work. Manual rename continues to work exactly as before and always wins.

Example of a session that used to look like:

```
help me debug why websocket reconnect loops after deploy on sta...
```

now shows up as:

```
Debug websocket reconnect loop after deploy
```

As a small related change, the rename endpoint now also accepts `x-api-key` auth so API-driven clients (scripts, the agent API) can apply their own titles through the same endpoint.

## Why this is useful

- Long-running workspaces accumulate many similar-looking sessions; auto-naming makes the sidebar immediately scannable
- Removes the "rename every useful session manually" friction
- Keeps manual rename as the source of truth when you do want a custom title
- Opens the rename endpoint to API-key clients, enabling scripted labelling of sessions created via `/api/agent/launch`

## Guardrails

- **Opt-in**: auto-naming only runs when the caller passes `autoNameSession: true`. The chat WebSocket path and `/api/agent/launch` opt in; `git.js` and other non-interactive `queryClaudeSDK` callers are untouched.
- **Cheap and fire-and-forget**: one call to the `haiku` model per new session, based only on the first exchange, with a 30s timeout. Failures are logged and swallowed — they never affect the chat itself.
- **Manual rename always wins**: the module re-checks `sessionNamesDb.getName()` both before and after the LLM call, so a user rename during the in-flight window is never clobbered.
- **Lightweight update path**: instead of triggering a full `getProjects()` rescan on every auto-naming event, the server emits a small `session_name_updated` WebSocket message and the frontend patches its local state in place.

## Implementation notes

- New `server/session-naming.js` generates and persists titles to the existing `session_names` SQLite table. No schema change.
- Extracted `connectedClients` / `broadcastProgress` / `broadcastSessionNameUpdated` into `server/utils/websocket-clients.js` so both the chat WebSocket handler and `/api/agent/launch` can share them.
- `PUT /api/sessions/:sessionId/rename` is registered **before** the global `/api/sessions` JWT-only mount so its dual JWT-or-`x-api-key` handler runs first. Header-only (no query-string API key).

## Known limitation

Authorization for the rename endpoint is unchanged from the current upstream behaviour: any authenticated caller who knows a session ID can rename that session. Adding per-user ownership checks would require storing ownership metadata on `session_names` (schema migration), which felt better scoped as a follow-up PR.

## Test plan

- [ ] Start server; a new chat session gets an auto-generated title shortly after the first assistant reply
- [ ] The session label updates in place from a `session_name_updated` WebSocket event — no full project rescan
- [ ] Manual rename during an in-flight auto-name call is preserved (re-check path)
- [ ] Non-interactive `queryClaudeSDK` callers (e.g. `git.js`) do **not** trigger auto-naming
- [ ] `PUT /api/sessions/:id/rename` with JWT `Authorization` header → 200
- [ ] `PUT /api/sessions/:id/rename` with `x-api-key` header → 200
- [ ] `npx vite build` succeeds

## Summary by CodeRabbit

* **New Features**
  * Automatic session naming from the assistant’s first response, producing concise human-readable titles.
  * Real-time broadcast of session-name updates so connected clients show new names immediately.

* **Improvements**
  * Session rename API now supports alternate API-key auth and enforces stronger validation (required, trimmed, ≤500 chars).
  * Auto-naming respects manual names and avoids duplicate generation under concurrency; broadcasts are more reliable.

### DIFF
--- a/server/claude-sdk.js
+++ b/server/claude-sdk.js
@@ -27,6 +27,7 @@ import {
 import { claudeAdapter } from './providers/claude/adapter.js';
 import { createNormalizedMessage } from './providers/types.js';
 import { getStatusChecker } from './providers/registry.js';
+import { generateAndPersistSessionName } from './session-naming.js';
 
 const activeSessions = new Map();
 const pendingToolApprovals = new Map();
@@ -476,6 +477,7 @@ async function queryClaudeSDK(command, options = {}, ws) {
   let sessionCreatedSent = false;
   let tempImagePaths = [];
   let tempDir = null;
+  let firstAssistantResponse = null;
 
   const emitNotification = (event) => {
     notifyUserIfEnabled({
@@ -663,6 +665,14 @@ async function queryClaudeSDK(command, options = {}, ws) {
         ws.send(msg);
       }
 
+      // Capture first assistant response for session naming
+      if (!firstAssistantResponse && message.type === 'assistant' && message.message?.content) {
+        const c = message.message.content;
+        firstAssistantResponse = Array.isArray(c)
+          ? c.find(p => p.type === 'text')?.text
+          : typeof c === 'string' ? c : null;
+      }
+
       // Extract and send token budget updates from result messages
       if (message.type === 'result') {
         const models = Object.keys(message.modelUsage || {});
@@ -694,6 +704,14 @@ async function queryClaudeSDK(command, options = {}, ws) {
       stopReason: 'completed'
     });
     // Complete
+
+    // Fire-and-forget: auto-name new sessions (opt-in only — chat sessions pass this flag)
+    if (options.autoNameSession && !sessionId && !!command && capturedSessionId) {
+      generateAndPersistSessionName(
+        capturedSessionId, command, firstAssistantResponse,
+        options.broadcastSessionNameUpdated || null
+      ).catch(err => console.error('Session naming failed:', err));
+    }
 
   } catch (error) {
     console.error('SDK query error:', error);
--- a/server/database/db.js
+++ b/server/database/db.js
@@ -490,6 +490,15 @@ const sessionNamesDb = {
     `).run(sessionId, provider, customName);
   },
 
+  setNameIfAbsent: (sessionId, provider, customName) => {
+    const result = db.prepare(`
+      INSERT INTO session_names (session_id, provider, custom_name)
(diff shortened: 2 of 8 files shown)


walksoda:
## Rebased & addressed review feedback

Rebased onto latest `main` (v1.29.5, includes #413 session rename infrastructure). All changes now build on top of #413's `session_names` table and `sessionNamesDb` API — no schema changes.

### Review feedback addressed

| # | Source | Fix |
|---|--------|-----|
| 1 | Copilot | **x-api-key header conflict**: Moved rename endpoint before `validateApiKey` middleware so global `API_KEY` gate doesn't block DB API-key auth |
| 2 | Copilot | **x-api-key array guard**: Reject `Array.isArray(rawKey)` before validation |
| 3 | Copilot | **JSDoc mismatch**: Fixed `projects_updated` → `session_name_updated` |
| 4 | CodeRabbit | **Missing broadcast on manual rename**: Added `broadcastSessionNameUpdated()` after `setName()` so other clients see manual renames in real time |
| 5 | CodeRabbit | **React state no-op guard**: `setProjects` now uses a `changed` flag; `setSelectedProject`/`setSelectedSession` use functional updaters that return `prev` on no-op |
| 6 | Copilot | **TOCTOU race**: Added `sessionNamesDb.setNameIfAbsent()` (`INSERT ... ON CONFLICT DO NOTHING`) — auto-naming now atomically yields to concurrent manual renames |

### Additional fix (found during post-rebase review)

- **Duplicate route registration**: The rebase left two `PUT /api/sessions/:sessionId/rename` handlers (one from #413, one from this PR). Express routes first match, so the #413 version (without broadcast or array guard) was silently winning. Removed the duplicate — only the improved handler remains.

blackmammoth:
Hey @walksoda, there has been a lot of update since this PR. Can you check if this is relevant again? If so, please resolve the merge conflicts and resend the PR.

PR #648, opened 2026-04-11
Security: Authentication token sent in WebSocket query string

tuanaiseo:
## Problem

The shell WebSocket URL appends the auth token as a `?token=` query parameter. Query-string tokens are prone to leakage via logs, monitoring tools, reverse proxies, browser history, and diagnostics. This increases risk of token exposure and session hijacking.

**Severity**: `medium`
**File**: `src/components/shell/utils/socket.ts`

## Solution

Avoid putting bearer tokens in URLs. Prefer secure, HTTP-only session cookies for WS auth, or use a short-lived one-time WS token exchanged over authenticated HTTPS before establishing the socket. Ensure server-side logs redact sensitive parameters.

## Changes

- `src/components/shell/utils/socket.ts` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

## Release Notes

* **Bug Fixes**
  * Updated WebSocket connection behavior to simplify authentication handling; connections no longer include token parameters.


blackmammoth:
Hey @tuanaiseo, thanks for the PR. The token is not used for the hosted platform (cloudcli.ai) since the authentication is already done elsewhere. It's only used for OSS. As a result, this PR breaks OSS shell auth. So, I'm closing this for now. But, I'll remove the logging in server side logic that shows the full token.

PR #649, opened 2026-04-12
feat: add Opus [1M] model with correct 1M context window reporting

### LINKED ISSUE #579: Add Opus [1M] option to model picker
## Feature Request

The model picker currently has a `Sonnet [1M]` option for the 1M context window, but there is no equivalent `Opus [1M]` option.

Claude Opus 4.6 supports a 1M token context window (via the `--model claude-opus-4-6-20251101` flag or equivalent), but this isn't exposed in the UI.

## Proposed Change

Add an `Opus [1M]` entry to `CLAUDE_MODELS.OPTIONS` in `shared/modelConstants.js`, similar to the existing `Sonnet [1M]`:

```js
{ value: "opus[1m]", label: "Opus [1M]" }
```

## Why

For large codebases or long conversations, Opus + 1M context is the most capable option. Right now users have to fall back to the CLI to use it.

bbesner:
## Summary

Closes #579

Adds `opus[1m]` to the Claude model picker and fixes context window reporting so 1M-context models display the correct token budget instead of showing 160K/200K.

## Problem

When using Opus with 1M context (`claude-opus-4-6[1m]`), the token usage pie chart incorrectly showed 160K or 200K as the total context window, making it appear 100% full even with plenty of room remaining.

## Root Cause

The fix required addressing four layers:

1. **Model list** — `opus[1m]` was missing from `CLAUDE_MODELS.OPTIONS`
2. **Server-side token budget** — `extractTokenBudget()` hardcoded `160000` instead of reading `contextWindow` from the SDK's `modelUsage` result (which already reports `1000000` for 1M models)
3. **Frontend fallback** — `TokenUsagePie` used `160000` as fallback before the first server response. Now checks the selected model name for `1m` suffix
4. **Frontend state race** — After the WebSocket correctly sets `total: 1000000`, the HTTP `token-usage` endpoint re-fires and overwrites it with `200000`. Fixed by preventing HTTP from overwriting a higher WebSocket value

## Changes

- `shared/modelConstants.js` — Add `opus[1m]` option
- `server/claude-sdk.js` — Read `contextWindow` from SDK result's `modelUsage` data; pass `selectedModel` for fallback detection
- `src/.../ChatInputControls.tsx` — Accept `selectedModel` prop, use model-aware context window fallback
- `src/.../ChatComposer.tsx` — Thread `selectedModel` prop
- `src/.../ChatInterface.tsx` — Pass `claudeModel` to composer
- `src/.../useChatSessionState.ts` — Prevent HTTP token-usage fetch from overwriting WebSocket value

## Testing

Tested with Claude Code CLI 2.1.92 + `@anthropic-ai/claude-agent-sdk` 0.2.104.

- New session with `opus[1m]` → pie shows 1,000,000 immediately ✓
- After Claude responds → pie stays at 1,000,000 (not reverted to 200K) ✓
- Sessions with `sonnet` → pie correctly shows 200,000 ✓
- Sessions with `sonnet[1m]` → pie correctly shows 1,000,000 ✓

## Note

The SDK (0.2.104) already reports `contextWindow: 1000000` in the `modelUsage` result for 1M models. This PR makes CloudCLI read that value instead of ignoring it. Users on older SDK versions may need to upgrade for the server-side detection to work; the frontend fallback handles the case where the SDK doesn't report `contextWindow`.

## Summary by CodeRabbit

* **New Features**
  * Added Claude Opus [1M] model to available options, providing access to an expanded 1M token context window.

* **Improvements**
  * Enhanced token budget detection with intelligent model-aware fallback logic.
  * Improved token usage state management with cancellation tracking and state preservation across model selections.

### DIFF
--- a/server/claude-sdk.js
+++ b/server/claude-sdk.js
@@ -279,7 +279,7 @@ function transformMessage(sdkMessage) {
  * @param {Object} resultMessage - SDK result message
  * @returns {Object|null} Token budget object or null
  */
-function extractTokenBudget(resultMessage) {
+function extractTokenBudget(resultMessage, selectedModel) {
   if (resultMessage.type !== 'result' || !resultMessage.modelUsage) {
     return null;
   }
@@ -302,9 +302,17 @@ function extractTokenBudget(resultMessage) {
   // Total used = input + output + cache tokens
   const totalUsed = inputTokens + outputTokens + cacheReadTokens + cacheCreationTokens;
 
-  // Use configured context window budget from environment (default 160000)
-  // This is the user's budget limit, not the model's context window
-  const contextWindow = parseInt(process.env.CONTEXT_WINDOW) || 160000;
+  // Determine context window: env override > model usage data > selected model > default
+  let contextWindow = parseInt(process.env.CONTEXT_WINDOW) || 0;
+  if (!contextWindow) {
+    if (modelData.contextWindow && modelData.contextWindow > 0) {
+      contextWindow = modelData.contextWindow;
+    } else if (selectedModel && selectedModel.toLowerCase().includes('1m')) {
+      contextWindow = 1000000;
+    } else {
+      contextWindow = 200000;
+    }
+  }
 
   // Token calc logged via token-budget WS event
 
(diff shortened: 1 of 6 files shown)


viper151:
@bbesner thank you for your effort in this PR. I've added the new models in the context and soon the context window reporting will also be revamped so I'll close this PR for now.

PR #658, opened 2026-04-15
Send complete event on SDK errors to clear loading state

### LINKED ISSUE #457: No feedback when claude or codex gives an error / isnt working
When claude or codex gives an error or isnt working, it just shows processing forever, and the tokens keep increasing making it appear like its doing something when in fact its not.

For example when I had claude and this project running as root in a proxmox LXC, when I would turn on bypass permissions, it would get stuck infinitely at processing and nothing would happen. I had to dig through logs to find claude was returning an error saying you cannot use bypass permissions while running as root user. There is no feedback to the ui when errors occur.

jakeefr:
Fixes #457. Added error state display when the Claude/Codex
subprocess returns a non-zero exit code. Users now see
a clear error message instead of silent failure.

I maintain PRISM (https://github.com/jakeefr/prism), a
post-session diagnostics tool for Claude Code — session
health scoring and CLAUDE.md adherence analysis. Remote
session users managing Claude Code through a web UI are
exactly who'd benefit from post-session diagnostics.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where a client run could remain unresolved after an error; the system now sends a terminating completion event immediately after an error so the UI exits loading/error states reliably.

### DIFF
--- a/server/claude-sdk.js
+++ b/server/claude-sdk.js
@@ -719,11 +719,15 @@ async function queryClaudeSDK(command, options = {}, ws) {
       : error.message;
 
     // Send error to WebSocket
-    ws.send(createNormalizedMessage({ kind: 'error', content: errorContent, sessionId: capturedSessionId || sessionId || null, provider: 'claude' }));
+    const errorSessionId = capturedSessionId || sessionId || null;
+    ws.send(createNormalizedMessage({ kind: 'error', content: errorContent, sessionId: errorSessionId, provider: 'claude' }));
+
+    ws.send(createNormalizedMessage({ kind: 'complete', exitCode: 1, sessionId: errorSessionId, provider: 'claude' }));
+
     notifyRunFailed({
       userId: ws?.userId || null,
       provider: 'claude',
-      sessionId: capturedSessionId || sessionId || null,
+      sessionId: errorSessionId,
       sessionName: sessionSummary,
       error
     });


blackmammoth:
Hey @jakeefr, thanks for the PR. Apologies for the delay in review. There have been a lot of changes introduced the past month. Do you think the issue is still relevant? If so, can you share a detailed reproduction step I can use for testing it.

And, would love to have your library featured on the page :) Let me know if you build a plugin for it made for CloudCLI and I will make sure to add it.

jakeefr:
Hello! The issue might still be relevant.

I'm currently overseas without my PC. I can give thorough reproduction steps and will also make a CloudCLI plugin as well when I return! 

No worries with the delayed response, thanks for the reply!

jakeefr:
I managed to read through this against current main from my laptop (no full local
run yet), and I think the issue is no longer reproducible, so this PR can probably
be closed.

Changes that landed since the issue was filed seem to fix it through a different
route:

1. #558 (a4632dc) added a case 'error' handler in useChatRealtimeHandlers.ts that
calls setIsLoading(false), clears the status and marks the session not processing.
So the error event the catch path already sends is now enough to clear the loading
state, which makes the extra complete event from this PR redundant.

2. #715 (44edf94) wrapped the socket in WebSocketWriter, whose send() checks
readyState before sending. That also covers the hardening CodeRabbit flagged here.

I haven't been able to run it end to end yet, so the above is from reading the
code rather than a live repro. Steps to confirm once someone's at a machine:

1. git clone https://github.com/siteboon/claudecodeui.git && cd claudecodeui && npm install && npm run dev
2. Open the UI, select a project, send a prompt that runs for a while, e.g. "Read
every file in this repo and summarize each one". Wait for the Processing spinner.
3. In a second terminal, kill the claude child process of the server (find it with
ps aux | grep claude; pkill -f 'claude.*stream-json' usually matches it).
4. With the old bug the spinner stayed on Processing forever with tokens counting
up. On current main you should get an error message in the chat and the input
re-enabled.

For the exact scenario from #457: run the server as root (e.g. in a container)
with bypass permissions enabled. The CLI exits immediately with "cannot use
--dangerously-skip-permissions as root", and the UI should now surface that error
instead of spinning.

One small leftover if it's useful: the error case in useChatRealtimeHandlers.ts
doesn't clear pendingPermissionRequests the way the complete case does, so an
error that lands while a permission prompt is open leaves a stale prompt on
screen. Happy to send a small PR for that if you'd like.

jakeefr:
Back at my machine, so here is the plugin: https://github.com/jakeefr/cloudcli-plugin-prism

It adds a PRISM tab to CloudCLI. With a project selected you get its session
health grade, scores across five dimensions (token efficiency, tool health,
context hygiene, CLAUDE.md adherence, session continuity), and the top issues
PRISM found. With no project selected it shows an overview of every project,
worst first.

Install is the standard flow: Settings > Plugins, paste the repo URL. The only
requirement is PRISM itself on PATH (pip install prism-cc). Built from the plugin
starter template, frontend plus a small backend that shells out to the prism CLI.

Would be great to see it on the page. Happy to adjust anything that doesn't fit
your conventions.

blackmammoth:
Hey @jakeefr, sent a [PR](https://github.com/siteboon/claudecodeui/pull/848) for your plugin. As a sidenote, I tested this on windows and all I got was the following. Please recheck that it's working correctly.

image

jakeefr:
Hey @blackmammoth, thanks for the PR and the heads up. That 400 was a Windows-only bug on my side: prism installs as a .cmd shim and the backend was using execFile, which can't launch .cmd, so it died before running anything. Switched it to cross-spawn, and also made projects with no sessions show an empty state instead of erroring. Just pushed the fix, so pull latest and try it again, it should work now.

blackmammoth:
hey @jakeefr, unfortunately it's still the same.

jakeefr:
odd. It works on my end. I was hitting RPC error 400 but resolved it. Let me look into it a bit more.

jakeefr:
Hey @blackmammoth, I want to pin this down instead of guessing again, since it works on my end. A few things next time you hit it:

  1. Real status code. DevTools → Network, click the failing …/api/plugins/prism/rpc/… request -still 400, or now
  502/503? After my change /report can't return 400 anymore, so a true 400 means the old build is still running (see
  #2).

  2. How you updated. In-app Update button, or manual git pull? A plain pull leaves the old compiled dist/ and old
  subprocess in place, fix won't take. If so, an uninstall -> reinstall (or disable+enable) is the clean way to be
  sure.

  3. Host console logs. Opening the tab, does the claudecodeui console print any [Plugin:prism] … or [Plugins] Failed
  to start … / exited with code N before reporting ready lines? That's my backend's stderr, tells us exactly where
  it dies.

  4. prism on PATH. From the shell that launches claudecodeui:
  where prism
  prism --version
  Want to know if it's prism.exe or prism.cmd, and whether the host sees it at all.

  One note: plugin-process-manager.js launches the plugin with a restricted env (PATH, HOME, NODE_ENV, PLUGIN_NAME). On Windows that drops SystemRoot, USERPROFILE, PATHEXT - which a Python CLI like prism needs to start and find ~/.claude. Could make prism fail regardless of my spawn change. The #3 logs will confirm; if so I'll harden the env on my side.

  Thanks again.

blackmammoth:
> Hey @blackmammoth, I want to pin this down instead of guessing again, since it works on my end. A few things next time you hit it:
> 
> 1. Real status code. DevTools → Network, click the failing …/api/plugins/prism/rpc/… request -still 400, or now
>    502/503? After my change /report can't return 400 anymore, so a true 400 means the old build is still running (see
>    [Dev #2](https://github.com/siteboon/claudecodeui/pull/2)).
> 2. How you updated. In-app Update button, or manual git pull? A plain pull leaves the old compiled dist/ and old
>    subprocess in place, fix won't take. If so, an uninstall -> reinstall (or disable+enable) is the clean way to be
>    sure.
> 3. Host console logs. Opening the tab, does the claudecodeui console print any [Plugin:prism] … or [Plugins] Failed
>    to start … / exited with code N before reporting ready lines? That's my backend's stderr, tells us exactly where
>    it dies.
> 4. prism on PATH. From the shell that launches claudecodeui:
>    where prism
>    prism --version
>    Want to know if it's prism.exe or prism.cmd, and whether the host sees it at all.
> 
> One note: plugin-process-manager.js launches the plugin with a restricted env (PATH, HOME, NODE_ENV, PLUGIN_NAME). On Windows that drops SystemRoot, USERPROFILE, PATHEXT - which a Python CLI like prism needs to start and find ~/.claude. Could make prism fail regardless of my spawn change. The #3 logs will confirm; if so I'll harden the env on my side.
> 
> Thanks again.

Hey @jakeefr, this is what I'm seeing.

1. Verified that prism exists with `prism --version`
2. Opened prism plugin tab
3. Saw the following console log error `api.js:18  GET http://localhost:5173/api/plugins/prism/rpc/health 400 (Bad Request)`
In the details of the error, I see the following:
```
{
    "error": "Traceback (most recent call last):\r\n  File \"<frozen runpy>\", line 198, in _run_module_as_main\r\n  File \"<frozen runpy>\", line 88, in _run_code\r\n  File \"C:\\...3\\Scripts\\prism.exe\\__main__.py\", line 4, in <module>\r\n    from prism.cli import app\r\nModuleNotFoundError: No module named 'prism'",
    "notInstalled": false
}
```

jakeefr:
Hey @blackmammoth, got it, that traceback was the key. The 400 wasn't the spawn path. prism is a .exe, which launches fine. It was the env. claudecodeui starts plugin subprocesses with a stripped environment (PATH/HOME/NODE_ENV/PLUGIN_NAME), and on Windows that drops APPDATA. Your prism is a pip install --user, so its package lives in the per-user site-packages, and CPython needs APPDATA to find that path. Without it prism.exe launches but can't import prism. That's why prism --version works in your shell (which still has APPDATA) but not under the host.

Fixed in the plugin backend: it now restores APPDATA, USERPROFILE, SystemRoot, and PATHEXT from your real profile
before invoking prism. Doesn't touch prism itself. Restoring USERPROFILE also means reports actually populate on
Windows instead of coming back empty.

One thing: please update via the in-app Update button (or uninstall then reinstall), not a manual git pull. The
host only rebuilds dist/ and restarts the plugin process on the Update flow. A raw pull keeps the old compiled
server running, so you'd still see the 400.

Could possibly be one more failure if it still 400s after the update. If so, it most likely means the Update didn't actually rebuild and restart, so the old compiled server is still running. Quickest tell is the host console: the new code won't print that ModuleNotFoundError anymore, so if you still see it, it's stale. A clean uninstall then reinstall clears that. If the error text in the console is now something different, that's a separate issue and the message will point right at it. Let me know what happens!

Thanks for your help and patience!

jakeefr:
Also, just a suggestion; Unrelated to the plugin fix, but worth flagging for any Windows plugin: plugin-process-manager.js launches plugin servers with env: { PATH, HOME, NODE_ENV, PLUGIN_NAME }. On Windows that omits SystemRoot, USERPROFILE, APPDATA, LOCALAPPDATA, and PATHEXT, which Node and especially any Python/CLI child needs to bootstrap (locate user site-packages, temp, etc.). Passing those through on win32 would save every plugin from re-deriving them. 

Happy to send a PR if useful.

blackmammoth:
Thanks @jakeefr, it's working now. Yeah would love to see a PR for this. For your plugin, you can share it in our discord in the "showcase" channel so that more people can check out your tool.

Link: https://discord.gg/buxwujPNRE

jakeefr:
Awesome. Will do, thank you!

PR #669, opened 2026-04-18
perf(file-tree): parallelize directory traversal and widen default ignore list

llkongs:
## Problem

On a SMB-mounted project with ~120 k files, `GET /api/projects/:name/files` takes **~2 minutes** to return; VS Code opens the same mount in 1 s thanks to lazy loading. Users hit similar slowness on any NFS / CIFS / sshfs mount or on monorepos where per-`stat()` round-trip dominates.

## Root cause

`getFileTree` (in `server/index.js`) does two things that serialize every round-trip:

1. A `for` loop that `await`s `fsPromises.stat(itemPath)` on each entry — so on a 1 000-file directory with 10 ms SMB round-trip, that alone is 10 s, and every subtree adds linearly.
2. A separate `fsPromises.access(item.path, R_OK)` probe **before** each recursive call — doubling the round-trip count for every directory.

The hardcoded ignore list also misses several large build / cache directories common outside the JS world (`__pycache__`, `.pytest_cache`, `.tox`, `.venv`, `target`, `.gradle`, `coverage`, `.next`, `.nuxt`, …).

## Change

- Extract the hardcoded skip list into an `IGNORED_DIRS` `Set` and extend it.
- Issue `stat()` and recursive `getFileTree()` calls for all children of a directory via `Promise.all` — lets Node pipeline the round-trips and overlap subtree traversals.
- Drop the redundant `access()` probe; EACCES now surfaces from the recursive call's own `readdir` `try/catch`, saving one RTT per directory.

No API-shape change: clients receive the same tree structure, same fields, same sort order. Tested with a local clone and pointed at both a native ext4 project and a CIFS-mounted 120 k-file project.

## Results (local)

| Project | Before | After |
|---|---|---|
| Native ext4, 43 files (PVECluster docs) | ~50 ms | ~30 ms |
| CIFS mount, ~120 k files (mixed JS / Python repos) | ~120 s | ~15 s (expected — not yet benchmarked in production) |

## Follow-ups (not in this PR)

- Lazy-load by folder expand (would require a frontend change to pass `?path=` and call per-folder instead of once per project).
- Honour a project-level `.gitignore` via the `ignore` npm package.

## Summary by CodeRabbit

* **Refactor**
  * Faster file-system traversal via optimized concurrent processing for snappier operations.
  * Broader exclusion of common system directories to reduce unnecessary scanning.
  * More robust handling of inaccessible or unreadable directories to improve stability and consistency.


viper151:
@llkongs  can you check the coderabbit comments??

PR #675, opened 2026-04-20
feat: update Claude models list, default, and context window detection

nunocodex:
## Summary
- Add all officially supported Claude models with full API IDs
  (Opus 4.7, Opus 4.6, Sonnet 4.6, Opus 4.5, Sonnet 4.5, Haiku 4.5)
- Remove redundant "Claude" prefix from labels (provider name already shown)
- Change default model to Sonnet 4.6 (matches Claude Code CLI default)
- Add `getClaudeContextWindow()` to resolve context window per model
  (200k standard, 1M for [1m] variants)
- TokenUsagePie now shows accurate context window based on selected model
  instead of hardcoded 160k

## Screenshots
image
image

## Test plan
- [ ] Verify model selector shows all models with clean labels
- [ ] Switch between models and verify the token pie updates to 200k or 1M
- [ ] Verify `CONTEXT_WINDOW` env var still works as fallback

## Summary by CodeRabbit

* **New Features**
  * More Claude model choices added (including Opus 4.7) and explicit support for 1M-context models.

* **Bug Fixes**
  * Token budget and context calculations now use the selected Claude model when available, improving accuracy for usage tracking.

* **Changes**
  * Default Claude model changed to Sonnet (user-facing default selection updated).

### DIFF
--- a/server/claude-sdk.js
+++ b/server/claude-sdk.js
@@ -282,9 +282,10 @@ function transformMessage(sdkMessage) {
 /**
  * Extracts token usage from SDK result messages
  * @param {Object} resultMessage - SDK result message
+ * @param {string} [selectedModel] - The user-selected model identifier (preserves [1m] suffix)
  * @returns {Object|null} Token budget object or null
  */
-function extractTokenBudget(resultMessage) {
+function extractTokenBudget(resultMessage, selectedModel) {
   if (resultMessage.type !== 'result' || !resultMessage.modelUsage) {
     return null;
   }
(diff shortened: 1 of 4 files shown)


nunocodex:
Thanks for the review @coderabbitai! Both findings have been addressed in commit a8b341d:

**1. `[1m]` suffix preservation in `extractTokenBudget`**
- Added optional `selectedModel` parameter to `extractTokenBudget(resultMessage, selectedModel)`.
- Context window is now resolved from `selectedModel || modelKey`, so user-selected aliases like `opus[1m]` / `sonnet[1m]` correctly map to the 1M window even after the SDK resolves `modelKey` to the versioned API id (e.g. `claude-opus-4-6`).
- Call site in the SDK runner now passes `sdkOptions.model`.

**2. Provider-aware fallback in `ChatComposer`**
- The `TokenUsagePie` fallback now gates `getClaudeContextWindow` on `provider === 'claude'`.
- Non-Claude providers (codex, cursor, gemini) fall back to `VITE_CONTEXT_WINDOW` or the original `160000` default, preserving prior behaviour.
- `tokenBudget?.total` still takes precedence when present, so real SDK values always win over fallbacks.

As a related cleanup, the Claude model list was restored to the original SDK-alias form with only Opus 4.7 added (`claude-opus-4-7`), which also removes the hard-coded `4.6` labels from the `[1m]` entries flagged in the inline review.

viper151:
@nunocodex Opus 4.7 needed a new claude agent sdk  version which is now in the latest version and ready to be used. I've also added opus 4.6 as well on the list.I'll close this PR since it's not relevant probably anymore.

PR #687, opened 2026-04-22
feat: add configurable BASE_PATH for subpath deployments

rjmayott:
## Summary

Adds support for serving CloudCLI from a configurable subpath (e.g. `/s/myproject/`) via the `BASE_PATH` environment variable, using **runtime injection** so a single build works at any mount point.

When `BASE_PATH` is unset, behavior is identical to current — **zero impact on root deployments**.

> **Relationship to #815:** That PR fixes runtime *router basename detection* when the HTML already contains prefixed asset URLs. This PR provides the rest of the stack — server middleware, fetch/WS URL prefixing, dynamic manifest, service worker, HTML injection — so the prefix actually makes it through every part of the request chain (especially `fetch('/api/…')` calls that React Router's basename doesn't cover). The two are complementary.

## What's in this PR

**Frontend**
- `src/utils/basePath.ts` — exports `BASE_PATH` (from `window.__CLOUDCLI_BASE_PATH__`) and an `assetUrl()` helper
- `authenticatedFetch` auto-prepends `BASE_PATH` for root-relative URLs (covers ~50 callers across 20+ files without touching each one)
- Bare `fetch()` auth endpoints, `EventSource` URLs, health check, and WebSocket URLs include `BASE_PATH`
- All component image `src` attributes use `assetUrl()`
- The existing `window.__ROUTER_BASENAME__` reading in App.tsx now has an authoritative server-injected value (still falls back to #815's URL sniffing if unset)

**Server**
- `server/index.js`: normalized `BASE_PATH` constant (leading slash, no trailing slash), early middleware to strip the prefix from `req.url` (supports direct subpath access without a smart reverse proxy), dynamic `/manifest.json` with prefixed `start_url`/`scope`/icons, `index: false` on `express.static` so SPA catch-all handles `index.html`, and XSS-safe HTML injection of `window.__CLOUDCLI_BASE_PATH__` + `window.__ROUTER_BASENAME__`
- `server/modules/websocket/services/websocket-server.service.ts`: strips `BASE_PATH` from `pathname` on a segment boundary before routing to `/ws`, `/shell`, `/plugin-ws/*`

**Service worker (`public/sw.js`)**
- Derives `BASE_PATH` from its own script location at runtime
- Whitelist-style fetch handler (parses request URL, checks `origin` + `pathname.startsWith(...)` rather than substring matching)
- Cached URLs, navigation responses, and push-notification icon paths all use the derived prefix

**Build (`vite.config.js`)**
- `base` reads from `VITE_BASE_PATH` (default `/`), so root deployments build identically to before; subpath deployments build with `VITE_BASE_PATH=/s/myproject` (or `./` for relative)

## Usage

```bash
# Root deployment (unchanged)
npm run build && npm run server

# Subpath deployment behind a reverse proxy
VITE_BASE_PATH=/s/myproject npm run build
BASE_PATH=/s/myproject npm run server
```

Reverse-proxy example (Caddy):

```caddyfile
:443 {
    handle /s/myproject/* {
        uri strip_prefix /s/myproject
        reverse_proxy 127.0.0.1:3001
    }
}
```

The server middleware also strips the prefix if your proxy doesn't, so a simpler proxy that just forwards `/s/myproject/*` without stripping also works.

## Test plan

- [x] Without `BASE_PATH`: app works at `http://localhost:3001/` — regression-clean
- [x] With `BASE_PATH=/s/test` behind Caddy: full UI loads at `/s/test/`
- [x] Registration, login, onboarding flow works through proxy
- [x] WebSocket connections (chat + shell + plugin-ws) connect at `/s/test/ws`, `/s/test/shell`, `/s/test/plugin-ws/<name>`
- [x] Logo icons (Claude, Cursor, Codex, Gemini) render
- [x] `manifest.json` returns prefixed paths
- [x] Service worker registers at the correct path
- [x] API calls route through `/s/test/api/...`
- [x] `npm run typecheck` + `npm run build` pass
- [x] CodeRabbit's prior review comments addressed: XSS-safe runtime-config injection, segment-boundary BASE_PATH stripping, parsed-URL pathname checks in sw.js, query-string handling on bare `BASE_PATH`, leading-slash normalization, `typeof rawUrl === 'string'` guard in `authenticatedFetch`

## Notes for reviewers

- Rebased onto current `main` (post-#815). All CodeRabbit feedback from the original review round folded into the squashed commit.
- The WebSocket BASE_PATH stripping ported from the old inline handler into the new `createWebSocketServer` module introduced by #715.
- No new dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ BASE_PATH=/s/myproject npm run server
 ```
 
 **Reverse proxy example (Caddy):**
-```
+```caddyfile
 :443 {
     handle /s/myproject/* {
         uri strip_prefix /s/myproject
--- a/README.md
+++ b/README.md
@@ -105,6 +105,31 @@ npx @cloudcli-ai/cloudcli@latest sandbox ~/my-project
 
 Supports Claude Code, Codex, and Gemini CLI. See the [sandbox docs](docker/) for setup and advanced options.
 
+#### Subpath Deployment (BASE_PATH)
+
+Serve CloudCLI from a subpath (e.g. `/s/myproject/`) instead of the root `/`. This is useful when running multiple CloudCLI instances behind a single reverse proxy.
+
+**Build with relative asset paths:**
+```bash
+VITE_BASE_PATH=./ npm run build
+```
+
+**Start the server with a base path:**
+```bash
+BASE_PATH=/s/myproject npm run server
+```
+
+**Reverse proxy example (Caddy):**
+```caddyfile
+:443 {
+    handle /s/myproject/* {
+        uri strip_prefix /s/myproject
+        reverse_proxy 127.0.0.1:3001
+    }
+}
+```
+
+When `BASE_PATH` is not set, CloudCLI serves from `/` as usual — no changes needed for standard deployments.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -105,6 +105,31 @@ npx @cloudcli-ai/cloudcli@latest sandbox ~/my-project
 
 Supports Claude Code, Codex, and Gemini CLI. See the [sandbox docs](docker/) for setup and advanced options.
 
+#### Subpath Deployment (BASE_PATH)
+
+Serve CloudCLI from a subpath (e.g. `/s/myproject/`) instead of the root `/`. This is useful when running multiple CloudCLI instances behind a single reverse proxy.
+
+**Build with relative asset paths:**
+```bash
+VITE_BASE_PATH=./ npm run build
+```
+
+**Start the server with a base path:**
+```bash
+BASE_PATH=/s/myproject npm run server
+```
+
+**Reverse proxy example (Caddy):**
+```
+:443 {
+    handle /s/myproject/* {
+        uri strip_prefix /s/myproject
+        reverse_proxy 127.0.0.1:3001
+    }
+}
+```
+
+When `BASE_PATH` is not set, CloudCLI serves from `/` as usual — no changes needed for standard deployments.
 
 ---
 
--- a/index.html
+++ b/index.html
@@ -34,7 +34,7 @@
     <script>
       if ('serviceWorker' in navigator) {
         window.addEventListener('load', () => {
-          navigator.serviceWorker.register('/sw.js')
+          navigator.serviceWorker.register((window.__CLOUDCLI_BASE_PATH__ || '') + '/sw.js')
             .then(registration => {
               console.log('SW registered: ', registration);
             })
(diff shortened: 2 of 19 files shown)


blackmammoth (commented):
Hey @rjmayott, thanks for the PR. Apologies for the delay in review. We had merged PR #815 which allows you to use cloudcli in subpath deployments. Is this PR still relevant? If so, can you update the merge conflicts so that it can be reviewed?

rjmayott:
Hey @blackmammoth, that's no problem at all, totally understand! I cleaned my branch up and rebased onto the current `main` (which now includes #815).

What this PR adds on top:

- **Server middleware** that strips the BASE_PATH prefix from incoming `req.url`, so subpath deployments work without requiring the reverse proxy to be configured to strip the prefix bidirectionally.
- **`authenticatedFetch` auto-prefix**: the ~50 `fetch('/api/…')` callers in the frontend would otherwise bypass any subpath the user mounted at, and React Router's `basename` doesn't help with `fetch()` URLs. The wrapper auto-prepends `BASE_PATH` for root-relative URLs.
- **WebSocket URL construction** in `src/contexts/WebSocketContext.tsx`, `src/components/shell/utils/socket.ts`, and the new `server/modules/websocket/services/websocket-server.service.ts` (which strips `BASE_PATH` on segment boundaries when routing `/ws`, `/shell`, `/plugin-ws/*`).
- **Dynamic `manifest.json`** that rewrites `start_url`, `scope`, and icon `src` paths with the prefix. (Browser fetches manifest directly, so the reverse proxy doesn't have a chance to rewrite its body.)
- **Service worker `BASE_PATH` derivation** from its own script path, so cached asset URLs, push-notification icons, and navigation URLs respect the subpath.
- **Vite `base` from `VITE_BASE_PATH`** at build time so the emitted HTML has prefixed asset/manifest references, which is what #815's detector reads.
- **Server-side HTML injection** of `window.__CLOUDCLI_BASE_PATH__` and `window.__ROUTER_BASENAME__`, so #815's detection has an authoritative source rather than having to sniff URLs.


Happy to revise the approach or anything if you'd prefer — let me know!

blackmammoth:
Thank you @rjmayott for the detailed explanation. I understand what you're doing but the way that we have setup the project, it will need a lot of testing to make sure that no endpoint is left behind and everything is supposed to work as is. For now, I have added a nginx.conf file that would help with subpath deployments that will proxy the /ai/api/... call to /api (See PR #820). 

We'll include your solution in the future after having done some project refactors. In the meantime, I'm closing this for now.

PR #696, opened 2026-04-23
fix(chat): unstick composer after Claude finishes its turn

antopiahk:
## Summary

- **Bug**: after Claude finished a turn, the composer was stuck showing "Thinking…" and disabled — users had to click STOP before they could send their next message.
- **Cause**: `useChatSessionState` had a `useEffect` that treated membership in `processingSessions` as a reason to force `isLoading=true`. But `processingSessions` means "Claude finished, awaiting user reply" (see `useSessionStatusMap.ts:10-22`, which owns the sidebar's pink "need you" badge). The `complete` handler adds the session to that set for the sidebar, and the effect immediately echoed it back into the loading state — undoing the `setIsLoading(false)` that `complete` had just performed.
- **Fix**: delete the effect. The navigate-to-a-running-session case is already handled by the server-sent `session-status` message (`useChatRealtimeHandlers.ts:138-176`), so this effect was redundant as well as wrong. Also removed the now-unused `processingSessions` prop on `useChatSessionState`.

Net diff: `-13 / +0`.

## Test plan

- [ ] Start a new Claude session; when the turn completes, the composer becomes editable and submittable (no STOP button, no "Thinking…").
- [ ] While Claude is working, the STOP button and "Thinking…" indicator still appear as before.
- [ ] Navigate to a different session that is currently running — the thinking indicator appears (via server `session-status`).
- [ ] Navigate to a finished session with the sidebar "need you" badge — composer is editable and the badge still shows until a message is sent.
- [ ] Sending a new message correctly clears the "need you" badge (already handled by `onSessionNotProcessing` in the submit path).

## Summary by CodeRabbit

* **New Features**
  * Added file upload support for Claude and Gemini queries
  * Project coloring and archiving functionality
  * Session status indicators and keyboard shortcuts for navigation
  * Command palette for session and project search
  * Session recovery on page reload
  * Attention indicator for sessions awaiting user action

* **Improvements**
  * Updated default Claude model to Opus 4.7
  * Enhanced session management interface
  * Improved theming and visual styling
  * Support for subdirectory deployments

* **Documentation**
  * Added deployment and development workflow guide


antopiahk:
Retargeting to startino/claudecodeui fork.

PR #702, opened 2026-04-24
fix: avoid Claude chat exit when bypass mode runs as root

michaeledi:
## Summary
- map Claude chat `bypassPermissions` to `dontAsk` when CloudCLI runs as `root`
- apply the same safeguard to the Claude-specific skip-permissions toggle
- add regression coverage for the root permission-mode mapping
- document the runtime behavior and test command

## Bug scenario
Claude chat fails immediately with `Claude Code process exited with code 1` when:
- CloudCLI runs as `root` (for example via `systemd` on a server)
- the user enables the chat mode button's `bypassPermissions` mode, or enables Claude skip-permissions
- the Claude provider is used through the chat UI (SDK path)

Shell mode does not show the same failure because it launches the interactive Claude CLI instead of the SDK stream-json path.

## Reproduction
1. Run CloudCLI as `root`
2. Open a Claude chat session
3. Switch the chat mode button to `bypassPermissions`, or enable Claude skip-permissions in settings
4. Send any message
5. Observe that the request fails immediately with `Claude Code process exited with code 1`

## Root cause
CloudCLI forwards Claude chat requests through `@anthropic-ai/claude-agent-sdk`, which launches Claude Code with stream-json arguments and `--permission-mode bypassPermissions`.

Claude rejects bypass mode when executed as `root`/`sudo`, returning:

`--dangerously-skip-permissions cannot be used with root/sudo privileges for security reasons`

That stderr is swallowed by the SDK and surfaces in CloudCLI only as a generic exit code 1.

## Suggested fix location
The safest place to fix this is the Claude SDK option mapping in:
- `server/claude-sdk.js`

Specifically, the permission-mode resolution should normalize both:
- explicit chat `permissionMode === 'bypassPermissions'`
- Claude skip-permissions settings

When CloudCLI is running as `root`, both should map to `dontAsk` instead of `bypassPermissions`.

## Validation
- `npm run test:claude-sdk-permissions`

## Release Notes

* **Bug Fixes**
  * Fixed Claude chat failures when CloudCLI runs as root. Permission bypass requests now map to "do not ask" mode to prevent SDK errors.

* **Documentation**
  * Added clarification on permission mode behavior when running with root privileges.

* **Tests**
  * Added test coverage for permission-mode resolution in Claude chat.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ All notable changes to CloudCLI UI will be documented in this file.
 
 ### Bug Fixes
 
+* Claude chat now maps both skip-permissions and explicit `bypassPermissions` chat mode to `dontAsk` when CloudCLI runs as root, avoiding Claude SDK exit code 1 from forbidden bypass mode
+
 * iOS scrolling main chat area ([3969135](https://github.com/siteboon/claudecodeui/commit/3969135bd427fbf48f29bb3dbfedb47791ca78dc))
 * migrate PlanDisplay raw params from native details to Collapsible primitive ([fc3504e](https://github.com/siteboon/claudecodeui/commit/fc3504eaed8ca7ed9214838d148ea385b8352c31))
 * precise Claude SDK denial message detection in deriveToolStatus ([09dcea0](https://github.com/siteboon/claudecodeui/commit/09dcea05fbc8c208d931aa1f08618f0e8087392f))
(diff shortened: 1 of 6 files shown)


blackmammoth:
Hey @michaeledi, I tried to replicate the reproduction steps and everything is working as expected. So, I'm closing this PR for now. If you think it's still necessary, you can resubmit another PR.

PR #719, opened 2026-04-28
fix(chat): prevent double send on mobile by removing redundant submit handlers

peterus:
## Summary

`PromptInputSubmit` is rendered as `type=\"submit\"` inside a `<form>` whose `onSubmit` already invokes `handleSubmit`. The component additionally wired `onMouseDown` and `onTouchStart` to call `onSubmit` directly, creating three submit paths.

On iOS Safari a single tap fires `touchstart` and a synthetic `mousedown` before React state propagates `isLoading=true`, so the `disabled={!input.trim() || isLoading}` guard does not catch the second invocation. Result: two user messages, two image-upload roundtrips per tap (visible as duplicate `.tmp/images/<timestamp>` folders).

This PR removes the redundant `onMouseDown` / `onTouchStart` handlers; the form's `onSubmit` is the single, browser-debounced submit path.

## Test plan

- [ ] Desktop click on the send button sends exactly once
- [ ] Desktop Enter in the textarea sends exactly once
- [ ] iOS Safari tap (with attached image) sends exactly once — single bubble, single `.tmp/images/<ts>` folder
- [ ] Send-button `disabled` state still gates submission while `isLoading`

## Summary by CodeRabbit

* **Refactor**
  * Simplified event handling in the chat composer's submit control by removing redundant pointer/touch event interception, relying instead on standard form submission mechanics.

### DIFF
--- a/src/components/chat/view/subcomponents/ChatComposer.tsx
+++ b/src/components/chat/view/subcomponents/ChatComposer.tsx
@@ -396,14 +396,6 @@ export default function ChatComposer({
             <PromptInputSubmit
               disabled={!input.trim() || isLoading}
               className="h-10 w-10 sm:h-10 sm:w-10"
-              onMouseDown={(event) => {
-                event.preventDefault();
-                onSubmit(event as unknown as MouseEvent<HTMLButtonElement>);
-              }}
-              onTouchStart={(event) => {
-                event.preventDefault();
-                onSubmit(event as unknown as TouchEvent<HTMLButtonElement>);
-              }}
             />
           </div>
         </PromptInputFooter>


happytechca:
Thanks for this PR. I am having the same issue in Chrome on Android.

wuc623123:
Confirming this is still reproducible on the latest packaged release.

**Environment**
- Package: `@cloudcli-ai/cloudcli@1.32.0` (downstream rebrand of this repo)
- Browser: Firefox 151.0.1 on Linux (also reported on iOS Safari and Android Chrome above)
- File checked: `dist/assets/index-AaSE7-3t.js`

**Reproducer pattern that's still in the bundle**

```js
onTouchStart:ue=>{ue.preventDefault(),k(ue)}
```

`k(ue)` here is the submit handler. With the form already submitting on `submit`, the redundant `onTouchStart` invokes it a second time on touch devices that synthesize both events.

**Patch verified working**

Stripping that `onTouchStart` (or replacing with `onTouchStart:void 0`) eliminates the double-send on every touch device I could test. A simple in-place sed on the production bundle reproduces what this PR does at source level:

```bash
sed -i 's|onTouchStart:ue=>{ue.preventDefault(),k(ue)}|onTouchStart:void 0|' dist/assets/index-*.js
```

This has been running in production for me for several weeks with zero regressions on desktop click behaviour. Would be great to get this merged — the bug is mild on desktop but actively painful on mobile (every prompt sends twice, every image uploads twice). Happy to retest if you want a fresh confirmation after rebase.

blackmammoth (approved):

blackmammoth:
Hey @peterus, @happytechca, and @wuc623123 this has been merged. Also, if you’d like to join project discussion or community chat, we have a Discord here: [link](https://discord.gg/buxwujPNRE).

PR #721, opened 2026-04-28
fix(websocket): reset unmountedRef so token refresh reconnects

peterus:
## Summary

`useWebSocketProviderState` keeps an `unmountedRef` that the effect cleanup sets to `true`, and `connect()` short-circuits when the ref is `true`. The ref is initialized to `false` once via `useRef(false)` but never reset.

The effect re-runs whenever `token` changes (auth refresh, login). On re-run, the previous cleanup has already set `unmountedRef.current = true`; the new effect calls `connect()`, which returns immediately at the unmounted guard. The socket then stays permanently disconnected for the lifetime of the provider — the `onclose`-driven reconnect timer also bails out, so no recovery happens.

Fix: reset `unmountedRef.current = false` at the top of the effect so re-runs can reconnect; the cleanup-set-to-true behavior on actual unmount is preserved.

## Test plan

- [ ] Initial mount: WebSocket connects, real-time messages arrive
- [ ] Trigger a token refresh (or re-login) while connected — connection re-establishes against the new token
- [ ] Real-time messages continue arriving after the refresh
- [ ] Provider unmount still closes the socket cleanly and prevents reconnect attempts (no zombie reconnect loop)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a WebSocket connection issue where reconnection attempts would fail after component unmounting or when authentication tokens are refreshed. This ensures real-time connections reliably reestablish in these scenarios, preventing service interruptions and maintaining consistent data synchronization.


blackmammoth (approved):
Thanks @peterus!

PR #730, opened 2026-04-30
Feature/font customization

rm1dev:
## Summary by CodeRabbit

* **New Features**
  * Customizable interface font (Default/Custom) and font size
  * Customizable code editor font for code blocks and inline code
  * Automatic text-direction detection for messages and composer
  * Font settings persist across sessions, are applied at startup, and update live when changed
  * Added localization for new appearance/font controls in multiple languages


blackmammoth (commented):
Hey @rm1dev, thanks for the PR. However, here are some concerns:

1. It's not clear how to change fonts. I think it would be better if the custom fonts are from a dropdown list instead.
2. The font size changer doesn't work. I changed it from 12 to 20px and it's still the same.
After this changes have been addressed, I would love to review it again. So, I'm making this a draft for now.

blackmammoth:

Issue #101, opened 2025-07-19
When using Node.js v24.4.0 on Windows, not WSL, an error occurs during the installation of npm install


ad56740051:
Node.js v24.4.0
When using Node.js v24.4.0 on Windows, not WSL, an error occurs during the installation of npm install
-----------------------------------------------------------------------------------------
I:\claudecode>npm install
npm warn deprecated xterm-addon-fit@0.8.0: This package is now deprecated. Move to @xterm/addon-fit instead.
npm warn deprecated xterm@5.3.0: This package is now deprecated. Move to @xterm/xterm instead.
npm warn cleanup Failed to remove some directories [
npm warn cleanup   [
npm warn cleanup     '\\\\?\\I:\\claudecode\\node_modules\\date-fns',
npm warn cleanup     [Error: EPERM: operation not permitted, rmdir 'I:\claudecode\node_modules\date-fns'] {
npm warn cleanup       errno: -4048,
npm warn cleanup       code: 'EPERM',
npm warn cleanup       syscall: 'rmdir',
npm warn cleanup       path: 'I:\\claudecode\\node_modules\\date-fns'
npm warn cleanup     }
npm warn cleanup   ],
npm warn cleanup   [
npm warn cleanup     'I:\\claudecode\\node_modules\\node-pty',
npm warn cleanup     [Error: EPERM: operation not permitted, rmdir 'I:\claudecode\node_modules\node-pty\build\deps\winpty\src\Release\obj\winpty-agent\deps\winpty\src'] {
npm warn cleanup       errno: -4048,
npm warn cleanup       code: 'EPERM',
npm warn cleanup       syscall: 'rmdir',
npm warn cleanup       path: 'I:\\claudecode\\node_modules\\node-pty\\build\\deps\\winpty\\src\\Release\\obj\\winpty-agent\\deps\\winpty\\src'
npm warn cleanup     }
npm warn cleanup   ]
npm warn cleanup ]
npm error code 1
npm error path I:\claudecode\node_modules\node-pty
npm error command failed
npm error command C:\Windows\system32\cmd.exe /d /s /c node-gyp rebuild
npm error conpty.cc
npm error I:\claudecode\node_modules\nan\nan.h(1173,27): warning C4996: 'v8::String::WriteUtf8': Use WriteUtf8V2 instead. [I:\claudecode\node_modules\node-pty\build\conpty.vcxproj]
npm error   (编译源文件“../src/win/conpty.cc”)
npm error
[...]

PR #736, opened 2026-05-04
fix: rebuild native addons after npm install --ignore-scripts

Nieaunder7:
## Problem

The web-terminal plugin fails to start on Linux with a misleading error:

```
Error: [web-terminal] Cannot find module 'node-pty' - run npm install in ...
```

The actual cause is that `node-pty` requires compilation of a native binary (`pty.node`), but `npm install --ignore-scripts` skips the `install` lifecycle script (`node scripts/prebuild.js || node-gyp rebuild`). On Linux, `node-pty@1.1.0` ships no prebuild, so the binary is never compiled and `require('node-pty')` fails at runtime.

Related issue: #101

## Root cause

`plugin-loader.js` runs `npm install --ignore-scripts` (correctly, for security). This prevents arbitrary `postinstall` hooks from untrusted plugins — but it also prevents native addon compilation.

## Fix

Add `rebuildNativeAddons()` that runs after `npm install --ignore-scripts` in both `installPluginFromGit` and `updatePluginFromGit`. It scans `node_modules` for packages with `binding.gyp` and runs `npm rebuild <package>` for each one. This is scoped to native addon packages only, limiting the script execution surface.

Rebuild failures are treated as non-fatal warnings so installation succeeds on systems without build tools (Python, node-gyp); the plugin reports a clear startup error in that case rather than silently failing.

## Platform impact

| Platform | Effect |
|----------|--------|
| Linux x64 | `pty.node` compiled via node-gyp — fixes the bug |
| macOS / Windows | prebuild already present in package, rebuild succeeds quickly and is a no-op |
| Docker (no build tools) | rebuild fails, logged as warning, installation continues |

## Bug Fixes

* Native addons are now automatically rebuilt during plugin installation and updates to ensure proper functionality.
* Added improved error handling with warning notifications if native addon rebuilding encounters issues.

### DIFF
--- a/server/utils/plugin-loader.js
+++ b/server/utils/plugin-loader.js
@@ -94,6 +94,65 @@ export function validateManifest(manifest) {
 }
 
 const BUILD_TIMEOUT_MS = 60_000;
+const REBUILD_TIMEOUT_MS = 120_000;
+
+/**
+ * Run `npm rebuild` for packages that have a binding.gyp (native addons).
+ * --ignore-scripts skips the install/build lifecycle scripts that native modules
+ * need to compile their .node binaries, so we rebuild them explicitly afterwards.
+ * This is scoped to packages with binding.gyp to limit script execution surface.
+ */
+function rebuildNativeAddons(dir, callback) {
+  const nodeModulesDir = path.join(dir, 'node_modules');
+  let nativePackages = [];
+  try {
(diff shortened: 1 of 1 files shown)


Nieaunder7:
Closing: the fix partially undermines the intentional --ignore-scripts security design. Will reconsider the approach before resubmitting.

PR #745, opened 2026-05-05
feat: pin manually renamed session names

gcdevuk:
## Summary

- Adds `is_name_pinned` flag so manual session renames via the sidebar edit button are preserved permanently — previously they were overwritten on the next sync cycle
- Falls back to extracting the first user message from the session JSONL as the session name when `history.jsonl` is unavailable (common on headless/VPS deployments)

## Changes

- **migrations.ts** — adds `is_name_pinned` column
- **schema.ts** — adds column to schema definition
- **sessions.db.ts** — upsert skips `custom_name` update when pinned; setting a custom name now also sets `is_name_pinned = 1`
- **claude-session-synchronizer.provider.ts** — new `extractFirstUserMessage()` fallback for session naming; refactored `parseSessionFile` to try `history.jsonl` name map first, then first user message

## Test plan

- [ ] Rename a session in the sidebar, trigger a sync, confirm name persists
- [ ] Create a new session on a headless deployment without `history.jsonl`, confirm it picks up the first user message as the name
- [ ] Confirm existing unnamed sessions still show "Untitled Claude Session"

## Release Notes

* **New Features**
  * Added ability to pin custom session names to prevent them from being overwritten during updates.
  * Enhanced session naming for Claude sessions with improved fallback logic for better automatic naming.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -123,24 +123,10 @@
         "url": "https://github.com/sponsors/sindresorhus"
       }
     },
-    "node_modules/@ampproject/remapping": {
-      "version": "2.3.0",
-      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.3.0.tgz",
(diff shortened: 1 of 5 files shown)


blackmammoth:
Hey there @gcdevuk, thanks for the PR. However, it's too large for review. Can you submit it in small chunks?

PR #752, opened 2026-05-08
Fix: index Cursor sessions from agent-transcripts/, support nested depths

kix-sisigai:
## Why

Recent cursor-agent versions write JSONL transcripts under `~/.cursor/projects/<project-dir>/agent-transcripts/<chatId>/...`. The legacy `~/.cursor/chats/<projectHash>/` directory still exists but now holds only SQLite `store.db` files, used by the loader (`cursor-sessions.provider.ts`), not JSONL the indexer can parse. Result: on `main`, `CursorSessionSynchronizer` finds zero transcripts for any current Cursor user — every initial scan returns 0, advances `scan_state.last_scanned_at`, and from then on the file watcher only catches whatever brand-new chats happen after startup.

The transcript layout itself isn't fixed either — the same install can have chats at both `agent-transcripts/<chatId>/<chatId>.jsonl` and `agent-transcripts/<chatId>/<sub>/<chatId>.jsonl`, so the hard-coded `path.dirname(path.dirname(filePath))` silently skipped the deeper variant.

## Changes
- Switch the cursor synchronizer to scan `~/.cursor/projects/<name>/agent-transcripts/` for transcripts; the legacy hashed `chats/` dir is left to the loader, which already has its own md5 hashing.
- Locate `worker.log` by walking up from the transcript file, bounded to `~/.cursor/projects`, so both nested depths work and a transcript outside the cursor home returns null instead of asserting on path arithmetic.
- Pass the already-resolved `projectPath` from the project loop into `processSessionFile` to avoid redundant per-file resolution; the watcher entry point still resolves it via the walk-up helper.
- Drop the now-dead `md5` helper and `crypto` import from the synchronizer.
- Add `server/modules/providers/tests/cursor-session-synchronizer.test.ts` covering: transcripts at both depths, a project missing `worker.log`, the legacy `chats/<hash>/store.db` presence (must not be claimed by the indexer), watcher-added transcripts, transcripts outside `~/.cursor/projects`, and the incremental `since` filter.

## Testing
1. \`TSX_TSCONFIG_PATH=server/tsconfig.json node --import tsx --test server/modules/providers/tests/cursor-session-synchronizer.test.ts\` → 1/1 pass.
2. Existing \`server/modules/providers/tests/mcp.test.ts\` → 4/4 still pass.
3. \`tsc --noEmit -p server/tsconfig.json\` → clean.
4. Smoke-tested against a real \`~/.cursor\` with 109 jsonls across two projects: pre-fix indexed 1 (the one chat that was modified after CloudCLI started, picked up by the watcher); post-fix initial scan indexes 62 — every transcript whose project has a parseable \`worker.log\`.

## Summary by CodeRabbit

* **Refactor**
  * Updated transcript discovery mechanism to use location-aware detection instead of hash-based lookup, improving reliability across varying project structures and directory layouts.

* **Tests**
  * Added comprehensive test coverage for transcript indexing behavior, including edge cases and multiple directory structure scenarios.

### DIFF
--- a/server/modules/providers/list/cursor/cursor-session-synchronizer.provider.ts
+++ b/server/modules/providers/list/cursor/cursor-session-synchronizer.provider.ts
@@ -1,4 +1,3 @@
-import crypto from 'node:crypto';
 import fs from 'node:fs';
 import fsp from 'node:fs/promises';
 import os from 'node:os';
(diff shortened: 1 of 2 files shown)


blackmammoth:
Hey @kix-sisigai, thank for the PR! Apologies for the delay in review. There has been a ton of changes since this PR. Can you check if this PR is still relevant? If so, can you resolve the merge conflicts and I will make sure to review again.

blackmammoth:

PR #782, opened 2026-05-24
feat(sidebar): tooltip for the active-session indicator dot

navarrotech:
## Summary

The pulsing green dot to the left of a sidebar session row signals that the session had activity in the last 10 minutes (`isActive: diffInMinutes < 10` in `src/components/sidebar/utils/utils.ts`), but its meaning was undocumented. Users hovering it learn nothing.

This PR adds a translated tooltip and an `aria-label` so the same information reaches keyboard / screen-reader users.

## Implementation notes

- Uses the existing shared `Tooltip` component from `src/shared/view/ui/Tooltip.tsx`. It is portal-positioned, so it is not clipped by the sidebar's scroll container.
- `position="right"` keeps the tooltip out of the row, since the dot sits flush against the left edge.
- The dot now has `role="status"` and `aria-label`, so screen readers announce it even though the visual cue is the colour + animation.
- Translation key `tooltips.activeSessionIndicator` added to all eight locale files: en, de, it, ja, ko, ru, tr, zh-CN.

## Test plan

- [ ] Open a session that was active in the last 10 minutes (pulsing green dot visible). Hover the dot. Tooltip appears explaining the indicator.
- [ ] Switch language via the language selector. Hover again. Tooltip text is translated.
- [ ] Sessions older than 10 minutes do not render the dot (and therefore no tooltip), as before.
- [ ] No regression: pencil / save / cancel / trash row buttons still work as expected on hover.

## Summary by CodeRabbit

* **New Features**
  * Active session indicator now displays a tooltip showing recent activity status (last 10 minutes).

* **Localization**
  * Tooltip translations added in English, German, Italian, Japanese, Korean, Russian, Turkish, and Simplified Chinese.

[Review Change Stack](https://app.coderabbit.ai/change-stack/siteboon/claudecodeui/pull/782?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


navarrotech:
Testing:
{FFB66687-9962-4EB9-9B67-FDB3A375868D}

blackmammoth (approved):

PR #788, opened 2026-05-24
fix(chat): re-anchor initial scroll across lazy content reflow

shockstricken:
## Summary

The initial scroll-to-bottom for a freshly-loaded chat session fires once via \`setTimeout(scrollToBottom, 200)\` and clears its pending flag. When markdown rendering, syntax highlighting, or embedded images finish laying out *after* that 200ms window, \`scrollHeight\` grows but nothing re-anchors the viewport. The user sees the tab visually "scrolled way up" with the latest assistant message off-screen until they manually scroll or send a new message.

This replaces the one-shot timeout with a \`requestAnimationFrame\` loop that re-scrolls every frame while \`scrollHeight\` is still growing. The loop terminates when either:

- \`scrollHeight\` has been stable for 3 consecutive frames (content settled), or
- 60 frames have elapsed (~1s safety cap).

The existing \`pendingInitialScrollRef\` flag is reused so session changes cancel the in-flight loop cleanly, and a cleanup function cancels any pending \`requestAnimationFrame\` on unmount.

## Why \`requestAnimationFrame\` over \`ResizeObserver\`

A \`ResizeObserver\` on the scroll container would also work and would naturally retry on every reflow, but it has two downsides for this case:
- Cleanup is more involved (observer lifecycle versus a single rAF id).
- It would fire on every reflow forever, requiring a separate "we already initial-scrolled" flag and probably a debounce — more state for the same result.

The bounded rAF loop achieves the same correctness with less state.

## Behavior diff

| Scenario | Before | After |
|---|---|---|
| Session with already-rendered text only | Lands at bottom (no change) | Lands at bottom (no change) |
| Session with images/long markdown that render after 200ms | Lands above the bottom | Lands at bottom |
| User switches session mid-scroll-loop | rAF keeps running on the old session | rAF cancels immediately via the pending-flag guard |

## Test plan

- [ ] \`npm run typecheck\` clean (verified locally)
- [ ] Open a chat session with no images / short content — viewport lands at bottom (regression check)
- [ ] Open a chat with many images and long markdown — viewport lands at bottom after content settles
- [ ] Switch between sessions rapidly — no console errors, no scroll fighting
- [ ] Scroll up in a fresh session while it's still loading content — verify the loop respects user intent (note: the current implementation doesn't, but it's bounded to 1s so the worst case is brief)

## Notes

- File touched: \`src/components/chat/hooks/useChatSessionState.ts\` only.
- No changes to public component props, no new dependencies, no behavior change for sessions whose layout is already stable on first frame.

Fork base: \`shockstruck/cloudcli\` (rebrand of this repo's main).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved initial chat scrolling so the viewport reliably stays at the latest messages. Scrolling now adapts while content (images, formatted text, code blocks) finishes rendering, preventing the view from getting stuck above new messages.


blackmammoth:
Hey @shockstricken, thanks for the PR. Can you send a detailed reproduction step for the issue this PR attempts to solve?

blackmammoth:
Hey @shockstricken, thanks for the PR. Since there's no update to the above comment, I'm gonna assume the issue no longer exists and I'm going to close it. If you still think the PR is necessary, you can resubmit it at anytime.

PR #802, opened 2026-05-28
feat: add Codex permission env fallback

CoderLuii:
## Summary
- Add `CLOUDCLI_CODEX_PERMISSION_MODE` as a server-side fallback for Codex chat requests that omit `permissionMode`.
- Preserve explicit request permission modes and validate values against `default`, `acceptEdits`, and `bypassPermissions`.
- Keep CloudCLI's default unchanged: unset or invalid env values still resolve to `default`.
- Document the fallback, concrete deployment examples, and the `bypassPermissions` safety tradeoff.

## Why
This came from the HolyClaude use case in CoderLuii/HolyClaude#18 and the v1.2.6 downstream patch:

- https://github.com/CoderLuii/HolyClaude/issues/18
- https://github.com/CoderLuii/HolyClaude/releases/tag/v1.2.6

The current CloudCLI browser chat submit path sends `permissionMode`, so normal UI requests keep their selected mode. The fallback is for direct WebSocket clients or custom frontends that omit `permissionMode`, and for container operators who want a deployment-level Codex default.

Examples:

- `default`: cautious or shared deployments where omitted-mode requests should stay on the standard sandbox and approval path.
- `acceptEdits`: trusted self-hosted workspaces where Codex should apply file edits without repeated prompts while staying in `workspace-write`.
- `bypassPermissions`: single-user trusted local containers where the operator intentionally wants full access inside the container and mounted workspace.

Explicit request values always win over the env var.

## Testing
- `npm ci` completed; npm reported 39 advisories in the dependency tree.
- `node --check server/openai-codex.js`
- `node --test server/codex-permission-mode.test.js`
- `git diff --check`
- `npx eslint server/openai-codex.js server/codex-permission-mode.js server/codex-permission-mode.test.js`
- `npm run lint` completed with 291 warnings and 0 errors.
- `npm run typecheck`
- `npm run build` completed with Browserslist, CSS minify, and chunk-size warnings.

Not live-tested: an actual Codex chat runtime session.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,24 @@ To use Claude Code's full functionality, you'll need to manually enable tools:
 
 **Recommended approach**: Start with basic tools enabled and add more as needed. You can always adjust these settings later.
 
+### Codex Permission Defaults
+
+Codex chat supports three permission modes:
+
+- `default` — uses Codex's `workspace-write` sandbox with the `untrusted` approval policy.
+- `acceptEdits` — uses the `workspace-write` sandbox with approval disabled, so Codex can apply workspace edits without prompting while staying sandboxed.
+- `bypassPermissions` — uses `danger-full-access` with no approval. Use it only in trusted local or sandboxed workspaces.
+
+The UI sends the selected mode with each Codex request. For self-hosted deployments that need a server-side fallback when a client does not send a mode, set `CLOUDCLI_CODEX_PERMISSION_MODE` to one of those values. If unset or invalid, CloudCLI uses `default`.
+
+Examples:
+
+```bash
+CLOUDCLI_CODEX_PERMISSION_MODE=default npx @cloudcli-ai/cloudcli
+CLOUDCLI_CODEX_PERMISSION_MODE=acceptEdits npx @cloudcli-ai/cloudcli
+CLOUDCLI_CODEX_PERMISSION_MODE=bypassPermissions npx @cloudcli-ai/cloudcli
+```
+
 ---
 
 ## Plugins
--- a/server/codex-permission-mode.js
+++ b/server/codex-permission-mode.js
@@ -1,0 +1,52 @@
+export const CODEX_PERMISSION_MODE_ENV = 'CLOUDCLI_CODEX_PERMISSION_MODE';
+
+const CODEX_PERMISSION_MODES = new Set(['default', 'acceptEdits', 'bypassPermissions']);
+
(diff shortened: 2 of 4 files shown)


blackmammoth:
Hey @CoderLuii, thanks for the PR. Can u resolve the merge conflict and I'll review it.

blackmammoth:
Hey @CoderLuii, on the pr description it says _"Self-hosted and containerized CloudCLI deployments **_sometimes_** need a process-level Codex permission default for clients that do not send permissionMode"_. When would you need this? When is the '**sometimes**'?

CoderLuii:
@blackmammoth thanks for calling that out. You were right that sometimes was too vague.

The concrete use case came from HolyClaude: CoderLuii/HolyClaude#18. HolyClaude ships CloudCLI inside a trusted self-hosted Docker image where users run multiple coding agents against the same mounted workspace. The current CloudCLI browser chat submit path sends `permissionMode`, but deployment owners still need a server-side fallback for direct WebSocket clients or custom frontends that omit it.

Concrete examples:

- A cautious or shared deployment can leave the env var unset, so omitted-mode Codex requests stay on `default`.
- A trusted self-hosted workspace can set `CLOUDCLI_CODEX_PERMISSION_MODE=acceptEdits`, so Codex can apply file edits without repeated prompts while still using the `workspace-write` sandbox.
- A single-user local container can intentionally set `CLOUDCLI_CODEX_PERMISSION_MODE=bypassPermissions` when they want Codex to have full access inside the container and mounted workspace. That is powerful, so it should only be used where the operator trusts the workspace and prompts.

That is why this PR keeps explicit request values first, and only uses `CLOUDCLI_CODEX_PERMISSION_MODE` when the request omits `permissionMode`. It also keeps upstream CloudCLI's default unchanged: if the env var is unset or invalid, Codex still uses `default`.

I also cleaned the generated noise out of the PR description. Sorry about that; it made the PR harder to review than it needed to be.

I refreshed the branch against current `main`, resolved the conflicts, and kept the newer Codex model-resolution path intact. The PR is now intended as the upstream version of that small server-side fallback, so downstream images do not need to patch CloudCLI just to set a deployment-level Codex default.

blackmammoth:
Hey @CoderLuii, thanks for the PR. I have closed this as the changes are too complicated and very specific to one provider. I'll try to find a way that works globally for all providers.

PR #812, opened 2026-05-30
fix(gemini): graceful fallback for stale Gemini session IDs on resume

bourgois:
## Summary

Fixes the upstream defect where Gemini CLI assigns a new internal session ID on each run, causing all subsequent resume attempts to fail with exit code 42 and \"Invalid session identifier\".

### Chat path (`server/gemini-cli.js`)
- Always refresh `cliSessionId` in the `onInit` handler instead of only writing it once, so the stored value stays current after every run
- On exit code 42 + \"Invalid session identifier\" in stderr: clear the stale `cliSessionId`, suppress the buffered error messages, and retry once with `--resume latest`
- Suppress the `complete` WebSocket event before the retry so the client doesn't receive a spurious exitCode 42 completion followed by a second stream
- Buffer \"invalid session identifier\" stderr lines instead of forwarding immediately — discarded silently on a successful retry, flushed to the client only if the retry also fails

### Terminal/PTY path (`server/modules/websocket/services/shell-websocket.service.ts`)
- The shell path built `gemini --resume "<id>"` with no fallback, leaving the terminal stuck on a stale ID
- Add a three-level fallback: `--resume <id>` → `--resume latest` → fresh session, with a PowerShell-compatible variant on Windows (mirroring how the Claude provider already handles this)

## Test plan

- [ ] Start a Gemini session, send a message, close and reopen — chat should resume without any visible error
- [ ] Open the same session in the terminal tab — should recover via `--resume latest` fallback instead of dying with exit 42
- [ ] Verify that after a successful resume the stored `cliSessionId` is updated (future restarts should also work)
- [ ] Confirm a genuine exit-code-42 (non-session reason) still surfaces an error to the UI and does not trigger a retry loop
- [ ] Verify on Windows that the PowerShell fallback chain behaves correctly

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session resilience: CLI now detects invalid/stale sessions and automatically retries resuming to avoid repeated failures.
  * Single automatic stale-resume retry introduced to recover from transient resume errors.
  * Added fallback resume behavior: if resuming a stored session fails, the system falls back to resuming the latest session, then starts fresh.
  * More reliable session tracking: stored session identifiers are refreshed on init to reduce future resume errors.
  * Reduced noisy error emissions during automatic retry attempts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/siteboon/claudecodeui/pull/812?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/gemini-cli.js
+++ b/server/gemini-cli.js
@@ -130,6 +130,7 @@ async function spawnGemini(command, options = {}, ws) {
     let sessionCreatedSent = false; // Track if we've already sent session-created event
     let assistantBlocks = []; // Accumulate the full response blocks including tools
     let stderrHasInvalidSession = false; // Detect stale cliSessionId errors
+    let invalidSessionStderrBuffer = []; // Hold invalid-session messages until we know if we'll retry
 
     // Use tools settings passed from frontend, or defaults
     const settings = toolsSettings || {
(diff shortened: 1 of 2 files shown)


blackmammoth:
Hey @bo, there has been a complete revamp of the websocket and id logic so this be fixed. If the issue still persists, let me know and we'll circle back on it.

PR #823, opened 2026-06-04
fix: render CJK glyphs in terminal (fontFamily + optional WebGL renderer)

### LINKED ISSUE #822: Terminal renders CJK (Japanese/Chinese/Korean) text as black boxes — WebGL renderer + Latin-only fontFamily
### Describe the bug
In the built-in terminal (Shell view), CJK characters (Japanese / Chinese / Korean)
are frequently rendered as solid black boxes (□ "tofu") instead of the real glyphs.

The same Japanese text renders correctly everywhere else in the app (chat panel,
sidebar, etc.), so this is specific to the xterm.js terminal — not the OS fonts.

The breakage is per-glyph / intermittent: most CJK in a line renders, but common
kana/kanji (e.g. ク, レ, 差, 正) randomly turn into boxes. That pattern is the
signature of a font-fallback / glyph-atlas problem in the WebGL renderer, not of
missing OS fonts (a real coverage gap would drop *all* CJK, not a random subset).

### Steps to reproduce
1. Open the terminal and run a command that prints CJK, e.g. `echo 日本語テスト 中文 한국어`
   (or run any CLI that emits Japanese/Chinese/Korean output).
2. Some CJK characters render as solid black boxes.

### Screenshots

Image

### Environment
- claudecodeui: `main`, self-hosted
- Browser: Brave (Chromium) on Linux / Wayland
- OS CJK fonts installed (`noto-fonts-cjk`); identical text renders fine in the
  non-terminal UI, confirming the OS has the glyphs.

### Root cause (from source)
Two compounding issues under `src/components/shell/`:

1. **`fontFamily` has no CJK font** — `src/components/shell/constants/constants.ts`:
   ```ts
   export const TERMINAL_OPTIONS: ITerminalOptions = {
     // ...
     fontFamily: 'Menlo, Monaco, "Courier New", monospace',
     // ...
   };
   ```
   Menlo / Monaco / Courier New have no CJK coverage, so CJK can only come from the
   generic `monospace` fallback.

2. **WebGL renderer is loaded unconditionally** — `src/components/shell/hooks/useShellTerminal.ts`:
   ```ts
   try {
     nextTerminal.loadAddon(new WebglAddon());
   } catch {
     console.warn('[Shell] WebGL renderer unavailable, using Canvas fallback');
   }
   The xterm.js WebGL (and canvas) renderer rasterizes glyphs into a texture atlas and
   handles *fallback* glyphs (those not in the specified fonts) poorly — CJK obtained
   via the generic `monospace` fallback frequently comes out as the missing-glyph box.
   The DOM renderer doesn't have this issue because it uses native browser text
   rendering / font fallback.

### Suggested fix
- **Minimal:** add cross-platform CJK monospace fonts to `TERMINAL_OPTIONS.fontFamily`,
  giving the atlas an explicit CJK font to rasterize from:
  ```ts
  fontFamily:
    'Menlo, Monaco, "Courier New", "Noto Sans Mono CJK JP", "Noto Sans CJK JP", ' +
    'Meiryo, "MS Gothic", "Microsoft YaHei", "PingFang SC", "Hiragino Sans", monospace',
  ```
- **Robust:** make `fontFamily` user-configurable (relates to #730) and/or expose an
  option to disable the WebGL addon (fall back to the DOM renderer), which renders CJK
  reliably even with an imperfect font stack.

mhit:
## Summary

CJK characters (Japanese / Chinese / Korean) are frequently rendered as solid black boxes ("tofu") in the built-in xterm.js terminal, while the same text renders fine elsewhere in the app. This fixes the two compounding root causes.

## Root cause

1. **`fontFamily` has no CJK font** (`src/components/shell/constants/constants.ts`). `Menlo, Monaco, "Courier New", monospace` has no CJK coverage, so CJK can only come from the generic `monospace` fallback.
2. **WebGL renderer is loaded unconditionally** (`src/components/shell/hooks/useShellTerminal.ts`). The xterm.js WebGL/Canvas renderer rasterizes glyphs into a texture atlas and handles *fallback* glyphs poorly, so CJK obtained via generic `monospace` frequently comes out as the missing-glyph box. The intermittent, per-glyph nature of the breakage is the signature of this atlas/fallback issue (a real OS coverage gap would drop *all* CJK).

## Changes

- Add cross-platform CJK monospace fallbacks to `TERMINAL_OPTIONS.fontFamily`, giving the glyph atlas an explicit CJK font to rasterize from.
- Add `isWebglDisabled()` and gate the `WebglAddon` behind it, so the WebGL renderer can be turned off (falling back to the DOM renderer, which renders CJK reliably). Resolution order:
  1. `localStorage['terminal-disable-webgl'] === 'true'` (runtime, no rebuild)
  2. `VITE_TERMINAL_DISABLE_WEBGL === 'true'` (build-time default)
  - Defaults to keeping WebGL enabled, so behavior is unchanged unless opted out.

## How to test

1. Open the terminal and run `echo 日本語テスト 中文 한국어`.
2. With the font change alone, CJK should render correctly in most environments.
3. If boxes still appear, run `localStorage.setItem('terminal-disable-webgl','true')` in DevTools and reopen the terminal — the DOM renderer renders CJK reliably even with an imperfect font stack.

## Notes / trade-offs

- The DOM renderer is slower than WebGL for heavy output; that's why WebGL stays on by default and the opt-out is user-controllable. This relates to the font-configurability request in #730.
- No server-side changes; client rebuild only.

Fixes #822

## Summary by CodeRabbit

* **Improvements**
  * Enhanced terminal font rendering with better support for CJK characters (Japanese, Chinese, Korean)
  * Added option to disable WebGL renderer for improved compatibility on certain systems
  * Improved terminal initialization with graceful fallback handling


blackmammoth:
Hey @mhit, thanks for the PR. I tried to reproduce the issue but couldn't replicate it. Here were my steps:
1. Run CloudCLI on main branch
2. Go to shell tab
3. Send `echo 日本語テスト 中文 한국어` to claude code
4. Observed that it's displaying it as expected with no black box. 


Is my reproduction system correct? I checked this both on Windows and Linux.
image

blackmammoth:

PR #618, opened 2026-04-04
fix: persist selected thinking mode in chat composer

CoderLuii:
## Summary
- initialize the chat composer thinking mode from local storage
- persist the selected mode and stop resetting it after each send

## Testing
- validated in HolyClaude integration flow while reproducing the thinking-mode selection reset in the main chat UI

## Summary by CodeRabbit

* **New Features**
  * Thinking mode preference persistence: your selected thinking mode is now saved and restored across sessions for a consistent experience.

* **Bug Fixes**
  * Thinking mode no longer resets when sending messages, so your chosen mode remains active until you change it.

[Review Change Stack](https://app.coderabbit.ai/change-stack/siteboon/claudecodeui/pull/618?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #826, opened 2026-06-04
fix(chat): persist thinking mode selection

blackmammoth:
Closes PR #618 

Initialize the composer thinking mode from localStorage so reloads keep the user's selected mode instead of falling back to Standard.

Keeps the chosen mode after sending because the selector behaves like a preference, not a one-shot modifier for a single prompt.

## Summary by CodeRabbit

* **New Features**
  * The chat composer now persists your selected thinking mode across messages and sessions, so your preference is retained automatically.
* **Bug Fixes**
  * Thinking-mode prefixes are applied only for the Claude provider as intended.
  * Sending a message no longer resets the chosen thinking mode, preserving it between sends.

### DIFF
--- a/src/components/chat/hooks/useChatComposerState.ts
+++ b/src/components/chat/hooks/useChatComposerState.ts
@@ -143,6 +143,21 @@ const createFakeSubmitEvent = () => {
   return { preventDefault: () => undefined } as unknown as FormEvent<HTMLFormElement>;
 };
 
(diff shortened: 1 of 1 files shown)


viper151:
@blackmammoth  as discussed, thinking mode is not relevant anymore. There is an effort we could replace it with but it changes per model

viper151 (changes requested):
@blackmammoth as discussed, thinking mode is not relevant anymore. There is an effort we could replace it with but it changes per model

PR #830, opened 2026-06-04
fix(sidebar): add keyboard support and larger hit targets to project actions

iRukhaimi:
## What

The project **star toggle** in the sidebar was a click-only `div` with no keyboard affordance. This PR makes it keyboard-operable:
- `role="button"`, `tabIndex`, and an `aria-label`
- Enter/Space key handler
- a `focus-visible` ring

It also bumps the **rename** and **delete** action buttons from 6×6 to 7×7, so the hit targets are easier to reach next to the truncated project title.

## Why

Self-contained accessibility/UX hardening. **No behavior change for mouse users; additive a11y/UX.**

## Scope

This is item (1) from the upstream scope discussion: https://github.com/siteboon/claudecodeui/discussions/829

Part of AlKindy's open-source work. Signed-off-by per DCO.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard accessibility for the star control, enabling toggling via Enter/Space and ensuring proper focus/aria labeling.
  * Isolated star clicks from row selection so starring no longer triggers row navigation; project selection is now activated via the project name control.

* **Style**
  * Increased action icon sizes and updated focus/opacity styling for clearer visibility and interaction.


blackmammoth:
Hey @iRukhaimi, thanks for the PR. However, I had the following concerns:

1. the focus ring is not UI wise good.
2. I don't see much difference after changing the hit targets size. It still looks almost identical in size as before.

Since the above concerns are major points of this PR that need to be addressed, I'll close it for now. When there's an update, you can resubmit and I'll make sure to review.

PR #849, opened 2026-06-08
fix: auto-detect claude binary on Linux/macOS + inject GitHub token into git remote operations

lepisodesangwoo:
## Summary

Two independent but related usability bugs fixed in a single PR.

---

## Fix 1: Claude binary auto-detection on Linux / macOS

**File:** `server/shared/claude-cli-path.ts`

### Problem

`resolveClaudeCodeExecutablePath()` on non-Windows platforms did nothing more than return the configured path verbatim. If `CLAUDE_CLI_PATH` was not set the function returned the bare string `'claude'`, which the SDK then tried to spawn directly.

This causes two failure modes depending on what is installed:

| Situation | Error |
|-----------|-------|
| `claude` not on `PATH` | `Claude Code native binary not found at claude` |
| npm global wrapper on `PATH` instead of native binary | `Claude Code native binary at … exists but failed to launch` (ENOENT on shebang interpreter) |

The second failure is particularly confusing: `@anthropic-ai/claude-code` installed via `npm install -g` creates a **shell-script wrapper** at e.g. `/usr/local/share/npm-global/bin/claude`. The SDK's `K2()` function sees no `.js` extension and treats it as a native binary, then spawns it with `child_process.spawn` directly. If the shebang interpreter is not in `PATH` at spawn time the process emits `ENOENT` → "exists but failed to launch".

The native Anthropic installer places the real binary at `~/.local/bin/claude` (Linux) or similar, but this is also frequently absent from the server process's `PATH`.

### Fix

Added `resolveUnixClaudeExecutablePath()` which runs a **priority-ordered probe** before falling back to the bare command:

1. Explicit `CLAUDE_CLI_PATH` / absolute path → used as-is (existing behaviour)
2. Well-known native-installer locations: `~/.local/bin/claude`, macOS `~/Library/…`, Homebrew, etc.
3. Common npm global prefix paths
4. Active nvm version, all installed nvm versions
5. `which` / `command -v` shell lookup
6. Bare `'claude'` as last resort (lets the OS decide at spawn time)

This removes the need to set `CLAUDE_CLI_PATH` in the vast majority of setups.

---

## Fix 2: Stored GitHub token not used for git push / pull / fetch / publish

**File:** `server/routes/git.js`

### Problem

The UI lets users store GitHub personal access tokens under **Settings → Credentials → GitHub Tokens**. These tokens were looked up and used only during project cloning (`project-clone.service.ts`) and **completely ignored** for every other remote git operation:

- `POST /api/git/fetch`
- `POST /api/git/pull`
- `POST /api/git/push`
- `POST /api/git/publish`

A user who saved a token and expected push/pull to "just work" would receive **Permission denied / authentication failed** from git, even though the token was sitting right there in the database.

### Fix

Added `withGithubToken(projectPath, userId, fn)` helper:

- Retrieves the user's most recently active GitHub token via `githubTokensDb.getActiveGithubToken()`.
- Temporarily rewrites the `origin` remote URL to embed the token (`https://TOKEN@github.com/…`) for the duration of the wrapped `fn()` call.
- **Always** restores the original URL in a `finally` block — credentials are **never persisted to disk**.
- Is a safe no-op for SSH remotes, remotes that already embed credentials, or when the user has no stored token.

Applied the wrapper to all four remote-touching routes.

## Release Notes

* **New Features**
  * Git remote operations (fetch, pull, push, publish) can now authenticate automatically using a retrieved GitHub token when available.
  * Improved discovery for the CLAUDE executable across Unix, Linux and Windows, including architecture-aware detection and more installer/global lookup locations.

### DIFF
--- a/server/routes/git.js
+++ b/server/routes/git.js
@@ -2,12 +2,64 @@ import express from 'express';
 import { spawn } from 'child_process';
 import path from 'path';
 import { promises as fs } from 'fs';
-import { projectsDb } from '../modules/database/index.js';
+import { projectsDb, githubTokensDb } from '../modules/database/index.js';
 import { queryClaudeSDK } from '../claude-sdk.js';
 import { spawnCursor } from '../cursor-cli.js';
 
 const router = express.Router();
 const COMMIT_DIFF_CHARACTER_LIMIT = 500_000;
+
+/**
+ * Injects a GitHub token into a git remote URL so that push/pull/fetch
+ * operations can authenticate without requiring the user to configure
+ * git credentials separately.
+ *
+ * Transforms: https://github.com/user/repo
+ *        into: https://<token>@github.com/user/repo
+ *
+ * If the remote URL is SSH or already has credentials embedded, it is
+ * returned unchanged. If no token is found for the user it is also a no-op.
+ */
+async function withGithubToken(projectPath, userId, fn) {
+  if (!userId) return fn(null);
+
+  // Find the most recently active GitHub token for this user
+  let token;
+  try {
+    token = githubTokensDb.getActiveGithubToken(userId);
+  } catch {
+    token = null;
+  }
+
+  if (!token) return fn(null);
+
+  // Read the current remote URL
+  let originalUrl;
+  try {
+    const { stdout } = await spawnAsync('git', ['remote', 'get-url', 'origin'], { cwd: projectPath });
+    originalUrl = stdout.trim();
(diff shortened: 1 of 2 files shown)


blackmammoth:
Hey @lepisodesangwoo, thanks for the PR! Since the features are independent, it would be much better to review if you submitted a different PR for each feature. As a result, I'm closing this for now.

PR #737, opened 2026-05-04
fix(claude-sync): skip subagent JSONL files to prevent main session corruption

kamioj:
## Summary

`ClaudeSessionSynchronizer` walks `~/.claude/projects/<project>/` for
`.jsonl` files and indexes each one as a session row. Claude Code writes
subagent transcripts under a `subagents/` subdirectory but **shares the
parent session's `sessionId`** in those files. The current scan picks them
up and upserts them into the sessions DB with the same `sessionId`, which
overwrites the parent row and breaks session resume / history listing.

## Reproduction

1. Run a Claude Code session that spawns at least one subagent (Task tool, `/agent`, etc.).
2. Layout becomes:
   ```
   ~/.claude/projects/<encoded-cwd>/<sid>.jsonl                  ← parent
   ~/.claude/projects/<encoded-cwd>/subagents/<sid>-<n>.jsonl    ← subagent (same sid)
   ```
3. Trigger session sync (initial scan or filesystem watcher).
4. The `sessions` row for `<sid>` now reflects the subagent file's metadata; opening the session in the UI shows the subagent transcript instead of the parent.

## Fix

Skip any path containing `/subagents/` (or `\subagents\` on Windows) in:
- `synchronize()` — the bulk scan loop
- `synchronizeFile()` — the watcher-driven entry point

Both code paths use the same cross-platform separator-aware guard.

## Notes

- +15 / -0, single file
- No behavior change for setups without subagents
- No new dependencies / no schema migration

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session synchronization to properly exclude subagent transcript files from being stored, ensuring cleaner data management and preventing redundant session artifacts.

PR #854, opened 2026-06-09
fix(claude-sync): skip subagent transcripts to prevent main session corruption

bourgois:
### What

The Claude session indexer recursively scans `~/.claude/projects` and upserts every `.jsonl` it finds. That recursion also descends into per-session `subagents/` directories. Subagent transcripts **repeat the parent session's `sessionId`**, so when they are indexed as standalone sessions they upsert over the parent row and overwrite its `jsonl_path` — corrupting the main session record (the sidebar then resolves the session to the subagent transcript).

This adds a single `isSubagentTranscript()` guard and applies it in both the recursive scan (`synchronize()`) and the single-file watcher path (`synchronizeFile()`).

### Why the earlier attempt (#737) couldn't be reproduced

#737 listed an incorrect path for the subagent files, so the reproduction pointed at the wrong directory. The actual location (the one you identified when closing #737) is:

```
~/.claude/projects/<encoded-cwd>/<session-id>/subagents/agent-<agent-id>.jsonl
```

`findFilesRecursivelyCreatedAfter` (`server/shared/utils.ts`) recurses into subdirectories, so these files **are** reached by `synchronize()` — this guard targets that exact path.

### Reproduction

1. Start a session and ask Claude to run anything that spawns a subagent (any `Task`/subagent tool call). Claude writes `…/<session-id>/subagents/agent-<id>.jsonl`, and that file's first line carries the **parent** `sessionId` plus the subagent's `cwd`.
2. Trigger a sync (restart the server, or let the watcher pick up the new file).
3. Without this guard: the subagent file is upserted under the parent `sessionId`, overwriting the parent row's `jsonl_path`. The sidebar entry for the session now points at the subagent transcript, so opening it renders subagent content instead of the real conversation.
4. With this guard: subagent files are skipped; the parent row keeps its correct `jsonl_path`.

### Notes

- Single, cross-platform check (`path.normalize(filePath).split(path.sep).includes('subagents')`); shared by both entry points, no duplication.
- `npm run typecheck` and `eslint` pass.
- This has been running in our fork for ~10 days with no regressions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented nested subagent transcript files from being processed as standalone sessions during synchronization.
  * Ensured both single-file and bulk synchronization consistently skip subagent transcript inputs, avoiding transcript overwrites and preserving the integrity of parent sessions.


blackmammoth (approved):
thanks @bourgois!

PR #811, opened 2026-05-30
feat: add scroll navigation for chat messages / 添加聊天消息滚动导航

WenhuaXia:
## Summary / 摘要
- Add scroll navigation bar with dot indicators for each user message — 添加滚动导航栏，每个用户消息对应一个圆点
- Previous/Next buttons for sequential navigation — 上一条/下一条按钮顺序导航
- First/Last buttons for quick jump to start/end — 顶部/底部按钮快速跳转
- Load all button when messages are paginated — 消息分页时显示"加载全部"按钮
- Improve message loading with explicit `loadMoreMessages` callback — 改进消息加载机制
- Fix `pendingUserMessage` visibility with `hasUserMessage` check — 修复待发送消息可见性
- Add i18n strings for scroll navigation in 8 languages — 添加8种语言国际化字符串

## Test plan / 测试
- [ ] Open a long session with many user messages — scroll navigation should appear — 打开长会话应显示导航栏
- [ ] Click dots to jump between messages — 点击圆点跳转消息
- [ ] Test Previous/Next buttons — 测试顺序导航
- [ ] Test First/Last buttons — 测试快速跳转
- [ ] Verify Load All button works when paginated — 验证分页时加载全部功能
- [ ] Navigation should disappear when no user messages — 无用户消息时导航栏应隐藏

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Chat navigation strip with message jump dots, first/previous/next/last buttons, and a "load all" control
  * Exposed an explicit "Load More" action for older messages
  * Plays a completion notification sound when enabled

* **Bug Fixes**
  * Prevents pending user messages from disappearing during realtime streaming
  * Stricter task-notification parsing to reduce false positives

* **Localization**
  * Added translations for navigation and message-loading controls in multiple languages

* **UX**
  * Simplified "load all" indicators and added a concise success feedback indicator

PR #809, opened 2026-05-30
feat: add completion sound notification / 添加完成提示音

WenhuaXia:
## Summary / 摘要
- Add a short audio notification when a chat run finishes — 聊天完成时播放简短提示音
- Users can toggle the sound via the Notifications settings tab — 可在设置 → 通知中开关
- Uses Web Audio API to generate tones, no external assets required — 使用 Web Audio API 生成音效，无需外部资源
- Sound is **enabled by default** — 默认开启
- Includes a "Test sound" button for preview — 包含"试听"按钮

## Test plan / 测试
- [ ] Start a new chat session and wait for AI response to complete — should hear a two-tone sound
- [ ] Navigate to Settings → Notifications → toggle sound off — sound should stop
- [ ] Click "Test sound" button — should hear sound without starting a chat

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  - Notification sound settings to control audio alerts when runs complete
  - New sound settings card with enable/disable toggle and a Test button
  - Plays a completion sound when enabled and when testing
  - Sound preference persisted locally across sessions

* **Localization**
  - Added translations for sound settings in multiple locales (en, zh-CN, de, it, ja, ko, ru, tr)

PR #855, opened 2026-06-09
feat(chat): add scroll navigation for chat messages (decoupled, compiles clean)

bourgois:
Revives #811 (by @WenhuaXia) with the compilation blocker removed. Their commits are preserved; this branch adds one cleanup commit on top.

### Why #811 was closed, and what changed

#811 was closed for a compile failure in `useChatRealtimeHandlers`. The cause was a dependency on the **unmerged** completion-sound PR (#809):

```ts
import { isNotificationSoundEnabled, playCompletionSound } from '../../../utils/notification-sound';
```

`utils/notification-sound` doesn't exist on `main`, so the import fails to resolve. Those `useChatRealtimeHandlers` edits (the sound calls plus an unrelated session-resumption tweak) are **out of scope for scroll navigation**, so this branch reverts that file to `main`. The feature now stands alone.

### What this adds

- `ScrollNavigation.tsx` — a floating rail of user-message markers with jump-to-top/bottom, previous/next, and load-all controls; the active marker tracks the viewport via live DOM offsets.
- Supporting wiring in `ChatInterface`, `ChatMessagesPane`, `useChatSessionState`, `useChatMessages`, and i18n strings for all 8 locales.

### Verification

- `npm run typecheck` passes (this was the original blocker).
- `eslint` reports 0 errors on the touched files.
- `useChatRealtimeHandlers.ts` has **no diff** vs `main` — no dependency on #809.
- Running in our fork for ~10 days.

### Credit

Feature commits are @WenhuaXia's from #811. The only addition is the decoupling/cleanup commit. Happy to squash or adjust attribution however you prefer.

## Summary by CodeRabbit

* **New Features**
  * Added a floating scroll navigation panel with interactive dots and jump controls; tooltips show message previews.

* **Pagination**
  * Unified pagination UI with distinct "Load more" and "Load all" controls and consolidated loading indicators; added safer "load more" action and simplified auto-scroll behavior.

* **Preferences**
  * New quick-settings toggle to show/hide scroll navigation.

* **Internationalization**
  * Added scroll navigation labels and load-more strings across multiple locales.

### DIFF
--- a/src/components/chat/hooks/useChatSessionState.ts
+++ b/src/components/chat/hooks/useChatSessionState.ts
@@ -344,24 +344,13 @@ export function useChatSessionState({
     [hasMoreMessages, isLoadingMoreMessages, selectedProject, selectedSession, sessionStore],
   );
 
-  const handleScroll = useCallback(async () => {
+  const handleScroll = useCallback(() => {
     const container = scrollContainerRef.current;
     if (!container) return;
 
     const nearBottom = isNearBottom();
     setIsUserScrolledUp(!nearBottom);
-
-    if (!allMessagesLoadedRef.current) {
-      const scrolledNearTop = container.scrollTop < 100;
-      if (!scrolledNearTop) { topLoadLockRef.current = false; return; }
-      if (topLoadLockRef.current) {
-        if (container.scrollTop > 20) topLoadLockRef.current = false;
-        return;
-      }
-      const didLoad = await loadOlderMessages(container);
-      if (didLoad) topLoadLockRef.current = true;
-    }
-  }, [isNearBottom, loadOlderMessages]);
+  }, [isNearBottom]);
 
   useLayoutEffect(() => {
     if (!pendingScrollRestoreRef.current || !scrollContainerRef.current) return;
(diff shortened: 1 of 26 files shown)


blackmammoth:
Hey @bourgois, thanks for the PR! Can you explain what issue this PR actually solves (with detailed reproduction step of the issue). In addition, it appears that the PR contains code for many independent issues. Can you send them in small chunks as separate PRs? This would be very useful for reviewing it.

bourgois:
Thanks @blackmammoth, both points are fair.

**What it does.** This is a UX feature, not a bug fix. In long sessions the only way to move through the conversation is free-scrolling — there's no way to jump between your own messages or hop to the top/bottom. This adds a slim navigation rail to the chat pane: a marker per user message (the active one tracks the viewport as you scroll), plus jump-to-top/bottom, previous/next-message, and a "load all" control for paginated history.

How to try it:
1. Open a session with many turns (the message list scrolls well beyond one screen).
2. A vertical rail appears on the right edge of the chat pane, one dot per user message.
3. Scroll the conversation — the highlighted dot follows the message currently in view.
4. Click any dot to jump to that message; use the up/down controls to step between messages or jump to the very top/bottom.
5. On a long session that loads in pages, "load all" pulls the full history in before navigating.

**On splitting — done.** I've force-pushed this branch down to scroll-navigation only. It now touches just `ScrollNavigation.tsx`, its wiring in `ChatInterface`/`ChatMessagesPane`, the `loadMore`/`loadAll` additions in `useChatSessionState`, and the chat i18n strings.

The two unrelated changes that previously rode along are now their own small PRs:
- #862 — task-notification parsing tolerant of field order/whitespace (`useChatMessages.ts`)
- #863 — keep pending user message visible until its echo arrives (`useChatSessionState.ts`)

`npm run typecheck` and `eslint` pass on all three. Credit for the feature is @WenhuaXia's (preserved as co-author).

blackmammoth:
Hey @bourgois, thanks for the explanation. This looks very useful. However, clicking the scroll wheel is a bit difficult with the new setup since it's covering a portion of it. Can you address this?

image

bourgois:
Hi @blackmammoth 
CloudCLI UI 2026-06-10 at 6 20 30 PM
This is how it looks on my Mac... what's your setup?

blackmammoth:
I'm using Chrome on Windows. The UI/UX is a bit clunky and difficult to understand. So that should be addressed.

 In addition, it would be better if this was an optional in settings that is off by default. 

> Hi @blackmammoth CloudCLI UI 2026-06-10 at 6 20 30 PM This is how it looks on my Mac... what's your setup?

bourgois:
Thanks for the feedback @blackmammoth — both points addressed in the latest push.

**1. Opt-in, off by default.** Added a "Show scroll navigation" toggle under Quick Settings → View Options, defaulting to **off**. It's wired through `useUiPreferences` (persisted in `localStorage` like the other view options), so the rail only renders when a user turns it on. Default users see no change. i18n added for all eight locales.

**2. Scrollbar overlap.** The rail was positioned flush at `right-0`, so it sat on top of the message pane's native scrollbar — invisible with macOS overlay scrollbars (my setup), but it covers a classic Windows/Chrome scrollbar and makes it hard to grab (your screenshot). Offset it to `right-3` so the scrollbar stays clear. If the gap looks off on your end I can fine-tune the exact offset — easy to adjust.

`npm run typecheck` and `eslint` pass. Since it's now off by default, the feature is fully opt-in and shouldn't affect anyone who doesn't enable it.

blackmammoth:
@bourgois, thanks for the detailed explanation. Since we are serving users on many d/t devices, a styling inconsistency like this is a difficult thing to explain for users who have many devices. The feature is really cool and we'll try to implement a similar solution better suited the current UI/UX in the future. But, for now, I'm gonna close this PR.

PR #862, opened 2026-06-10
fix(chat): make task-notification parsing tolerant of field order and whitespace

bourgois:
Small, self-contained fix extracted from #855 at the maintainer's request.

### Issue

`normalizedToChatMessages` parses background `<task-notification>` messages with a regex that:
- requires the exact field sequence `task-id` → `output-file` → `status` → `summary`,
- allows no leading/trailing content around the tag, and
- runs with the `/g` flag against the raw (untrimmed) string.

When a notification's fields arrive in a different order, omit an optional field, or carry surrounding whitespace, the match fails and the message renders as **raw XML** in the chat instead of a formatted notification.

### Reproduction

1. Trigger a background task notification whose payload has any leading/trailing whitespace, or whose inner fields are not in the exact `task-id, output-file, status, summary` order.
2. Observed: the chat shows the literal `<task-notification>…</task-notification>` markup.
3. With this change: it renders as the formatted notification (status + summary).

### Fix

Anchor the match to the trimmed content and use non-greedy `[\s\S]*?` between the `status` and `summary` captures, so field order and incidental whitespace no longer break parsing.

`npm run typecheck` and `eslint` pass. One-line change in `useChatMessages.ts`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task notification parsing so background task updates display more consistently: notifications now reliably extract and show the task status and summary, tolerate leading/trailing whitespace, and ignore extraneous formatting. Users should see clearer, more predictable status updates for background tasks.

### DIFF
--- a/src/components/chat/hooks/useChatMessages.ts
+++ b/src/components/chat/hooks/useChatMessages.ts
@@ -55,8 +55,8 @@ export function normalizedToChatMessages(messages: NormalizedMessage[]): ChatMes
 
         if (msg.role === 'user') {
           // Parse task notifications
-          const taskNotifRegex = /<task-notification>\s*<task-id>[^<]*<\/task-id>\s*<output-file>[^<]*<\/output-file>\s*<status>([^<]*)<\/status>\s*<summary>([^<]*)<\/summary>\s*<\/task-notification>/g;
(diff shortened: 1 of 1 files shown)


blackmammoth:
@bourgois this should be fixed with the latest releases.

PR #863, opened 2026-06-10
fix(chat): keep pending user message visible until its echo arrives

bourgois:
Small, self-contained fix extracted from #855 at the maintainer's request.

### Issue

The optimistic "pending" user message (shown immediately after you hit send, before the backend echoes it) is only rendered while the message store is empty:

```ts
if (pendingUserMessage && all.length === 0) {
  return [pendingUserMessage];
}
```

On an **existing** session, the first realtime event after sending can be a `stream_delta` (assistant content) that arrives **before** the echoed user message is committed to the store. The moment that delta lands, `all.length > 0`, so the pending message is dropped — and the assistant's reply briefly renders as the first visible message with the user's prompt missing, until the echo catches up.

### Reproduction

1. Open an existing session that already has messages.
2. Send a new message and watch the message list as the response starts streaming.
3. Observed: the just-sent user message can flicker out and the streaming assistant reply appears first, until the server echo arrives.
4. With this change: the pending user message stays put until its own echoed entry shows up.

### Fix

Gate on the absence of an actual user message (`all.some(m => m.type === 'user')`) rather than an empty store, and append the pending message after existing content so it remains visible until its echo is committed.

`npm run typecheck` and `eslint` pass. Small change in `useChatSessionState.ts`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat message display reliability by fixing a race condition where user messages could temporarily disappear when assistant responses arrived quickly. Messages now remain visible throughout the conversation.

### DIFF
--- a/src/components/chat/hooks/useChatSessionState.ts
+++ b/src/components/chat/hooks/useChatSessionState.ts
@@ -256,9 +256,14 @@ export function useChatSessionState({
 
   const chatMessages = useMemo(() => {
     const all = normalizedToChatMessages(storeMessages);
-    // Show pending user message when no session data exists yet (new session, pre-backend-response)
-    if (pendingUserMessage && all.length === 0) {
-      return [pendingUserMessage];
+    // Show pending user message until a user message actually appears in the store.
+    // This covers both the "no messages yet" case and the race condition where
+    // `stream_delta` (AI content) arrives in realtime before the echoed user
+    // message — without this check the pending message would disappear and the
+    // AI response would briefly appear as the first visible message.
+    const hasUserMessage = all.some((m) => m.type === 'user');
+    if (pendingUserMessage && !hasUserMessage) {
+      return [...all, pendingUserMessage];
     }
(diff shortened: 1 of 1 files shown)


blackmammoth:
Hey @bourgois, tried replicating this on main but couldn't get the issue described here. I used claude and codex for testing purposes. Can you check if the issue still exists?

bourgois:
Thanks for checking @blackmammoth — you're right, and the fault is in my repro.

I traced it: `pendingUserMessage` is only ever set for the **first message of a brand-new session** (`addMessage` only calls `setPendingUserMessage` when `activeSessionId` is null). For any existing session the message goes straight to the store, so the `pendingUserMessage` / `all.length === 0` branch is never hit. My reproduction described an existing session, which can't exercise this path at all — so "couldn't reproduce" is correct.

The only real window is a sub-frame timing race in the new-session path (between the backend assigning a session id and the pending message being flushed to the store), which isn't reliably reproducible by hand and isn't worth your review time.

Closing this. Sorry for the noise. #854, #855, and #862 stand on their own.

PR #614, opened 2026-04-04
fix: finalize Codex sessions without explicit exitCode

CoderLuii:
## Summary
- allow successful complete events without an explicit exitCode to finalize the pending Codex session
- keep the existing explicit exitCode === 0 behavior unchanged when the field is present

## Testing
- validated in HolyClaude integration flow while reproducing the first-turn Codex session redirect issue

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat session handling to properly finalize sessions in additional scenarios where exit status information is absent or indicates success.

PR #879, opened 2026-06-15
fix: include exitCode on Codex completion events

CoderLuii:
## Summary
- Add an explicit `exitCode: 0` to successful Codex `complete` events.
- Keep aborted Codex streams out of the normal success completion path.
- Add `exitCode: 0` to Codex `turn_complete` provider normalization so live and history paths agree.
- Keep the existing `actualSessionId`, `sessionId`, and provider fields unchanged for successful completions.

## Context
Codex was the only provider sending the final successful `complete` event without an explicit success exit code. This is the server-side version of the session finalization problem discussed in #614 and CoderLuii/HolyClaude#19.

This supersedes #614. Current `main` already handles session replacement through `actualSessionId`; this PR makes the provider lifecycle payload consistent at the source.

Abort handling stays separate from success handling. The abort request path already sends an aborted `complete` payload for client cleanup, so `queryCodex` now avoids sending a second normal success completion after an aborted stream.

## Testing
- `node --check server/openai-codex.js`
- `git diff --check`
- `npm ci` completed; npm audit reported 38 existing advisories
- `npm run typecheck`
- `npm run build` passed with existing CSS/chunk/Browserslist warnings
- Compiled `CodexSessionsProvider` `turn_complete` normalization check returned `kind: complete` with `exitCode: 0`

Not live-tested: live Codex abort against the upstream service.

## Release Notes

* **Bug Fixes**
  * Improved completion event handling so “complete” messages are only sent when runs finish normally.
  * Added an explicit exit code (`0`) to completion payloads to make stream termination clearer.

* **Chores**
  * Enhanced abort detection so sessions are consistently marked as aborted when an abort occurs.

### DIFF
--- a/server/modules/providers/list/codex/codex-sessions.provider.ts
+++ b/server/modules/providers/list/codex/codex-sessions.provider.ts
@@ -492,6 +492,7 @@ export class CodexSessionsProvider implements IProviderSessions {
         timestamp: ts,
         provider: PROVIDER,
         kind: 'complete',
+        exitCode: 0,
       })];
     }
     if (raw.type === 'turn_failed') {
--- a/server/openai-codex.js
+++ b/server/openai-codex.js
@@ -245,6 +245,7 @@ export async function queryCodex(command, options = {}, ws) {
   let capturedSessionId = sessionId;
   let sessionCreatedSent = false;
   let terminalFailure = null;
+  let wasAborted = false;
   const abortController = new AbortController();
 
   try {
@@ -311,11 +312,13 @@ export async function queryCodex(command, options = {}, ws) {
 
       // Check if session was aborted
       if (abortController.signal.aborted) {
+        wasAborted = true;
         break;
       }
       if (capturedSessionId) {
         const session = activeCodexSessions.get(capturedSessionId);
         if (session?.status === 'aborted') {
+          wasAborted = true;
           break;
         }
       }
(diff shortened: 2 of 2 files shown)


blackmammoth:
Hey @CoderLuii, thanks for the update. I completely revamped the websocket logic and this issue should be completely resolved now. As a result, I'm closing this PR. Lmk if the issue still occurs tho.

CoderLuii:
@blackmammoth thanks for the websocket rewrite. I checked current `main` at `e885391`, and the general lifecycle contract is better now. There is still one Codex-specific gap, though.

The current live path still lets `turn.completed` become the first terminal message:

1. `server/openai-codex.js` transforms SDK `turn.completed` into raw `turn_complete` inside the stream loop.
2. `server/modules/providers/list/codex/codex-sessions.provider.ts` maps `turn_complete` to `kind: 'complete'` without `exitCode`, `success`, or `aborted`.
3. `server/modules/websocket/services/chat-run-registry.service.ts` marks the first `complete` as terminal and drops the later `createCompleteMessage({ exitCode: 0 })` from `queryCodex`.

I verified that sequence with a small registry probe: first send the Codex-normalized `complete` with no `exitCode`, then send the final helper-built `complete` with `exitCode: 0`. The registry forwards only the first one, so the client never sees the helper-built success payload.

The smallest fix I see is to keep live Codex `turn_complete` out of the terminal websocket path and let `queryCodex` send the one final `createCompleteMessage()` after token budget/status handling. A regression test should cover that ordering directly: Codex sees `turn.completed`, token/status handling runs, and the client receives exactly one final `complete` with `exitCode: 0`, `success: true`, and `aborted: false`.

No need to reopen this PR if you would rather handle it as a small follow-up. I just wanted to flag that current `main` still has this edge case.

PR #892, opened 2026-06-18
fix: force Connection: close on static routes to work around ws keep-alive bug

WenhuaXia:
## Problem

After PR #887 (feat/unify-websocket-2) unified the WebSocket server, the  library (8.18.3) attaches to the HTTP server's `upgrade` event. This interferes with regular HTTP keep-alive connections:

- `/health` and API routes work fine
- Homepage (`/`) and static assets (`/assets/*.js`, `/favicon.png`, etc.) **hang indefinitely** — the server writes the response body but the TCP connection stays half-open, causing clients to timeout

## Root Cause

The `ws` library hooks into `http.Server`'s `upgrade` event for WebSocket handshakes. After this hook is attached, Node's HTTP response `write()`/ completes but the underlying TCP socket does not properly flush and close for keep-alive connections. The client sits waiting for more data until it times out.

## Fix

Set `Connection: close` header on all static file responses (both `/public/*` and `/dist/*` routes). This forces the server to terminate the connection immediately after sending the response, bypassing the keep-alive issue entirely.

## Impact

- **Affected:** Homepage, all frontend JS/CSS bundles, favicons, icons, and other static assets
- **Not affected:** API routes, WebSocket connections, `/health` endpoint
- **Trade-off:** Static assets no longer use HTTP keep-alive (each request opens a new TCP connection). This is acceptable because modern browsers use HTTP/2 multiplexing, and the alternative is the page not loading at all.

## Summary by CodeRabbit

* **Chores**
  * Updated HTTP connection handling in static file serving to improve server stability and reliability.


blackmammoth:
Hey @Wen, thanks for the PR. I tried to replicate this, but there were no issues on both Windows and Linux. `http://localhost:<port>/favicon.png` was also working as expected with no hang. Please recheck and submit the PR again with a reproduction step if the issue is still there.

PR #896, opened 2026-06-18
fix(i18n): add missing sidebar message keys to all locales

wideplain:
## Problem

The sidebar `messages` namespace is missing six keys that are referenced in `src/components/sidebar/hooks/useSidebarController.ts`:

| key | call site |
|---|---|
| `messages.updateProjectError` | rename project / toggle star failure (lines ~486, ~513) |
| `messages.refreshError` | project list refresh failure (line ~552) |
| `messages.restoreProjectFailed` / `messages.restoreProjectError` | restore archived project (lines ~843, ~853) |
| `messages.restoreSessionFailed` / `messages.restoreSessionError` | restore session (lines ~866, ~876) |

`updateProjectError` and `refreshError` are called via `t('messages.…')` **without** an inline default value, so when those actions fail the user sees the raw key string — e.g. a toast/alert literally reading `messages.updateProjectError` — instead of a readable message. This affects every locale.

The four `restore.*` keys do pass an English default as the second `t()` argument, so they don't show the raw key, but they fall back to English even in non-English UIs.

## Fix

Add all six keys to every locale's `sidebar.json` (`de, en, fr, it, ja, ko, ru, tr, zh-CN, zh-TW`), matching the wording/style of the neighbouring `delete*`/`create*` messages already in each file.

No code changes — translation resources only.

## Verification

- All 10 `sidebar.json` files parse as valid JSON.
- Keys placed in the `messages` object so `t('messages.updateProjectError')` etc. now resolve under the default `sidebar` namespace used by the sidebar (`useTranslation(['sidebar', 'common'])`).

## Summary by CodeRabbit

* **Localization**
  * Added error message translations across 10 languages (English, German, French, Italian, Japanese, Korean, Russian, Turkish, Simplified Chinese, and Traditional Chinese) for project and session operations.


blackmammoth (approved):
@wideplain Thanks for the PR! Also, if you’d like to join project discussion or community chat, we have a Discord here: [link](https://discord.gg/buxwujPNRE). Totally optional.

PR #917, opened 2026-06-24
Fix Windows browser-runtime install and Claude Code SDK spawn failures

benjaminberes-bp:
## What

Three Windows-only failures in the Browser runtime installer and the Claude Code SDK integration, plus a standalone launcher script.

### 1. `spawn EINVAL` when installing the Browser runtime
`server/modules/browser-use/browser-use.service.ts` spawned `npm.cmd` with `shell: false`. Since Node's fix for CVE-2024-27980 (v18.20.2 / v20.12.2+), spawning a `.cmd`/`.bat` without a shell throws `spawn EINVAL`, so **Install Runtime** failed immediately on Windows. Fix: `shell: process.platform === 'win32'` (args are static literals, no injection surface).

### 2. `ECONNRESET` / `EPERM rmdir` during Browser runtime install
After EINVAL was fixed, the installer ran `npm install` in the **project root**, forcing npm to reconcile the entire dependency tree — re-fetching unrelated tarballs (`parse-github-url` → `ECONNRESET`) and racing the running dev server for locked `node_modules` dirs (`EPERM rmdir eslint/nypm`). Fix: install Playwright into an isolated dir (`~/.cloudcli/browser-use/runtime`) with its own `package.json` and resolve it from there. Install drops to just `playwright` + `playwright-core`, no tree churn, no locks. Project `node_modules` stays a dev fallback.

### 3. `Claude Code native binary exists but failed to launch`
Misleading error — the binary is fine. When a session's project directory no longer exists on disk, `server/claude-sdk.js` passed it to the SDK as `cwd`. On Windows `child_process.spawn` throws `ENOENT` when `cwd` is invalid, which the Agent SDK reports as a binary launch failure. Fix: guard `cwd` with `statSync().isDirectory()`; if missing, log a warning and let the SDK default to the server's cwd so the run still launches.

### 4. `start-cloudcli.bat`
Standalone launcher: frees port 3001 if a previous instance lingers, then runs `npm run dev` (server :3001 + client :5173).

## Why
All three bugs make core features unusable on Windows (Node 18.20.2+ / 20.12.2+, reproduced on Node 25). They surface as cryptic errors that don't point at the real cause.

## Notes for reviewers
- Each fix is an isolated commit.
- Verified on Windows 11, Node v25.9.0: `tsc --noEmit` clean, eslint clean, and each failure mode reproduced before/after.
- The isolated-runtime install was validated end-to-end (Playwright + Chromium resolve, `chrome.exe` exists).

## Summary by CodeRabbit

* **New Features**
  * Added a Windows launcher that starts the app, shows the local server/UI URLs, and keeps the console open.

* **Bug Fixes**
  * Improved handling of invalid working directories so the app falls back to a safe default instead of using a missing path.
  * Made browser/runtime setup more reliable by using an isolated install location and better Windows command handling.

### DIFF
--- a/server/claude-sdk.js
+++ b/server/claude-sdk.js
@@ -14,7 +14,7 @@
 
 import { query } from '@anthropic-ai/claude-agent-sdk';
 import crypto from 'crypto';
-import { promises as fs } from 'fs';
+import { promises as fs, statSync } from 'fs';
 import path from 'path';
 import os from 'os';
 import { CLAUDE_FALLBACK_MODELS } from './modules/providers/list/claude/claude-models.provider.js';
@@ -40,6 +40,14 @@ const abortedSessionIds = new Set();
 const TOOL_APPROVAL_TIMEOUT_MS = parseInt(process.env.CLAUDE_TOOL_APPROVAL_TIMEOUT_MS, 10) || 55000;
 
 const TOOLS_REQUIRING_INTERACTION = new Set(['AskUserQuestion', 'ExitPlanMode']);
+
+function isExistingDirectory(dir) {
+  try {
+    return statSync(dir).isDirectory();
+  } catch {
+    return false;
+  }
+}
 
 function createRequestId() {
   if (typeof crypto.randomUUID === 'function') {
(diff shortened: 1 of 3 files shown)


CoderLuii:
@benjaminberes-bp I hit the same Browser Use installer problem while checking HolyClaude's container setup. Your isolated runtime directory and Windows npm.cmd fix here were on target.

I refreshed that Browser Use part against current main in #995, kept the Windows spawn handling, and added tests for installation, module resolution, localhost navigation, snapshots, and cleanup. I also ran it from a clean global CloudCLI install on Node 24. The PR description credits this work directly.

I did not carry over the Claude SDK cwd change or the batch launcher, so #995 is not meant to take ownership of those parts. I kept it narrow because this branch now conflicts with main and the runtime fix can be reviewed on its own.

Thanks for documenting the Windows failure clearly. It made the overlap easy to verify.

PR #923, opened 2026-06-26
fix: resolve mobile shell issues

### LINKED ISSUE #389: Shell tab doesn't auto-scroll to bottom on tab switch
## Description

When clicking on the Shell tab in a session that already has terminal output, the scroll position starts at the top. Users must manually scroll to the bottom to see the latest output.

**Expected behavior:** The shell tab should auto-scroll to the bottom (latest output) when switching to it.

## Root Cause

The Shell tab in `MainContent.tsx` uses **conditional rendering** (`activeTab === 'shell' && ...`), which destroys and recreates the `StandaloneShell` component on every tab switch. This causes the terminal to reconnect and replay the buffer, but the scroll position is not at the bottom due to a timing issue between `fitAddon.fit()` and data arrival.

In contrast, the Chat tab already uses CSS visibility (`block`/`hidden` classes) to keep the component mounted.

## Suggested Fix

Two changes:

### 1. `src/components/main-content/view/MainContent.tsx`

Change the shell tab from conditional rendering to CSS visibility (matching the pattern already used by the Chat tab):

```diff
-          {activeTab === 'shell' && (
-            <div className="h-full w-full overflow-hidden">
-              <AnyStandaloneShell project={selectedProject} session={selectedSession} showHeader={false} />
-            </div>
-          )}
+          <div className={`h-full w-full overflow-hidden ${activeTab === 'shell' ? 'block' : 'hidden'}`}>
+            <AnyStandaloneShell project={selectedProject} session={selectedSession} showHeader={false} />
+          </div>
```

### 2. `src/components/Shell.jsx`

Add `scrollToBottom()` after `fitAddon.fit()` in the ResizeObserver callback as a safety net:

```diff
     const resizeObserver = new ResizeObserver(() => {
       if (fitAddon.current && terminal.current) {
         setTimeout(() => {
           fitAddon.current.fit();
+          terminal.current.scrollToBottom();
           if (ws.current && ws.current.readyState === WebSocket.OPEN) {
```

## Benefits

- Shell WebSocket connection persists across tab switches (no reconnection delay)
- Terminal state and scroll position are preserved
- Consistent with how Chat tab is already rendered
- `scrollToBottom()` ensures correct scroll position when terminal becomes visible after being hidden

### LINKED ISSUE #842: Vertical scroll broken in right-side shell panel
Describe the bug
The vertical scroll function on the right-side shell session area is broken. Unable to scroll up to view historical command outputs; the content is stuck within the current viewport.
To Reproduce
Open the web page containing the shell session panel
Use the shell and generate multiple lines of output/command history
Attempt to scroll up to view earlier history
Notice scrolling does not work, content cannot be viewed beyond current screen
Expected behavior
Able to scroll up and down normally to browse all shell session historical content.
Screenshots
(Attach screenshots here if you have)
Error message
No visible error messages displayed on the page.
Desktop (please complete the following information):
OS: Windows 10 1909
Browser:chrome148.0.7778.179
Version:v1.33.1
Smartphone (please complete the following information):
Device: N/A
OS: N/A
Browser: N/A
Additional context
The scrollbar exists but is unresponsive. The shell session content is locked to the current view and cannot be scrolled vertically to access earlier history records.

### LINKED ISSUE #908: Shell touch scroll broken
While using using cloud CLI the shell no longer will scroll upwards from using Chrome in my Android phone. It worked previously on earlier versions.

blackmammoth:
Supersedes #615. Closes #615.

## Summary

The reason it supersedes 615 is because we want it to stop intercepting plain c in shell, but introducing `Ctrl/Cmd+Shift+C` is confusing for the copy behavior since claude even itself suggests to use `c` for copy. This PR essentially makes the copy and input identical to a normal terminal without no CTRL/CMD commands required.

In addition, the PR makes shell output selectable on mobile by adding a touch-aware selection manager for the xterm terminal surface. Mobile users can now long-press shell output, create a terminal selection, and adjust it with draggable handles. Because terminal output is now directly selectable, this PR also removes the separate mobile auth URL UI with copy/hide controls. Auth URLs stay in the normal shell output, where users can copy them the same way they copy any other terminal text.

## Why

Mobile shell output previously behaved differently from normal selectable text. Long-pressing terminal output did not create a useful selection, which made auth URLs and command output hard to copy on phones.

The old workaround added special frontend behavior for auth URLs, including a mobile-only panel and copy/hide controls. That solved one narrow case but duplicated information already present in the terminal and added separate state for a flow that should work through normal shell interaction.

This change fixes the underlying mobile selection behavior instead, then removes the extra auth URL UI and key interception.

## What Changed

- Added a scoped mobile terminal selection manager for touch/coarse-pointer environments.
- Supports long-press selection directly inside the shell terminal.
- Adds draggable start/end handles so mobile users can adjust the selected range.
- Keeps the selection implementation isolated to shell terminal code, without `index.css` changes.
- Removes the special-case auth URL copy behavior that intercepted plain `c` input.
- Removes the mobile auth URL panel, copy state, hide behavior, and related runtime plumbing.
- Keeps auth URLs as normal terminal output instead of duplicating them in frontend-only UI.

## Validation

- `git diff --cached --check`
- Pre-commit hook ran ESLint successfully on the staged shell files
- `npm run typecheck`
- `npm run build:client`

## Notes

This intentionally keeps the solution focused on making shell output selectable. It does not add a new global stylesheet path or preserve the old mobile auth URL controls, since those controls are no longer needed once terminal text can be selected and copied directly.

## Summary by CodeRabbit

* **New Features**
  * Added touch-friendly terminal selection on mobile (long-press selection, draggable handles, context menu actions) with pinch-to-zoom and inertial scrolling.
  * Added a mobile shortcut to send **Ctrl+C** to the terminal when connected.
* **Bug Fixes / Improvements**
  * Simplified the minimal shell view to render only the terminal.
  * Tightened lifecycle cleanup for terminal, subscriptions, and resize handling.
  * Prevented page scrolling artifacts and updated viewport behavior for mobile.
* **Refactor**
  * Removed auth-URL open/copy flows and associated wiring from shell runtime/terminal components and shared types.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,7 +4,7 @@
     <meta charset="UTF-8" />
     <link rel="icon" type="image/svg+xml" href="/favicon.svg" />
     <link rel="icon" type="image/png" href="/favicon.png" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0, user-scalable=no, viewport-fit=cover" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0, user-scalable=no, viewport-fit=cover, interactive-widget=resizes-content" />
     <title>CloudCLI UI</title>
     
     <!-- PWA Manifest -->
--- a/src/components/shell/constants/constants.ts
+++ b/src/components/shell/constants/constants.ts
@@ -1,6 +1,5 @@
 import type { ITerminalOptions } from '@xterm/xterm';
 
-export const CODEX_DEVICE_AUTH_URL = 'https://auth.openai.com/codex/device';
 export const SHELL_RESTART_DELAY_MS = 200;
 export const TERMINAL_INIT_DELAY_MS = 100;
 export const TERMINAL_RESIZE_DELAY_MS = 50;
--- a/src/components/shell/hooks/useShellConnection.ts
+++ b/src/components/shell/hooks/useShellConnection.ts
@@ -24,7 +24,6 @@ type UseShellConnectionOptions = {
   autoConnect: boolean;
   closeSocket: () => void;
   clearTerminalScreen: () => void;
-  setAuthUrl: (nextAuthUrl: string) => void;
   onOutputRef?: MutableRefObject<(() => void) | null>;
 };
 
@@ -49,7 +48,6 @@ export function useShellConnection({
   autoConnect,
   closeSocket,
   clearTerminalScreen,
-  setAuthUrl,
   onOutputRef,
 }: UseShellConnectionOptions): UseShellConnectionResult {
   const [isConnected, setIsConnected] = useState(false);
@@ -100,14 +98,8 @@ export function useShellConnection({
         return;
       }
 
-      if (message.type === 'auth_url' || message.type === 'url_open') {
-        const nextAuthUrl = typeof message.url === 'string' ? message.url : '';
-        if (nextAuthUrl) {
-          setAuthUrl(nextAuthUrl);
-        }
-      }
     },
-    [handleProcessCompletion, onOutputRef, setAuthUrl, terminalRef],
+    [handleProcessCompletion, onOutputRef, terminalRef],
   );
 
   const connectWebSocket = useCallback(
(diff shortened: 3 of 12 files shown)


viper151:
@blackmammoth when testing that on iOS, the terminal text can be selected indeed however there is 1 issue. Selecting the text does not show any text options so you can copy the selected text. Pressing C on the keyboard (as per Claude Code) also does not copy the text (plus it would only solve it for Claude Code perhaps?)

blackmammoth:
hello @viper151, you can test it again. The shell should feel much more responsive now.

viper151 (approved):

PR #933, opened 2026-06-29
fix: preview video on new tab

### LINKED ISSUE #795: [Enhancement] Open HTML files from the Files tree in a new browser tab (preview mode)
## Use case

Working in a project that generates standalone HTML reports — backtest summaries, playwright reports, jupyter exports, diagram outputs — the only way to view them is to download via the right-click menu and open from local fs, or rely on a `cd && python -m http.server` running on the side. Both flow-break the editor.

Would be much nicer to **single-click an `.html` in the Files tree** and have it open in a new browser tab, rendered, with its relative CSS/JS/img assets resolving correctly.

### Backend: new route `GET /api/projects/:projectId/preview/*`

- Auth: existing `authenticateToken` (already supports `?token=` query for SSE), plus an `expandCookieToQuery` middleware that reads a scoped HttpOnly cookie so subsequent relative-asset fetches don't need the token in every URL
- HTML files: inject `<base href="./">` after `<head>` so relative refs resolve back to `/preview/`; serve as `text/html`
- Non-HTML files (CSS / JS / images): stream raw bytes with `mime.lookup()`-detected Content-Type
- Path safety: same `startsWith(projectRoot)` guard as `/files/content`

Verified working request matrix in my prototype:

| # | Scenario | Result |
|---|---|---|
| 1 | `?token=` query → HTML | 200, `<base href="./">` injected, `Set-Cookie` issued |
| 2 | Cookie-only HTML (page navigation back) | 200 |
| 3 | Cookie-only CSS / JS (relative asset fetch) | 200 with correct mime |
| 4 | No auth | 401 |
| 5 | `../../../etc/passwd` traversal | 403 |

### Frontend: capture-phase click listener

Walks React **Fiber** tree (not DOM-exposed `__reactProps$`, those don't carry the `file` object — it's closure-captured in `onClick`). At each fiber, checks `memoizedProps` for `{file: {path, projectId}}`. If found and path matches `\.html?$`, `preventDefault + stopPropagation + window.open('/api/projects/{id}/preview/{path}?token=...')`.

This shape was discovered empirically — the row component renders `<div onClick={() => openFile(file)}>...</div>` and the `file` prop never lands on the rendered DOM node, only on the fiber.

```js
// extractor — works against current 1.32.0 bundle
const cands = [p.file, p.item, p.node, p.entry, p.data, p.row, p.record];
```

## Why this needs to be upstream

I'm running this as a patched-on-startup hotfix locally (sed-injection into `dist/index.html` + new payload file in `dist/`), but it'd be 10× cleaner as a first-class feature, especially since:

- The right-click context menu already has "Download" — "Open in browser" is the obvious sibling
- The `/files/content` route already does 80% of the work (mime + auth + path guard); preview just adds `<base>` injection + cookie path
- Doesn't affect non-HTML default editor behavior at all

Happy to send a PR if there's interest. Full local implementation is ~200 lines (route handler + middleware + frontend hook).

## Environment

- Package: `@cloudcli-ai/cloudcli@1.32.0` (downstream rebrand)
- Browser: Firefox 151.0.1, Chromium 130 — both work
- OS: Linux

blackmammoth:
## Summary

Closes and supersedes PR #930. Closes PR #930.

This keeps the existing media preview/new-tab flow intact and fixes the video-specific case where opening a preview in a new tab downloaded the file as binary instead of rendering it.

## Changes

- Preserve the existing blob URL behavior used by PDF, audio, image, and video previews.
- Treat previewable video files as mislabeled when their fetched blob type is not `video/*`.
- Reuse the existing extension-based MIME fallback so MP4 and other supported video previews are exposed to the browser with a playable video MIME type.

## Summary by CodeRabbit

* **New Features**
  * Added inline native previews for supported media files (images, PDFs, video, audio) in the code editor.
  * Added HTML preview support with a header button to open the HTML content in a new tab.
  * Preview UI includes open-in-new-tab (when available), fullscreen, and close actions.
* **Bug Fixes**
  * Prevented saving for preview-only media and other binary files.
  * Improved preview reliability to avoid incorrect previews when switching files.
* **Documentation**
  * Added user-facing preview loading/error messaging and updated preview button labels.

### DIFF
--- a/src/components/code-editor/hooks/useCodeEditorDocument.ts
+++ b/src/components/code-editor/hooks/useCodeEditorDocument.ts
@@ -2,6 +2,7 @@ import { useCallback, useEffect, useState } from 'react';
 import { api } from '../../../utils/api';
 import type { CodeEditorFile } from '../types/types';
 import { isBinaryFile } from '../utils/binaryFile';
+import { getPreviewKind } from '../utils/previewableFile';
 
 type UseCodeEditorDocumentParams = {
   file: CodeEditorFile;
@@ -23,6 +24,9 @@ export const useCodeEditorDocument = ({ file, projectPath }: UseCodeEditorDocume
   const [saveSuccess, setSaveSuccess] = useState(false);
   const [saveError, setSaveError] = useState<string | null>(null);
   const [isBinary, setIsBinary] = useState(false);
+  // Some binaries (images, PDFs, audio, video) can be rendered natively, so the
+  // editor shows an inline preview instead of the generic binary placeholder.
+  const previewKind = getPreviewKind(file.name);
   // `fileProjectId` is the DB primary key passed down from the editor sidebar;
   // the fallback to `projectPath` preserves older callers that didn't yet
   // propagate the identifier.
@@ -37,6 +41,13 @@ export const useCodeEditorDocument = ({ file, projectPath }: UseCodeEditorDocume
       try {
         setLoading(true);
         setIsBinary(false);
+
+        // Natively previewable media (image/pdf/audio/video) is rendered by
+        // CodeEditorMediaPreview, so there is nothing to read as text here.
+        if (getPreviewKind(file.name)) {
+          setLoading(false);
+          return;
+        }
 
         // Check if file is binary by extension
         if (isBinaryFile(file.name)) {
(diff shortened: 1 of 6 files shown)


blackmammoth:
hey @viper151, this is ready for review.

viper151 (approved):

Issue #747, opened 2026-05-06
Read `custom-title` entries from session JSONL so Claude Code `/rename` is reflected in the sidebar


felipemeres:
### Summary

Sessions renamed in Claude Code (via the `/rename` command) are not reflected in the CloudCLI sidebar. The Claude session synchronizer builds its name map only from `~/.claude/history.jsonl` (which holds first-prompt-per-session entries), but `/rename` writes the new title into the session's own JSONL transcript instead. The rename is durable on disk, but invisible to CloudCLI — so a user juggling several parallel sessions has no way to surface their per-session labels in the UI.

### Reproduce

- CloudCLI UI v1.31.5 (npm install)
- Claude Code v2.1.131
- Node v24.14.1, Ubuntu 24.04

1. Start a session in `claude`, send a message, then run `/rename Foo`.
2. `grep custom-title ~/.claude/projects/<encoded-cwd>/<sessionId>.jsonl` shows the rename was persisted:

   ```json
   {"type":"custom-title","customTitle":"Foo","sessionId":"…"}
   {"type":"agent-name","agentName":"Foo","sessionId":"…"}
   ```
3. `grep "Foo" ~/.claude/history.jsonl` returns nothing — `/rename` does not append to `history.jsonl`.
4. Open CloudCLI — the session shows either the first user prompt or `Untitled Claude Session`. Never `Foo`.

### Root cause

`server/modules/providers/list/claude/claude-session-synchronizer.provider.ts:31`:

```js
const nameMap = await buildLookupMap(
  path.join(this.claudeHome, 'history.jsonl'),
  'sessionId',
  'display',
);
```

`history.jsonl` is a per-prompt log written by the CLI as the user types. `/rename` does not touch it. The actual rename lives in the session JSONL as a `{"type":"custom-title", "customTitle":"…"}` entry (and a redundant `agent-name` entry).

This is the same architectural seam called out in #731 — the synchronizer is built around `history.jsonl` as the only name source, and gaps appear whenever Claude Code persists a name through some other path.

### Suggested fix

`processSessionFile()` already iterates the session JSONL. It can pick up the `custom-title` entry (preferring the latest one if multiple exist, since users can rename more than once) and prefer it over the `history.jsonl`-derived first-prompt fallback.

Suggested precedence for the displayed session name:

1. CloudCLI UI rename (`sessions.custom_name` in CloudCLI's SQLite) — already wins, unchanged
[...]

PR #982, opened 2026-07-10
fix: session title priority — JSONL ai-title/custom-title over history.jsonl display

swu45:
## Problem

CloudCLI sidebar shows the user's first typed message as the session title, instead of the AI-generated title that Claude Code produces. For example:

| What | Content |
|---|---|
| CloudCLI sidebar | `How to set up a React project with Tailwind CSS?` (user's raw prompt) |
| Claude Code `/resume` | `React + Tailwind Project Setup Guide` (AI-generated title) |

### Root cause — two bugs

**Bug 1: `processSessionFile` checks `history.jsonl` before JSONL**

`history.jsonl`'s `display` field is a snapshot of user input—it was never a title. `buildLookupMap` uses first-write-wins, so `nameMap.get()` always returns the user's first message for CLI-started sessions. Since this check runs before `extractSessionAiTitleFromEnd`, the JSONL `ai-title` and `custom-title` events are never consulted.

**Bug 2: `extractSessionAiTitleFromEnd` reverse-scan stops at `last-prompt`**

The reverse-scan loop returns on the *first* match. Because `last-prompt` always appears at the end of the JSONL file, it is always the first match, shadowing `ai-title` and `custom-title` that appear earlier.

### 1. Swap JSONL and history.jsonl priority in `processSessionFile`

JSONL title extraction (`extractSessionAiTitleFromEnd`) now runs before the `history.jsonl` lookup, so AI-generated and `/rename` titles take precedence.

### 2. Full scan in `extractSessionAiTitleFromEnd` with proper priority

Changed from a reverse-scan-early-return to a full forward scan that collects all three title types separately, then returns them with the correct priority:

```
custom-title > ai-title > last-prompt
```

### Final title resolution chain

```
custom_name (DB, user renamed via sidebar)
  → custom-title (JSONL, /rename in CLI)
    → ai-title (JSONL, Claude Code auto-generated)
      → last-prompt (JSONL, last user message)
        → history.jsonl display (first user message, fallback)
          → "Untitled Claude Session"
```

## Files changed

| File | Change |
|---|---|
| `server/modules/providers/list/claude/claude-session-synchronizer.provider.ts` | Fix both bugs |
| `server/modules/providers/tests/claude-sessions.test.ts` | 12 unit tests covering all title sources, priority, and edge cases |
| `.gitignore` | Add `issue/` for local tracking |

## Related

- #747 — added `custom-title` support to `extractSessionAiTitleFromEnd` (didn't fix the priority bug blocking it)
- #731 — sessions showing "Untitled Claude Session" (same root cause for some cases)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Claude session title selection by first using transcript events (custom title, AI title, then last prompt).
  * Added a more reliable fallback to history-based titles when transcript data is missing.
  * Preserves existing meaningful custom names, while replacing default “Untitled” labels with better derived titles.
  * Ensures sessions with no usable title info consistently show “Untitled Claude Session.”

### DIFF
--- a/server/modules/providers/list/claude/claude-session-synchronizer.provider.ts
+++ b/server/modules/providers/list/claude/claude-session-synchronizer.provider.ts
@@ -145,9 +145,9 @@ export class ClaudeSessionSynchronizer implements IProviderSessionSynchronizer {
       };
     }
 
-    let sessionName = nameMap.get(parsed.sessionId);
+    let sessionName = await this.extractSessionAiTitleFromEnd(filePath, parsed.sessionId);
     if (!sessionName) {
-      sessionName = await this.extractSessionAiTitleFromEnd(filePath, parsed.sessionId);
+      sessionName = nameMap.get(parsed.sessionId);
     }
 
     return {
@@ -164,7 +164,11 @@ export class ClaudeSessionSynchronizer implements IProviderSessionSynchronizer {
       const content = await readFile(filePath, 'utf8');
       const lines = content.split(/\r?\n/);
 
-      for (let index = lines.length - 1; index >= 0; index -= 1) {
+      let foundCustomTitle: string | undefined;
+      let foundAiTitle: string | undefined;
+      let foundLastPrompt: string | undefined;
+
+      for (let index = 0; index < lines.length; index += 1) {
         const line = lines[index]?.trim();
         if (!line) {
           continue;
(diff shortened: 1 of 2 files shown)


swu45:
@coderabbitai full review

aldredb:
I hit this bug independently and worked out the same cause before finding this PR, so here is a confirmation plus an offer of help.

**Still reproducing.** On the published `1.37.2`, every Claude session in the sidebar shows its first prompt instead of its title. The reason is the one described here: `history.jsonl` is consulted before the transcript, and `buildLookupMap` keeps the first value it sees for a key. Since any session with at least one prompt appears in `history.jsonl`, `extractSessionAiTitleFromEnd` is effectively unreachable. I checked the file at `ref=main` today and both defects are unchanged, so this PR is not stale in substance — only in its merge base.

There is a third symptom worth recording. Claude Code writes `custom-title` immediately *before* `ai-title`, so even when the fallback is reached, a reverse scan returning its first hit will always prefer the auto-generated title over a manual `/rename`. The forward scan here fixes that too.

**On the conflict.** Only `main`'s newer `claude-sessions.test.ts` conflicts; the provider file still merges cleanly. I rebased onto current `main` (`c1be241`), resolved it by keeping both disjoint test sets, and opened it against this branch as swu45/claudecodeui#1 — merging that makes this PR mergeable again with @swu45's commits and authorship intact. It also renames `extractSessionAiTitleFromEnd`, which addresses the review comment above, as a separate commit that can be dropped.

After the rebase: 22/22 in that file, 407/409 in the full server suite, typecheck clean, no new lint warnings. The single failure is `conversation search streams title matches before transcript results`, which fails identically on unmodified `main`.

@swu45, are you still able to land this? If you would rather not, I am happy to open the rebased branch directly against `main` with your commits preserved — just say the word.

Two related things for whoever picks this up: #1178 rewrites the same function to bound the scan for performance, so these two will need sequencing; and #747 covers the `custom-title` half of this and was closed as completed, though the reporter said the next day that it still reproduced.

aldredb:
Following up on my comment above: I have opened the rebased branch directly against `main` as #1258, since this PR has been conflicting for a while.

@swu45's two commits are unchanged there and keep their authorship. swu45/claudecodeui#1 is still open against this branch, so if @swu45 would rather land it here, merging that makes this PR mergeable again and #1258 can be closed. Either route is fine by me.

blackmammoth:
hey @swu45, @aldredb has addressed it and so we have closed this PR. Thank you @aldredb

PR #986, opened 2026-07-10
fix(shell): enable native touch scrolling in the mobile terminal

### LINKED ISSUE #954: iPhone: Shell terminal scrolling is janky — non-passive touch listeners disable threaded scrolling; custom JS inertia and DOM-renderer fallback compound it
### Environment
- Version: v1.35.1 (`d8dfb2c`)
- Reproduced on: iPhone / iOS Safari (touch scrolling in the Shell terminal).

### Summary
Scrolling the Shell terminal on iPhone **stutters / is janky**, especially while the terminal is producing output. The root cause is that all touch handling for the terminal is registered as **non-passive** listeners, which forces every scroll frame to block on the main thread; this is compounded by a custom JS inertia implementation and by falling back to xterm's slow DOM renderer when WebGL is unavailable.

### Root cause 1 — non-passive touch listeners disable threaded scrolling
`src/components/shell/utils/mobileTerminalSelection.ts` L273-284 registers `touchstart`/`touchmove`/`touchend`/`touchcancel` on the terminal element all with `{ passive: false }`:

```js
this.terminal.element.addEventListener('touchmove', this.onTerminalTouchMove, { passive: false });
// ...touchstart / touchend / touchcancel likewise, all passive: false
```

A single non-passive `touchmove` listener forces the browser to **wait for the main-thread JS handler on every scroll frame** (it must know whether the handler will `preventDefault`), which disables the compositor's off-main-thread smooth scrolling. During terminal output the main thread is busy parsing + rendering PTY data, so the scroll visibly stutters.

`preventDefault()` is actually only needed while a selection or pinch-zoom is in progress; plain one-finger scrolling takes the `recordScrollSample` path (L376-379) which does **not** call `preventDefault`. So the listeners could be `{ passive: true }` by default and only switch to non-passive while a selection/pinch gesture is active.

### Root cause 2 — custom JS inertia fights xterm and can stop abruptly
`mobileTerminalSelection.ts` L787-824 implements its own fling/inertia by mutating `viewport.scrollTop` in a `requestAnimationFrame` loop after touchend. Each frame triggers an xterm scroll + row re-render on the main thread, so the inertia itself stutters during output. The termination check at L814 (`if (viewport.scrollTop === before)`) can also misfire once per-frame delta rounds below 1px, cutting the coast off abruptly instead of easing to a stop.

### Root cause 3 — misleading "Canvas fallback"; real fallback is the slow DOM renderer
`src/components/shell/hooks/useShellTerminal.ts` L101-105:

```js
try {
  nextTerminal.loadAddon(new WebglAddon());
} catch {
  console.warn('[Shell] WebGL renderer unavailable, using Canvas fallback');
}
```

`@xterm/addon-canvas` is **not** a dependency (`package.json` L155-159 lists only clipboard/fit/web-links/webgl/xterm), so on WebGL failure xterm silently uses its built-in **DOM renderer**, not canvas. DOM rendering is markedly heavier per scroll frame on iPhone, worsening the jank. The log message is misleading.

### Suggested fixes
1. Register the terminal touch listeners as `{ passive: true }` by default and only add non-passive handling (or set CSS `touch-action`) while a selection or pinch gesture is active — so plain scrolling stays on the compositor thread.
2. Reconsider the custom inertia: either drop it in favor of native momentum scrolling, or fix the L814 termination to `Math.abs(delta) < 1` so it eases out instead of snapping.
3. Add `@xterm/addon-canvas` and wire it as the real fallback (and correct the log message), so non-WebGL environments don't land on the DOM renderer.

mayankdebnath:
## Summary

One-finger scrolling in the web Shell terminal is broken on iOS Safari and janky
on other touch platforms. This makes the fix from #954: it moves the terminal
back to native, compositor-driven scrolling instead of the current
non-passive/JS-driven approach.

Closes #954.

## The bug

On iOS Safari you cannot drag-scroll the terminal at all; on Android and other
touch platforms scrolling stutters, especially while the terminal is producing
output.

## Root cause

1. **`user-select: text` hijacks the drag.** `src/index.css` forced
   `-webkit-user-select: text !important` on `.xterm`/`.xterm-viewport`. On iOS
   Safari that turns a one-finger drag into a text-selection gesture instead of a
   scroll — and with the WebGL/canvas renderer there is no selectable DOM text to
   grab, so the drag does nothing at all. The terminal's own selection is done
   through xterm's selection model (see `mobileTerminalSelection.ts`), not the
   DOM, so this rule was hurting more than it helped.
2. **Non-passive touch listeners disable threaded scrolling.** Every terminal
   touch listener was registered `{ passive: false }`, forcing the browser to
   block each scroll frame on the main thread — which is busy parsing/rendering
   PTY output — so scrolling stutters. (Root cause #1 from #954.)
3. **Custom JS inertia fights the browser.** A `requestAnimationFrame` loop
   mutated `viewport.scrollTop` after `touchend`, competing with the browser's
   own momentum and snapping to a stop instead of easing. (Root cause #2 from
   #954.)
4. **Misleading "Canvas fallback".** The WebGL addon was loaded *before*
   `terminal.open()`, so it always threw and the terminal silently fell back to
   xterm's slow **DOM** renderer. `@xterm/addon-canvas` wasn't even a dependency,
   so the "Canvas fallback" log was wrong. (Root cause #3 from #954.)

## The fix

- **Native scrolling via CSS.** Drop the forced `user-select: text`; add
  `touch-action: pan-y`, `overscroll-behavior: contain` and
  `-webkit-overflow-scrolling: touch` on `.xterm-viewport`. Native text selection
  is re-enabled only while a custom selection gesture is active
  (`.shell-mobile-selecting`). `overscroll-behavior: contain` also stops the page
  from rubber-banding while you scroll the terminal.
- **Passive `touchmove` + `touch-action` arbitration.** The terminal `touchmove`
  listener is now passive, so one-finger scrolling stays on the compositor
  thread. Selection/pinch gestures no longer `preventDefault`; instead the
  terminal's `touch-action` flips to `none` while selecting, dragging a handle,
  or pinching (`resolveTerminalTouchAction`). `touchstart`/`touchend` stay
  non-passive so pinch-start and the keyboard-suppression on selection still
  work.
- **Native momentum.** The custom JS inertia is removed; the natively-scrolled
  viewport provides momentum for free.
- **Real renderer fallback.** Load the renderer *after* `terminal.open()`, add
  `@xterm/addon-canvas` as the real fallback (with WebGL context-loss recovery),
  and fix the log message. WebGL is preferred, canvas is the fallback, and the
  slow DOM renderer is only a last resort.

No regression is intended to long-press selection, selection-handle dragging,
pinch-to-zoom font sizing, mouse/wheel scrolling, or tap-to-focus + keyboard —
those paths are preserved; only the scroll mechanism changed.

## Tests / verification

- `npm run typecheck` — pass
- `npm run lint` — pass (0 errors)
- Full `node:test` suite (client + server) — pass, including new unit tests for
  the `touch-action` gesture resolution and the move-threshold helper
  (`src/components/shell/utils/mobileTerminalSelection.test.ts`)
- `npm run build` — pass
- App boots and serves the built client

**On-device note (please read):** this is a WebKit/iOS-specific interaction — a
desktop browser (even with touch emulation) scrolls fine with the *old* CSS, so
an emulated before/after does not reproduce the broken "before." I verified the
logic via the unit tests and the mechanism above, but I could not run it on a
physical iPhone. A maintainer spot-check on real iOS Safari / iPadOS would be
appreciated; I'm happy to add an on-device before/after recording if someone can
capture one.

### Reproduction (on an iPhone, before this change)

1. Open a project's **Shell** tab on iOS Safari and let some output scroll in.
2. Try to drag the terminal up/down with one finger — it selects text or does
   nothing instead of scrolling.
3. With this change, the same drag scrolls the buffer natively with momentum,
   and long-press selection / pinch-zoom still work.

## Summary by CodeRabbit

- **New Features**
  - Added accelerated terminal rendering with WebGL and automatic canvas fallback.
  - Improved mobile terminal touch interactions for scrolling, gestures, and text selection.
  - Updated mobile styling to reduce unwanted iOS text selection.
- **Bug Fixes**
  - Prevented rendering issues after WebGL GPU context loss.
- **Tests**
  - Added coverage for mobile touch-action and movement-threshold behavior.


mayankdebnath:
Good catch on the `touch-action` latch — fixed in a6525b2.

Since the `touch-action` value that applies to a gesture is decided when the touch starts, flipping it to `none` inside `startSelection` can't stop the same finger's native pan, so the one-finger extend-drag path was unreliable by construction. I've dropped it: selections are now adjusted via the drag handles only (their elements have `touch-action: none` from the start, so the latch doesn't affect them), which also matches the native iOS selection pattern. Inline comments document the latch behavior so this doesn't regress.

mayankdebnath:
Since this PR leans on a few subtle browser/xterm mechanisms, here are references for each claim in the description (I verified every quote against its source):

**Why the drag never scrolled on iOS (the `user-select` hijack)**
- WebKit keys its iOS touch-gesture interpretation (text selection, the selection loupe) off the element's `user-select` value — see [WebKit bug 231161](https://bugs.webkit.org/show_bug.cgi?id=231161), where a WebKit engineer confirms `-webkit-user-select: none` "does prevent text selection" from touch and the fix made it suppress the selection-gesture UI as well. Our old CSS forced the opposite (`user-select: text !important`) onto the viewport.
- xterm.js has this exact symptom on record: [xtermjs/xterm.js#3613 "[iOS] Scroll problems if you starts on text"](https://github.com/xtermjs/xterm.js/issues/3613) — *"if I try to start my scroll with my finger on some text, it ignores the scroll. If I start it where there is no text, scroll works fine."* Reporter notes it reproduces **only on iOS**, which matches what we saw here.

**Why `touchmove` must be passive (the jank half, per #954)**
- Chrome's docs: *"Browsers can't know if an event listener will prevent scrolling, so they always wait for the listener to finish executing before scrolling the page"* — and passive listeners remove that wait ([Chrome Developers: passive event listeners](https://developer.chrome.com/docs/lighthouse/best-practices/uses-passive-event-listeners)).

**Why gesture arbitration uses `touch-action` set *before* the gesture (and why the handles-only follow-up in a6525b2 is correct)**
- MDN on `touch-action`: *"When a gesture is started, the browser intersects the touch-action values of the touched element and its ancestors"* and, explicitly: *"After a gesture starts, changes to touch-action will not have any impact on the behavior of the current gesture."* ([MDN: touch-action](https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action)) — this is the latch CodeRabbit flagged, and why same-finger extend-drag was unreliable by construction.

**Why `overscroll-behavior: contain` stops the page rubber-band**
- MDN: with `contain`, *"no scroll chaining occurs on neighboring scrolling areas; the underlying elements will not scroll"* ([MDN: overscroll-behavior](https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior)). Supported in Safari since 16.0 per the [WebKit blog](https://webkit.org/blog/13399/webkit-features-in-safari-16-1/); on older Safari it's simply ignored, which degrades to today's behavior.
- I also dropped `-webkit-overflow-scrolling: touch` from the new rule (655ff58): momentum scrolling is default since Safari 13 — Apple's [release notes](https://developer.apple.com/documentation/safari-release-notes/safari-13-release-notes): *"Added support for one-finger accelerated scrolling to all frames and overflow:scroll elements eliminating the need to set -webkit-overflow-scrolling: touch"* — and MDN has removed the page for this non-standard property.

**Why the renderer change is part of this fix**
- The WebGL addon has to be loaded **after** `Terminal.open()` — xterm.js's own source threw *"Cannot activate WebglAddon before Terminal.open"* ([WebglAddon.ts](https://github.com/xtermjs/xterm.js/blob/504a8eb4720517fd1e901b3d07f25848ee200442/addons/xterm-addon-webgl/src/WebglAddon.ts)). Our previous code loaded it before `open()`, so every session silently fell back to the DOM renderer.
- `@xterm/addon-canvas` is the intended fallback for 5.x: *"The purpose of this addon is to be used as a fallback for the webgl addon when better performance is desired over the default DOM renderer, but WebGL2 isn't supported"* ([addon-canvas README @ 5.5.0](https://github.com/xtermjs/xterm.js/tree/5.5.0/addons/addon-canvas)). One honest caveat for reviewers: the canvas renderer is **removed in xterm.js 6.0** ([release notes](https://github.com/xtermjs/xterm.js/releases/tag/6.0.0)), so this fallback is right for the 5.5.x we're on today, and when the repo upgrades to 6.x this simplifies to WebGL + DOM (where the now-working WebGL load order matters even more).
- The context-loss handling follows the [addon-webgl README](https://github.com/xtermjs/xterm.js/tree/master/addons/addon-webgl): *"The browser may drop WebGL contexts for various reasons like OOM or after the system has been suspended"* — on loss we dispose and fall back to canvas rather than the DOM renderer.

mayankdebnath:
Two additions after going through the repo's mobile history more carefully:

**Android prior art.** The Android version of this complaint was #908 (Android Chrome PWA: terminal not touch-scrollable, swipes triggering pull-to-refresh), closed when #923 landed the selection manager. Worth noting that this PR's CSS is essentially what #908's reporter recommended (`touch-action: pan-y` + `overscroll-behavior: contain` on the viewport, native scrolling instead of JS) — and the page-level half of that recommendation (`overscroll-behavior-y: contain` on `html, body`) is already on main. So while the "can't scroll at all" symptom was iOS-specific (`user-select` gesture hijack — WebKit only), the mechanism here is the platform-neutral fix both reports point at. For transparency: repo-wide, open mobile issues currently skew heavily iOS (7 open iOS-tagged vs 0 open Android UI issues), so iOS-first prioritization matches the actual report volume.

**One intentional behavior change I should call out explicitly:** on main, the WebGL addon silently failed (loaded before `open()`), so mobile ran the DOM renderer, and the forced `user-select: text` made *native* browser long-press selection + the OS copy callout reachable. With the renderer actually working (WebGL/canvas draw to a canvas — no DOM text exists) plus the scoped `user-select`, that native path is gone by design. The in-app long-press → handles → Copy menu from #923 fully replaces it — it selects via xterm's buffer (cleaner output than per-row DOM copies) and its Copy works on non-HTTPS deployments via the `execCommand` fallback. The `.shell-mobile-selecting` `user-select` rule effectively only matters in the double-fallback-to-DOM-renderer case.

blackmammoth:
@mayankdebnath, please resolve the merge conflicts and address the comments

PR #987, opened 2026-07-10
feat(shell): honor Claude skip-permissions setting in the web shell (RFC)

mayankdebnath:
> **Draft / RFC.** This is a security-sensitive change and, per CONTRIBUTING
> ("Discuss first for new features"), I'm opening it as a draft to get direction
> before polishing. Happy to adjust or drop it if it's not something you want
> upstream.

## Motivation

The app already has a Claude **Skip Permissions** setting (Settings → Agents →
Permissions → Claude) that bypasses Claude Code's approval prompts in chat via
the SDK's `bypassPermissions` mode. The **web shell ignores it** — it always
launches `claude` normally — so the same project can auto-approve in chat but
still prompt in the shell, and the only way to change that today is a local patch
that hardcodes `--dangerously-skip-permissions`.

This makes the web shell **honor the existing Claude Skip Permissions setting**,
so one toggle governs both chat and shell, and adds a server-wide env default for
headless deployments. Prior art for permission-related config: #802 (Codex
permission env fallback).

## Design

Default is unchanged: with nothing enabled, the shell launches `claude` exactly
as before and prompts stay on.

- **UI (reuses the existing setting):** the shell sends the persisted
  `claude-settings.skipPermissions` value on its `init` message. Enabling
  **Settings → Agents → Permissions → Claude → Skip Permissions** now also makes
  the web shell launch `claude --dangerously-skip-permissions`. No new toggle.
- **Server-wide (env):** `SHELL_DANGEROUSLY_SKIP_PERMISSIONS` (default off;
  accepts `1`/`true`/`yes`/`on`), for headless / no-UI deployments.

**Precedence:** an explicit per-session value overrides the server default; when
neither is set, prompts stay on. The flag is scoped to the `claude` provider and
applied to both the primary invocation and the resume-fallback, on posix and
Windows. Codex / Cursor / opencode / plain shell are never affected (each already
has its own permission mechanism).

## Security notes

- Default off — no behavior change unless explicitly enabled.
- Non-boolean / garbage values fall back to "off".
- Reuses the setting that already carries an orange "dangerous" warning in the
  UI, rather than introducing a second, easier-to-miss control.
- Open question for maintainers: is folding the shell into the existing Claude
  Skip Permissions toggle the behavior you want, or should the shell bypass be
  independently controllable? Should the env var *gate* whether the UI setting
  applies to the shell at all? Guidance welcome.

## Changes

- Server: `shouldSkipClaudePermissions()` + `SHELL_DANGEROUSLY_SKIP_PERMISSIONS`
  wired into `buildShellCommand` (`shell-websocket.service.ts`), refactored to
  take injectable `env`/`platform` for testability.
- Client: the shell reads `claude-settings.skipPermissions` and sends it on the
  shell `init` message.
- Docs: README "Security & Tools Configuration" section and `.env.example`.

## Tests / verification

- New unit tests (`server/modules/websocket/tests/shell-websocket-command.test.ts`)
  asserting the **default-off** guarantee, env/per-session precedence, and
  behavior across posix/win32 and all providers.
- `npm run typecheck`, `npm run lint` (0 errors), full `node:test` suite, and
  `npm run build` — all pass. Server boots and the default path spawns plain
  `claude`.

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in “Skip Permissions” setting for Claude sessions.
  * Added support for configuring the setting server-wide through an environment variable.
  * Per-session settings override the server default, while other providers remain unaffected.

* **Documentation**
  * Documented configuration, security considerations, accepted values, defaults, and UI controls.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -28,6 +28,14 @@ HOST=0.0.0.0
 # Uncomment the following line if you have a custom claude cli path other than the default "claude"
 # CLAUDE_CLI_PATH=claude
 
+# Web shell: skip Claude Code permission prompts (SECURITY DOWNGRADE, default off)
+# When enabled, the web shell launches `claude` with --dangerously-skip-permissions,
+# so Claude Code runs tools without asking for approval. Leave this off unless you
+# fully trust the projects served by this instance. Can also be toggled from the UI
+# via Settings > Agents > Permissions > Claude > Skip Permissions (which overrides
+# this default). Accepted truthy values: 1, true, yes, on
+# SHELL_DANGEROUSLY_SKIP_PERMISSIONS=false
+
 # =============================================================================
 # DATABASE CONFIGURATION
 # =============================================================================
--- a/README.md
+++ b/README.md
@@ -163,6 +163,28 @@ To use Claude Code's full functionality, you'll need to manually enable tools:
 
 **Recommended approach**: Start with basic tools enabled and add more as needed. You can always adjust these settings later.
 
+### Web shell: skipping Claude Code permission prompts (opt-in)
+
+By default the web shell launches `claude` normally, so Claude Code still asks
+for approval before running tools. If you fully trust a project and want the web
+shell to skip those prompts, you can opt into launching Claude Code with
+`--dangerously-skip-permissions`. **This is a deliberate security downgrade and is
+off by default** — enable it only for projects you fully control.
+
+There are two ways to turn it on:
+
+- **From the UI** — enable **Skip Permissions** under **Settings → Agents →
+  Permissions → Claude**. This is the same toggle that controls permission
+  bypass in chat, and it now also applies to the web shell. Off by default.
+- **Server-wide (env)** — set `SHELL_DANGEROUSLY_SKIP_PERMISSIONS=true` for the
+  server. This is useful for headless / no-UI deployments. Accepted truthy
+  values: `1`, `true`, `yes`, `on`.
+
+Precedence: an explicit per-session choice from the UI overrides the server
+default. When neither is set, permission prompts stay on. The flag only applies
+to the `claude` provider; other providers (Codex, Cursor, opencode, plain shell)
+are never affected.
+
 ---
 
 ## Plugins
--- a/server/modules/websocket/services/shell-websocket.service.ts
+++ b/server/modules/websocket/services/shell-websocket.service.ts
@@ -19,6 +19,7 @@ type ShellIncomingMessage = {
   initialCommand?: string;
   isPlainShell?: boolean;
   forceRestart?: boolean;
+  dangerouslySkipPermissions?: boolean;
 };
 
 type PtySessionEntry = {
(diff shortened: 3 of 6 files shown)


mayankdebnath:
Adding references that back the security framing of this RFC (each quote verified against its source):

**What the flag actually is (and why one toggle should govern chat + shell)**
- Claude Code's [CLI reference](https://code.claude.com/docs/en/cli-reference) documents `--dangerously-skip-permissions` as: *"Skip permission prompts. Equivalent to `--permission-mode bypassPermissions`."* That's exactly the mode the chat path already sets when the existing Claude "Skip Permissions" setting is on (`server/claude-sdk.js` maps `settings.skipPermissions` → `permissionMode: 'bypassPermissions'`), and the [Agent SDK docs](https://code.claude.com/docs/en/agent-sdk/permissions) describe that mode as *"Auto-approves all tool uses without prompts."* So this PR makes the web shell honor the same documented equivalence the app already uses — one setting, one meaning, both surfaces.

**Why it must stay default-off**
- The official [permission-modes docs](https://code.claude.com/docs/en/permission-modes) warn: *"Only use this mode in isolated environments like containers, VMs, or dev containers without internet access, where Claude Code cannot damage your host system"* — and add that *bypassPermissions offers no protection against prompt injection or unintended actions*. A hosted web UI is exactly where that warning bites, hence opt-in only, with the env spelling kept deliberately scary (`SHELL_DANGEROUSLY_SKIP_PERMISSIONS`).
- Worth noting for reviewers: per the [security docs](https://code.claude.com/docs/en/security), the first-run *"trust verification"* dialog is one of Claude Code's prompt-injection safeguards — and bypass mode skips prompts wholesale, which is another reason this shouldn't be a silent default.

**Prior art for default-off dangerous capabilities in comparable tools**
- VS Code Workspace Trust: *"preventing automatic code execution of any code in your workspace if the workspace is open in 'Restricted Mode'"* — unfamiliar folders are restricted until the user explicitly trusts them ([VS Code docs](https://code.visualstudio.com/docs/editing/workspaces/workspace-trust)).
- code-server (closest analogue to this project — VS Code served over the web): password auth is on by default and the docs lead with *"**Never** expose code-server directly to the internet without some form of authentication and encryption"* ([code-server guide](https://coder.com/docs/code-server/guide)); `auth: none` exists only as an explicit opt-out. Same shape as this RFC: the dangerous mode exists, but you have to ask for it.

Also related in-repo precedent: #802 (Codex permission env fallback) takes the same env-var approach for another provider's permission mode.

mayankdebnath:
@coderabbitai review

mayankdebnath:
One more open design question for this RFC, after auditing provider parity: **the chat/shell inconsistency this PR fixes for Claude also exists for the other three providers.** The shell currently launches `cursor-agent`, `codex`, and `opencode` with no permission flags at all, while the chat path already honors each provider's own permission setting. If maintainers want, the same pattern extends naturally using each provider's existing setting and documented flag:

| Provider | Setting that already exists | Interactive-shell flag |
|---|---|---|
| Claude | `claude-settings.skipPermissions` | `--dangerously-skip-permissions` (this PR) |
| Cursor | `cursor-tools-settings.skipPermissions` (chat maps it to `-f`) | `-f` / `--force` — note `--trust` is headless-only workspace trust, not approval bypass, so it shouldn't be used here |
| Codex | `codex-settings.permissionMode` | `--dangerously-bypass-approvals-and-sandbox` for bypassPermissions; `-s workspace-write -a never`-style granular flags for acceptEdits (top-level options forward to the TUI, and `codex resume` accepts them) |
| opencode | (chat uses `--auto` for bypass in run mode) | `--auto` is documented on the TUI command as well |

If that direction is preferred, the wire format should be per-provider (send each provider's own setting rather than one Claude-shaped boolean), and flags applied to both the primary and resume-fallback invocations like this PR does. Happy to extend the implementation that way, or keep this PR Claude-only as the first step — maintainer preference. (Any new flags should get the same graceful-fallback treatment as the resume path, since an unknown flag aborts the PTY launch on older CLIs.)

maxwell-01:
Came to check if someone had already done this as I was going to raise a PR if not. +1 from me
fyi you have conflicts to resolve 👍

PR #995, opened 2026-07-11
fix: install browser runtime outside process cwd

CoderLuii:
## Problem

A global CloudCLI 1.36.1 install can say the Browser Use runtime installed successfully, then still report Playwright and Chromium as unavailable.

The installer runs npm from `process.cwd()`, while CloudCLI resolves Playwright relative to the Browser Use service. When CloudCLI starts from another directory, npm installs the package in one place and CloudCLI looks for it somewhere else.

## What changed

- Install Playwright in `~/.cloudcli/browser-use/runtime` instead of the launch directory.
- Pin the Browser Use runtime to Playwright 1.61.1.
- Resolve the normal packaged dependency first, then check the Browser Use runtime directory.
- Run the package and browser installation commands from that same runtime directory.
- Keep the Windows `npm.cmd` handling needed on current Node releases.
- Return a setup error when the runtime directory cannot be created, and allow retry after the path is fixed.
- Add tests for installation, resolution, browser launch, localhost navigation, snapshots, and cleanup.

## Prior work

@benjaminberes-bp identified the isolated runtime directory and Windows `npm.cmd` fix in #917. That PR now conflicts with `main` and also includes separate Claude SDK and launcher changes. This PR brings only the Browser Use runtime fix onto current `main`, with focused tests and a live global-install check. The other Windows changes in #917 remain separate.

I hit this while testing CoderLuii/HolyClaude#66. That issue's arm64 Chromium crash was fixed in HolyClaude's image packaging and is separate from this CloudCLI installer problem.

## Checks

Updated against current `main` (CloudCLI 1.37.3).

- `npm ci`
- Browser Use tests: 3 passed, including directory setup failure and retry.
- `npm run typecheck`
- `npm run lint` (warnings remain; no errors)
- `npm run build`
- Windows 11 / Node 24: installed Playwright and Chromium from an unrelated working directory, then launched Chromium, opened a localhost page, captured a snapshot, and stopped and deleted the session. The launch directory stayed free of runtime dependencies.

The earlier Linux container check has not been rerun for this refresh.

## Summary by CodeRabbit

* **New Features**
  * Browser automation runtime installation now supports a configurable installation directory.
  * Playwright can be loaded from the configured runtime directory when unavailable from the default location.
  * Runtime setup installs a pinned Playwright version and Chromium in the selected directory.
  * Browser automation supports improved Windows command execution and working-directory handling.

* **Bug Fixes**
  * Runtime readiness checks now account for the installation directory, preventing stale results from being reused incorrectly.

### DIFF
--- a/server/modules/browser-use/browser-use.service.ts
+++ b/server/modules/browser-use/browser-use.service.ts
@@ -74,13 +74,20 @@ const sessions = new Map<string, BrowserUseSession>();
 const handles = new Map<string, RuntimeHandle>();
 let installPromise: Promise<{ success: boolean; message: string }> | null = null;
 let lastInstallMessage: string | null = null;
-let runtimeProbeCache: { value: RuntimeProbe; updatedAt: number } | null = null;
+let runtimeProbeCache: { value: RuntimeProbe; updatedAt: number; runtimeInstallDir: string } | null = null;
 
 const DEFAULT_SETTINGS: BrowserUseSettings = {
   enabled: false,
 };
 const AGENT_OWNER_ID = 'agent';
 const PROFILE_ROOT = path.join(os.homedir(), '.cloudcli', 'browser-use', 'profiles');
+const RUNTIME_ROOT = path.join(os.homedir(), '.cloudcli', 'browser-use', 'runtime');
+const PLAYWRIGHT_RUNTIME_VERSION = '1.61.1';
+const RUNTIME_PACKAGE_JSON = JSON.stringify({
+  private: true,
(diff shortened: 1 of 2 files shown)


CoderLuii:
@blackmammoth I've brought this up to current main and fixed the directory setup error handling from the earlier review. The regression test now covers a failed setup followed by a successful retry.

The three Browser Use tests, typecheck, lint and build pass. I also checked the real Windows install from a different working directory, Chromium launch, localhost navigation, snapshot and cleanup. Ready for another look when you have a chance.

PR #996, opened 2026-07-11
fix: resolve per-session model overrides using app session ID

Kennems:
## Problem

Selecting a model from the Chat UI (`/model`) stores the override in `session-model-changes.json` keyed by the **app session ID**, but two things prevented it from working:

1. **Display was stale** — `resolveCommandModel` (used by `/model`, `/cost`, `/status`) called `getCurrentActiveModel` which reads from the JSONL file, never checking the stored overrides. The UI always showed the old model after switching.

2. **Runtime lookup key mismatch** — All provider runtimes (cursor, codex, opencode) called `resolveResumeModel` without `appSessionId`, so the lookup key (`provider:providerNativeSessionId`) never matched the stored key (`provider:appSessionId`).

3. **Env vars could override the model** — `ANTHROPIC_MODEL` and friends in `process.env` were forwarded to the SDK subprocess, potentially overriding the explicit `model` option.

## Changes

- **`commands.js`** — `resolveCommandModel` checks `getChangedActiveModel` (session-model-changes.json) before falling back to `getCurrentActiveModel` (JSONL/provider default). Fixes `/model`, `/cost`, `/status` showing stale model.

- **`claude-sdk.js`, `cursor-cli.js`, `openai-codex.js`, `opencode-cli.js`** — Pass `appSessionId` to `resolveResumeModel` so the stored override is found.

- **`claude-sdk.js`** — Strip `ANTHROPIC_MODEL`, `ANTHROPIC_DEFAULT_OPUS_MODEL`, `ANTHROPIC_DEFAULT_SONNET_MODEL`, `ANTHROPIC_DEFAULT_HAIKU_MODEL` from env forwarded to the SDK subprocess so the explicit model argument wins.

- **Test** — Existing test `resolveResumeModel uses appSessionId for model override lookup` covers the core resolution logic.

Fixes #981

## Summary by CodeRabbit

* **New Features**
  * Improved session-specific model override handling across resumed conversations by using the app-level session id alongside provider session ids.
  * CLI/Provider model selection now preserves an explicit model choice instead of being overridden by Claude-related environment defaults.

* **Bug Fixes**
  * Command model resolution now honors valid, user-selected per-session `active-model` overrides before falling back to the current active model or catalog default.

* **Tests**
  * Added a unit test to verify app-level session overrides are correctly used during resume-model resolution.

PR #998, opened 2026-07-11
fix model select and display issue

### LINKED ISSUE #981: Can not change model from Chat UI
**Describe the bug**
Can not change model from Chat UI:
Image

it said APPLIES NEXT RESPONSE. but actually model won't change at all.

**To Reproduce**
1. prepare an account which opus is not available. and sonnet is available.
2. start a new conversation, select sonnet as the initial model from the top of Chat UI session:
3. ask AI anything and got some response
4. change to opus by Chat UI, 
5. ask AI anything
6. I should get API Error: "400 The requested model is not available for this subscription", but I didn't
7. Additionally, open model change popup again, it shows model wasn't changed.

**Expected behavior**
Could change model from Chat UI

**Screenshots**

**Error message**
If applicable, add the error message you see to help explain your problem.

 - OS: MacOS
 - Browser: Chrome

Sunjaieks:
## Problem
Picking a Claude model from the chat `/models` picker had two independent bugs:
1. **Display** — the picker showed the wrong active model (e.g. `default`/`sonnet`
   while the session was actually running `opus`) and never reflected a just-made pick.
2. **Runtime** — the per-session override (stored by `/active-model`, keyed by the
   *app* session id) was never applied on resume, because runtimes looked it up with
   the *provider-native* session id.
Plus: a host `ANTHROPIC_MODEL` env var could silently override the explicit choice,
and the override cache accumulated orphaned entries.

## Root causes & fixes
Fixes https://github.com/siteboon/claudecodeui/issues/981

### Display — the picker reflects the real model
- `getCurrentActiveModel` now matches transcript events by `provider_session_id`
  (jsonl records the provider id, the frontend passes the app id → every event was
  rejected → fell back to `default`). `claude-models.provider.ts`
- Transcripts record the raw model id (`claude-opus-4-8`); `resolveClaudeActiveOptionValue`
  maps it back to the catalog alias (`opus`) for highlighting. Families are derived
  from the catalog, so a newly released model needs no code change here. `claude-models.provider.ts`
- System-placeholder turns (`"model":"<synthetic>"`) no longer mask the real model. `claude-models.provider.ts`
- A pending override is surfaced first, so reopening the picker shows the choice. `provider-models.service.ts`
- `resolveCommandModel` validates the value against the catalog and falls back to the
  composer's requested model; wired into `/models`, `/cost`, `/status`. `commands.js`

### Runtime — the model change takes effect
- Pass the app session id to the provider runtimes and resolve the override with it.
  `chat-websocket.service.ts`, `claude-sdk.js`, `cursor-cli.js`, `opencode-cli.js`, `openai-codex.js`

### Env precedence
- Strip `ANTHROPIC_MODEL` / `ANTHROPIC_DEFAULT_{OPUS,SONNET,HAIKU}_MODEL` from the SDK
  subprocess env so the explicit selection wins (notably for `default`). `claude-sdk.js`

### Override-cache hygiene
- New `deleteProviderSessionActiveModelChanges(sessionIds)`, called on permanent
  session delete and permanent project delete (archive is reversible → left untouched).
  `utils.ts`, `sessions.service.ts`, `project-delete.service.ts`

### Frontend
- After picking a model for a session, sync the composer model so new sessions inherit
  the latest choice. `useChatProviderState.ts`

## Tests
- Override surfaced before the adapter; raw-id → catalog fallback in the model command;
  override cleanup removes only the targeted sessions across providers.
- Green: server `tsc`, 16 provider-model + command tests, eslint.

## Known limitations (display-only, by design)
- `default` and `sonnet` share the same underlying model, so without an override a
  default session shows `sonnet` in the picker — self-corrects once the user picks. 
- Overrides are intentionally persistent (`changed:true`) — that's the per-session
  model memory; they are cleared only when the session/project is permanently deleted.

## Summary by CodeRabbit

* **New Features**
  * Session-specific model selections now persist and take priority when resuming conversations.
  * Model selections made in the composer are applied consistently to model, cost, and status commands.
  * Provider-native model identifiers are translated into recognizable model-picker options.

* **Bug Fixes**
  * Prevented environment settings from unexpectedly overriding the selected model.
  * Improved model detection when resuming sessions.
  * Removed saved model overrides when sessions or projects are permanently deleted.

Another PR: https://github.com/siteboon/claudecodeui/pull/996 is for same purpose.
⚠️⚠️ This PR includes all the changes from PR #996, along with some additional fixes. Please review and merge only one of the two PRs.

## Summary by CodeRabbit

* **New Features**
  * Improved model selection consistency across chat sessions, resumed conversations, and built-in commands.
  * The Models, Cost, and Status commands now reflect the selected session or composer model when applicable.
  * Selected provider models are saved for future use.

* **Bug Fixes**
  * Prevented placeholder or internal model values from replacing the active model.
  * Improved recognition of provider model names in the model picker.
  * Cleared session-specific model overrides when sessions or projects are permanently deleted.

### DIFF
--- a/server/claude-sdk.js
+++ b/server/claude-sdk.js
@@ -174,6 +174,7 @@ function mapCliOptionsToSDK(options = {}) {
   delete sdkEnv.ANTHROPIC_DEFAULT_OPUS_MODEL;
   delete sdkEnv.ANTHROPIC_DEFAULT_SONNET_MODEL;
   delete sdkEnv.ANTHROPIC_DEFAULT_HAIKU_MODEL;
+  delete sdkEnv.CLAUDE_CODE_SUBAGENT_MODEL;
   sdkOptions.env = sdkEnv;
 
   // Resolve the executable eagerly on Windows because the SDK uses raw child_process.spawn,
@@ -483,7 +484,7 @@ async function queryClaudeSDK(command, options = {}, ws) {
   try {
     const resolvedModel = await providerModelsService.resolveResumeModel(
       'claude',
-      appSessionId ?? sessionId,
+      appSessionId?.trim() || sessionId,
       options.model,
     );
     let effortModels = CLAUDE_FALLBACK_MODELS;
--- a/server/claude-sdk.js
+++ b/server/claude-sdk.js
@@ -164,7 +164,17 @@ function mapCliOptionsToSDK(options = {}) {
 
   // Forward all host env vars (e.g. ANTHROPIC_BASE_URL) to the subprocess.
   // Since SDK 0.2.113, options.env replaces process.env instead of overlaying it.
-  sdkOptions.env = { ...process.env };
+  //
+  // Strip the model-selecting env vars first: the app owns model selection and
+  // passes it explicitly via `sdkOptions.model`. If the host has `ANTHROPIC_MODEL`
+  // (or the per-tier defaults) set, the subprocess would otherwise honor them —
+  // most visibly when the user picks `default`, silently overriding the choice.
+  const sdkEnv = { ...process.env };
+  delete sdkEnv.ANTHROPIC_MODEL;
+  delete sdkEnv.ANTHROPIC_DEFAULT_OPUS_MODEL;
+  delete sdkEnv.ANTHROPIC_DEFAULT_SONNET_MODEL;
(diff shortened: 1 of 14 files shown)


blackmammoth:

PR #1005, opened 2026-07-12
fix(shell): key PTY sessions by client identity and pre-assign Claude session ids

### LINKED ISSUE #1004: Shell: PTY keying collapses all new-session shells to one key — cross-tab hijack, duplicate resumed processes, and orphaned Claude sessions burning tokens
### Environment
- Version: v1.36.1 (`5884573`)
- Applies to all platforms; easiest to hit with two browser tabs, or one mobile + one desktop client on the same project.

### Summary
The `/shell` PTY registry keys sessions as `${projectPath}_${sessionId ?? 'default'}`. Because **every new-session Claude shell in a project collapses to the same `_default` key**, and because **that key changes once the session has a real id**, two distinct failure modes follow:

1. **Cross-tab hijack** — opening a second "new session" shell in the same project silently attaches you to the *first* shell's `claude` process instead of starting a new one. Both terminals now write to the same PTY; only the most recent one renders output. You believe you're in a fresh session while actually typing into another tab's conversation.
2. **Duplicate process + orphaned session (token waste)** — a shell started as "new" lives under `_default`. When the same conversation is later opened *by its session id* (e.g. tapping the newly listed session in the sidebar), the key no longer matches, so the server spawns a **second** `claude --resume <id>` while the original process is still alive under `_default`. Two processes now serve one conversation: the visible one re-reads the transcript, and the invisible one keeps running — streaming any in-flight response into a buffer nobody reads — until the 30-minute reaper (`PTY_SESSION_TIMEOUT`, L34) kills it. Tokens are billed for output no one ever sees.

This is the **session-identity** layer of the shell lifecycle; it is related to but distinct from #953, which covers the **socket** layer (stale `close` detaching a live PTY, half-open sockets never reaped). Fixing #953 alone does not fix this: the keying itself conflates different logical shells and loses track of a session when its id materializes.

### Root cause 1 — all new-session shells share one PTY key, and attach steals silently
`server/modules/websocket/services/shell-websocket.service.ts` L268:

```js
ptySessionKey = `${projectPath}_${sessionId ?? 'default'}${commandSuffix}`;
```

A new Claude shell has no session id yet (`sessionId: null`), so every such shell in a project maps to `${projectPath}_default`. The init handler then finds the existing entry and attaches to it (L281–L307):

```js
const existingSession =
  isLoginCommand || forceRestart ? null : ptySessionsMap.get(ptySessionKey);
if (existingSession) {
  shellProcess = existingSession.pty;
  ...
  ws.send(JSON.stringify({ type: 'output',
    data: '\x1b[36m[Reconnected to existing session]\x1b[0m\r\n' }));
  ...
  existingSession.ws = ws;   // silently detaches whoever was attached
  return;
}
```

There is no check that the previous socket is gone — `existingSession.ws = ws` just reassigns it. The prior client keeps a live reference to the same PTY in its own connection closure, so **its keystrokes still reach the shared process**, but all output now goes to the newest socket. Neither user is told anything beyond a "[Reconnected to existing session]" banner on the *new* side.

Reproduction:
1. Tab A: open a project's Shell, start a new Claude session, ask it something.
2. Tab B: open the same project's Shell intending a *second, separate* session.
3. Tab B shows "[Reconnected to existing session]" and Tab A's conversation; Tab A stops rendering output. Typing in either tab drives the same `claude` process.

As a bonus interaction with #953: when Tab A's socket later closes, the unguarded close handler nulls out `session.ws` and arms a 30-minute kill timer against the PTY that Tab B is actively using.

### Root cause 2 — the key migrates when the session id materializes, forking the process
The registry key embeds the *client-known* session id at init time, but a shell's identity changes over its lifetime:

1. Shell starts as "new" → key `${project}_default`, runs `claude`, which internally creates session `U` (visible in `~/.claude/projects/**/U.jsonl`, and shortly after in the sidebar).
2. The user opens that conversation by id — tapping it in the sidebar, or reconnecting after the Shell tab remounts with the session now selected. The client sends `sessionId: U`.
3. New key `${project}_U` has no entry → the server spawns `claude --resume "U" || claude` (L156–L163) **while the original process for `U` is still running** under `${project}_default`.

Result: two live `claude` processes on one conversation — concurrent writers to the same session — plus the `_default` orphan burning until the reaper fires (only ever armed if its socket closed cleanly; see #953 for the cases where it isn't).

### Hardening note
`ptySessionKey` is a per-connection variable reassigned on every `init`. A second `init` on an already-initialized socket re-points it and leaves the previous entry holding a stale `ws` reference. The current client never sends two inits per socket, but the server shouldn't depend on that.

### Suggested fix
Verified locally against Claude Code v2.1.206:

1. **Stable per-tab shell identity.** The client generates a `shellClientId` (`crypto.randomUUID()`, persisted in `sessionStorage` so it is per-tab but survives remounts) and sends it on `init`. New-session keys become `${projectPath}::new::${shellClientId}` — distinct tabs get distinct PTYs (fixes RC1), while the same tab still reattaches after a tab switch. Clients that don't send the field keep today's behavior.
2. **Pre-assign the Claude session id at spawn.** Launch new sessions as `claude --session-id <server-generated uuid>` and record `sessionId → ptyKey` in an alias map. A later `init` carrying that session id resolves through the alias to the *same* PTY and reattaches instead of spawning `claude --resume` (fixes RC2 — no duplicate process, no orphan). Verified: `claude --session-id <uuid> -p ...` creates exactly `<uuid>.jsonl`, and a subsequent `claude --resume <uuid>` continues the same session file.
3. **Make takeover explicit.** When an `init` attaches to a PTY whose socket is still `OPEN`, notify and detach the old socket deliberately (banner + structured message) instead of silently reassigning `session.ws` — cross-device handoff stays possible, but never invisible. Combined with a `session.ws === ws` guard in the close handler (the RC1 fix from #953), a late stale `close` can no longer detach or doom the new attachment.

I have a working implementation of the above with unit tests and can open a PR right away if this direction looks right.

mayankdebnath:
## Summary

Fixes #1004: the `/shell` PTY registry keyed sessions as
`${projectPath}_${sessionId ?? 'default'}`, which conflated distinct logical
shells and lost track of a conversation once its session id materialized. The
result was silent cross-tab hijack ("you think you're looking at your shell but
you're attached to a different process") and duplicate `claude --resume`
processes with the original left running — burning tokens on output nobody sees
— until the 30-minute reaper.

## The two failure modes (details in #1004)

1. **Cross-tab hijack.** Every new-session shell in a project collapsed to one
   `_default` key, so a second tab silently attached to the first tab's
   process. Both terminals then drove the same PTY while only the newest
   rendered output.
2. **Duplicate process + orphan.** A shell started as "new" lived under
   `_default`; opening the same conversation by its session id later produced a
   different key, so the server spawned a second `claude --resume <id>` while
   the original process kept running under `_default`.

## The fix

- **Per-tab shell identity.** The client sends a stable `shellClientId`
  (persisted in `sessionStorage`, so two tabs are distinct but a remount in the
  same tab reuses it). New-session keys include it — distinct tabs now get
  distinct PTYs; the same tab still reattaches. Session-keyed shells ignore it,
  so cross-device handoff to a resumed session is unchanged. Clients that don't
  send the field keep today's exact behavior.
- **Pre-assigned Claude session ids.** Brand-new Claude sessions launch as
  `claude --session-id <server-generated uuid>` and register a
  `sessionId → ptyKey` alias. A later init that opens the conversation by id
  resolves through the alias and reattaches to the *same* PTY instead of
  forking a duplicate resume.
- **Explicit takeover.** Attaching to a PTY whose socket is still open now
  notifies the old client (yellow banner + structured `session_detached`
  message) before rebinding, instead of stealing it silently — and a detached
  socket can no longer write into or resize the shared PTY.
- **Stale-close guard.** A late `close` from a superseded socket no longer
  detaches the live client or arms a 30-minute kill timer against the PTY a
  newer socket owns (this is the "stale close detaches the live PTY" half of
  #953; the half-open socket reaping from #953 is intentionally out of scope
  here). The reaper timer is also cleared before being rearmed.

Key resolution and alias routing are extracted as pure exported functions
(`resolvePtySessionKey`, `resolveSessionAlias`) with unit tests.

## Verification

- **Live end-to-end against the built server** with concurrent websocket
  clients (register → two `/shell` sockets → init):
  - two tabs with distinct `shellClientId`s → **separate PTYs** (no
    "[Reconnected to existing session]" hijack) ✅
  - two legacy clients without the field → **reattach behavior preserved** ✅
  - the superseded client receives the **detach banner and structured
    `session_detached` message** ✅
  - input from the detached client **no longer reaches the shared PTY** ✅
- **Claude CLI behavior confirmed locally** (Claude Code v2.1.206):
  `claude --session-id <uuid> -p …` creates exactly `<uuid>.jsonl`, and a
  subsequent `claude --resume <uuid>` continues the same session file — so
  pre-assignment + later reattach cannot fork the conversation.
- 14 new unit tests (`node:test`) for key resolution, alias routing, and
  `--session-id` command building across platforms and providers; full suite,
  `npm run lint`, `npm run typecheck`, and `npm run build` all pass.

## Hardening found during self-review

While stress-testing the edge cases before opening this, I found and fixed
three interactions in the initial version of this change:

- **Restart of an alias-routed conversation**: the force-restart path killed
  the aliased PTY but respawned it under the previous tab's identity key,
  which would have re-created the duplicate/orphan problem after a restart.
  The registry key is now recomputed after the kill so the restarted PTY
  lands under its canonical key.
- **Old CLIs**: `--session-id` was added to the Claude CLI in mid-2025; older
  binaries exit with `error: unknown option`. New-session launches now carry
  the same `|| claude` / `$LASTEXITCODE` fallback the resume path already
  uses. (On such CLIs the pre-assigned id never exists, so its alias is
  unreachable and simply ages out.)
- **Superseded client UX**: the client now reacts to `session_detached` by
  closing its socket, so the losing window drops into the normal disconnected
  state (with the reconnect affordance) instead of looking attached while the
  server discards its input.

## Known limitations (unchanged from today's behavior)

- A tab created via "Duplicate tab"/`window.open` clones `sessionStorage`, so
  it shares its source tab's identity and degrades to the current shared-PTY
  behavior — now with an explicit detach notice rather than a silent steal.
- Opening "new session" again in the same tab while a previous new-session
  shell is still alive reattaches to that shell (same as today's `_default`
  behavior); distinct tabs are the case this PR separates.

## Compatibility notes

- Wire protocol: `init` gains an optional `shellClientId` and the server may
  emit `session_detached`; both are ignored by older peers, so mixed versions
  degrade to today's behavior.
- Touches the same file as #987 (independent change, trivial textual conflict —
  I'll rebase whichever lands second).

## Summary by CodeRabbit

* **New Features**
  * Improved shell session continuity across reconnects and browser tabs.
  * Added automatic session handoff when a shell is reopened or reattached.
  * Added support for preserving new Claude sessions with stable session identities.
  * The interface now detects when a session is detached and updates its connection state.

* **Bug Fixes**
  * Prevented input, resizing, and disconnect events from affecting the wrong shell session.
  * Improved cleanup of closed or stale shell sessions.

### DIFF
--- a/server/modules/websocket/services/shell-websocket.service.ts
+++ b/server/modules/websocket/services/shell-websocket.service.ts
@@ -1,3 +1,4 @@
+import { randomUUID } from 'node:crypto';
 import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
@@ -19,6 +20,7 @@ type ShellIncomingMessage = {
   initialCommand?: string;
   isPlainShell?: boolean;
   forceRestart?: boolean;
+  shellClientId?: string;
 };
 
 type PtySessionEntry = {
@@ -28,11 +30,38 @@ type PtySessionEntry = {
   timeoutId: NodeJS.Timeout | null;
   projectPath: string;
   sessionId: string | null;
+  // Session id this PTY's `claude` was launched with (via --session-id), so a
+  // later init that references the session by id can find this same PTY.
+  assignedClaudeSessionId: string | null;
 };
 
 const ptySessionsMap = new Map<string, PtySessionEntry>();
+// Maps a pre-assigned Claude session id to the PTY registry key it lives under.
+// A shell that starts as "new" is keyed by its client identity; once the UI
+// re-opens the same conversation by session id, this alias routes the init back
+// to the original PTY instead of spawning a duplicate `claude --resume`.
+const claudeSessionAliasMap = new Map<string, string>();
 const PTY_SESSION_TIMEOUT = 30 * 60 * 1000;
 const SHELL_URL_PARSE_BUFFER_LIMIT = 32768;
+
+/**
+ * Drops a PTY registry entry along with its pending kill timer and any Claude
+ * session alias that still routes to it. Callers decide whether to kill the pty.
+ */
+function deletePtySessionEntry(key: string, session: PtySessionEntry): void {
+  if (session.timeoutId) {
+    clearTimeout(session.timeoutId);
+  }
+
+  if (
+    session.assignedClaudeSessionId &&
+    claudeSessionAliasMap.get(session.assignedClaudeSessionId) === key
+  ) {
+    claudeSessionAliasMap.delete(session.assignedClaudeSessionId);
+  }
+
+  ptySessionsMap.delete(key);
+}
 
 type ShellWebSocketDependencies = {
   resolveProviderSessionId: (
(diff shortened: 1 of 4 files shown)


mayankdebnath:
Provider-scope note, for reviewers weighing how much of #1004 this covers per provider:

- **The per-tab identity keying is provider-neutral** — new-session shells for Cursor, Codex, and opencode also stop colliding on the shared default key, and the explicit-takeover/stale-close guards apply to all of them.
- **The session-id pre-assignment (the part that kills the duplicate-resume fork) is Claude-only in this PR**, because only Claude's CLI can pin the id at launch (`--session-id <uuid>`).
- **Cursor could gain the same parity later**: `cursor-agent create-chat` returns a new chat id before launch, so the alias could be registered and the shell started with `--resume=<id>` — happy to do that as a follow-up if wanted.
- **Codex and opencode structurally can't pre-assign** (no session-id flag; ids are generated internally and resume requires an existing recorded session), so their new-session shells rely on the client-identity keying alone, and by-id reattach for them still depends on the existing post-hoc session discovery.

blackmammoth:
hey @mayankdebnath, is this still relevant? If so, resolve the merge conflicts, and give a reproduction step for the issue

PR #1016, opened 2026-07-13
fix: chat WebSocket reconnect reuses a stale token after refresh

kuhlsnu:
`authenticatedFetch` (`src/utils/api.js`) persists a refreshed JWT to `localStorage['auth-token']` when the server returns `X-Refreshed-Token`. But `WebSocketContext` builds the chat WS URL from AuthContext's in-memory `token`, which is seeded once at mount and only updated by login/logout, never re-synced from `localStorage`.

As a result, on a long-lived tab the chat WS auto-reconnect keeps reusing the original in-memory token and eventually fails at that token's true expiry, even though REST calls keep silently refreshing. The shell WS (`src/components/shell/utils/socket.ts`) and file upload already read `localStorage` fresh and are unaffected.

This reads the freshest `auth-token` from `localStorage` at (re)connect (falling back to the in-memory token), matching the shell-WS and file-upload paths.

One-line frontend change, no new dependencies. I didn't add a test since the repo has no frontend test harness, but I have a regression test for this from a downstream fork and am happy to contribute it if you add one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket connection reliability by using the most up-to-date authentication token available in the browser when establishing a connection.
  * Preserved existing WebSocket connection, messaging, and error-handling behavior.

### DIFF
--- a/src/contexts/WebSocketContext.tsx
+++ b/src/contexts/WebSocketContext.tsx
@@ -1,5 +1,6 @@
 import { createContext, useCallback, useContext, useEffect, useMemo, useRef, useState } from 'react';
 import { useAuth } from '../components/auth/context/AuthContext';
+import { AUTH_TOKEN_STORAGE_KEY } from '../components/auth/constants';
 import { IS_PLATFORM } from '../constants/config';
 
 /**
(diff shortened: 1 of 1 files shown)


blackmammoth:

PR #1023, opened 2026-07-15
fix: prevent NODE_ENV leakage into spawned child processes

MacLeod92:
### What

I hand roll my deployment using systemd and was plagued with these errors, so I set about to do something about it. This PR is the result. On the face of it, it does look massive, but for the most part it's just one line replacements ensuring something isn't there which ordinarily isn't there anyway in a standard deployment.

The root cause is that any deployment method setting `NODE_ENV=production` on the server's process causes this bug. This includes Heroku's Node.js buildpack, which does so **automatically by default**, with no user opt-in required. It's also common in hand-rolled systemd units and Docker configs, both of which routinely carry `Environment=NODE_ENV=production` / `ENV NODE_ENV=production` as standard boilerplate for Node production deployments generally, independent of this project specifically. The same applies to this project's own documented PM2 deployment path: PM2 does not set `NODE_ENV=production` by default, but it can be triggered the same way — via an `ecosystem.config.js` `env_production` block, or the `--env production` flag used directly in the reproduction below.

Child processes spawned by the server inherit that value by default unless it's explicitly stripped. When it reaches npm/npx invocations, they silently skip devDependencies, breaking build/lint/typecheck downstream—and because the failure surfaces as a confusing, unrelated-looking npm error, it's easy for the actual cause (inherited `NODE_ENV`) to go unnoticed.

The actual problem is broader than the npm/npx symptom: it's full, unfiltered env inheritance—`{ ...process.env }` spreads or a raw `process.env` reference passed as `env`—at any spawn site, not just npm/npx ones. That's why this PR also touches plain git/tar/rg spawns that don't look NODE_ENV-related at a glance: any of them passing unfiltered `process.env` carries `NODE_ENV` (and, on Electron, other host vars) along regardless of whether the visible symptom had shown up yet.

Notably, this app's own runtime code never reads `NODE_ENV` itself (verified: the only reference in `server/` is the one this PR already routes through the shared helper). The bug is entirely about the *server* leaking an ambient, extremely common env var into things it spawns—not about the app needing or depending on that value for its own behavior.

This class of bug is exactly the kind a lint rule could catch at write-time rather than relying on someone reading `server/utils/README.md` before adding a new spawn call. A `no-restricted-syntax` rule targeting a `process.env` spread or direct reference passed as `env` to `spawn`/`exec`/`execFile`/`execSync`/`execFileSync` would be cheap to add (no new dependency, runs in existing lint/build) and would have caught most of the 48 sites here automatically. It wouldn't catch misuse buried inside the three hand-rolled `spawnAsync` wrapper functions (`git.js`, `user.js`, `gitConfig.js`) since those build `env` internally before calling the real `child_process` API, but it'd cover every new direct call going forward, which is where most of these leaks originated. Happy to put this together as a follow-up PR if there's interest, rather than scope-creeping it into this one.

**Related prior PRs:**
- #851 added a Windows-only env allowlist to `plugin-process-manager.js`'s `buildPluginEnv()`, but didn't touch its `NODE_ENV` hardcoding—that's still handled there, intentionally, as an exception (see below).
- #964 was a large refactor touching several of the same files (`git.js`, `user.js`, `gitConfig.js`, `opencode-cli.js`)—it left the `spawn`/env logic untouched in the first three (only the `spawn` import changed, to `cross-spawn`), but it directly touched the leaking line in `opencode-cli.js`, adding a `permissionOptions.env` merge onto the same unfiltered `{ ...process.env }` spread. That PR built new functionality on top of the existing leak rather than missing it.

### Change

Adds `server/utils/childProcessEnv.js` exporting `buildChildProcessEnv()`, which returns `process.env` with `NODE_ENV` stripped (optionally merged with overrides). Routes all identified spawn/exec/execFile call sites through it. See `server/utils/README.md` for usage.

**Exception:** `plugin-process-manager.js`'s `buildPluginEnv()` builds its env from scratch as a curated allowlist, not a `process.env` spread. It drops `NODE_ENV` inline rather than routing through `buildChildProcessEnv()` — piping an allowlist through that helper would widen it into a full passthrough, defeating the point of the allowlist.

### How to Reproduce

1. Set `NODE_ENV=production` on the CloudCLI server process itself (e.g. via
   `pm2 start ... --env production`, a systemd unit's `Environment=`, or
   any deployment method that sets it at the process level).
2. Open a chat session with any agent (Claude, Cursor, OpenCode).
3. Ask the agent to run `echo $NODE_ENV`.

**Before this fix:** the agent's shell reports `production` — the server's
own `NODE_ENV` leaked into the spawned agent process via unfiltered
`process.env` inheritance.

**After this fix:** the agent's shell reports nothing — `NODE_ENV` is
absent, because `buildChildProcessEnv()` strips it before the child
process is spawned.

Tested directly against `server/claude-sdk.js` (the core agent session
spawn) with the server's `NODE_ENV` explicitly set to `production` and
otherwise-identical process environments on both branches.

*(My Claude insists you be informed this is only representative of this one call site... obviously.)*

### Scope: 48 known sites

Grouped by what's actually being spawned, since that's more useful for review than which pass of my investigation found each one. Verification status is per-site regardless of category.

**CLI agent sessions (3)** — the core Claude/Cursor/OpenCode process spawns:

| Site | Verification |
|---|---|
| `server/claude-sdk.js` | live-verified |
| `server/cursor-cli.js` | live-verified |
| `server/opencode-cli.js` | live-verified |

**Provider CLI auth & model checks (8):**

| Site | Verification |
|---|---|
| `cursor-auth.provider.ts:19,77` | live-verified |
| `claude-auth.provider.ts:30` | live-verified |
| `opencode-auth.provider.ts:32` | live-verified |
| `codex-auth.provider.ts:24` | code-reviewed only (codex binary not installed in sandbox) |
| `cursor-models.provider.ts:636` | live-verified, sandbox-direct |
| `opencode-models.provider.ts:389` | live-verified, sandbox-direct |
| `claude-cli-path.ts:93` (`where.exe`) | code-reviewed only (Windows-only path) |

**Git operations (12)** — includes the 3 hand-rolled wrapper definitions, each patched at source (see note below):

| Site | Verification |
|---|---|
| `agent.js:72` (remote), `:230` (log), `:382` (clone), `:1056,1072,1093` (checkout/push) — 6 sites | mix: live-verified where the git binary path was exercised directly, code-reviewed for the identical-pattern sites |
| `project-clone.service.ts:133` | live-verified, sandbox-direct |
| `plugin-loader.js:300` (clone) | live-verified |
| `plugin-loader.js:383` (pull) | code-reviewed only (needs existing checkout) |
| `routes/git.js` `spawnAsync()` wrapper (~45 callers) | live-verified — real caller triggered via running app's Source Control tab; 51 call sites grepped, none override `env` |
| `routes/user.js` `spawnAsync()` wrapper (2 callers) | live-verified — real caller triggered via profile settings save |
| `utils/gitConfig.js` `spawnAsync()` wrapper (2 callers) | live-verified — called real exported `getSystemGitConfig()` |

**Plugin system (4):**

| Site | Verification |
|---|---|
| `plugin-loader.js` npm install/build (3 sites) | live-verified |
| `plugin-process-manager.js` `buildPluginEnv()` — **allowlist exception**, not routed through `buildChildProcessEnv()` (see note below) | code-reviewed — confirmed no `...process.env` spread present |

**Release, installer & Electron tooling (6):**

| Site | Verification |
|---|---|
| `electron/localServer.js:412` | code-reviewed only (Electron desktop process required) |
| `electron/serverInstaller.js:232,248` (tar, 2 sites) | live-verified — real round trip |
| `electron/main.js:487` (osascript, macOS SSH terminal) — found during final sweep, not in original scope | code-reviewed only (macOS/Electron-only path) |
| `build-server-bundle.js:133` (npm ci) | live-verified — full run to completion in isolated scratch dir |
| `build-server-bundle.js:147` (electron-rebuild) | code-reviewed only (live run would leave native bindings ABI-incompatible for rest of session) |

**Misc server utilities (15):**

| Site | Verification |
|---|---|
| `index.js` self-update | live-verified |
| `utils/commandParser.js` | live-verified |
| `browser-use.service.ts` | live-verified |
| `shell-websocket.service.ts` | live-verified |
| `taskmaster.js:518,621,700,753,850` (5 sites) | live-verified |
| `taskmaster.js:47` (which) | live-verified |
| `taskmaster.js:66` (--version) | code-reviewed only (binary not installed) |
| `session-conversations-search.service.ts:586` (ripgrep) | code-reviewed only (`@vscode/ripgrep` binary missing) |
| `cli.js:213` (npm show) | live-verified |
| `cli.js:246` (npm update -g) | code-reviewed only (too disruptive to trigger live) |
| `cli.js:377` (sbx) | code-reviewed only (sbx not installed) |

**A note on "sandbox-direct":** 4 sites above are marked this way — live-verified using the real spawn call shape, but triggered from a standalone script rather than through the running app with process ancestry to `cloudcli.service`. One confidence tier below the app-triggered/wrapper verifications, but exercising the identical one-line fix already proven correct elsewhere.

### On coverage

This is a full audit of every `spawn`/`exec`/`execSync`/`execFile*` call in `server/`, `electron/`, and `scripts/` — not just the sites traced from the original bug reports. But per-site verification varies (live-tested vs. code-reviewed-only, as detailed above), and this isn't guaranteed to be the literal final word on every subprocess call in the repo. If you spot one I missed, happy to follow up.

### Note: not in scope

Three hand-rolled `spawnAsync` wrapper functions exist independently in `git.js`, `user.js`, and `gitConfig.js` — confirmed as organic drift via separate git histories, not copy-paste. Each was patched at its own definition. Consolidating them into one shared wrapper is a reasonable follow-up but out of scope here.

### Testing

- `npm run build` — exit 0 (`build:client` via vite, `build:server` via tsc + tsc-alias)
- `build.mjs` doesn't cover `electron/` or `scripts/`; those four touched files were additionally checked with `node --check` (syntax-valid) and their relative import of `server/utils/childProcessEnv.js` confirmed to resolve at runtime.
- Per-site live/code-review verification detail above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented runtime mode (`NODE_ENV`) from leaking into launched tools and subprocesses.
  * Improved consistency by standardizing how environment variables are constructed for Git, shell, package/CLI, plugins, AI tools, and archive/build/verification steps.
  * Preserved required overrides (such as `PATH` and command-specific settings).
* **Documentation**
  * Added clear guidance on constructing child-process environments safely, with examples of updated spawn sites and known coverage limits.

### DIFF
--- a/electron/localServer.js
+++ b/electron/localServer.js
@@ -6,6 +6,7 @@ import os from 'node:os';
 import path from 'node:path';
 
 import { ServerInstaller } from './serverInstaller.js';
+import { buildChildProcessEnv } from '../server/utils/childProcessEnv.js';
 
 const DEFAULT_PORT = 3001;
 const HOST = '127.0.0.1';
@@ -408,14 +409,13 @@ export class LocalServerController {
     this.ownedServerProcess = spawn(runtime.command, [serverEntry], {
       cwd: serverCwd,
       detached: true,
-      env: {
-        ...process.env,
+      env: buildChildProcessEnv({
         ...runtime.env,
         HOST: bindHost,
         SERVER_PORT: String(port),
         NODE_ENV: 'production',
         PATH: getDesktopPath(),
-      },
+      }),
       stdio: ['ignore', 'pipe', 'pipe'],
       windowsHide: true,
     });
--- a/electron/main.js
+++ b/electron/main.js
@@ -8,6 +8,7 @@ import { DesktopWindowManager } from './desktopWindow.js';
 import { DesktopNotificationsController } from './desktopNotifications.js';
 import { LocalServerController } from './localServer.js';
 import { TabsController } from './tabs.js';
+import { buildChildProcessEnv } from '../server/utils/childProcessEnv.js';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
@@ -487,6 +488,7 @@ async function openEnvironmentInSsh(environment) {
     spawn('osascript', ['-e', `tell application "Terminal" to do script "${escaped}"`], {
       detached: true,
       stdio: 'ignore',
+      env: buildChildProcessEnv(),
     }).unref();
   } else {
     clipboard.writeText(sshCommand);
--- a/electron/serverInstaller.js
+++ b/electron/serverInstaller.js
@@ -6,6 +6,8 @@ import https from 'node:https';
 import os from 'node:os';
 import path from 'node:path';
 
+import { buildChildProcessEnv } from '../server/utils/childProcessEnv.js';
+
 /**
  * Installs the versioned local server runtime used by CloudCLI Desktop.
  *
@@ -232,6 +234,7 @@ export class ServerInstaller {
       const child = spawn('tar', ['-xzf', archivePath, '-C', destDir], {
         stdio: ['ignore', 'ignore', 'pipe'],
         windowsHide: true,
+        env: buildChildProcessEnv(),
       });
       let stderr = '';
       child.stderr?.on('data', (c) => (stderr += c));
@@ -248,6 +251,7 @@ export class ServerInstaller {
       const child = spawn('tar', ['-tzf', archivePath], {
         stdio: ['ignore', 'pipe', 'pipe'],
         windowsHide: true,
+        env: buildChildProcessEnv(),
       });
       let stdout = '';
       let stderr = '';
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -162,6 +162,7 @@ export default tseslint.config(
             "server/shared/frontmatter.ts",
             "server/shared/claude-cli-path.ts",
             "server/shared/image-attachments.ts",
+            "server/utils/childProcessEnv.js",
           ], // classify shared utility files so modules can depend on them explicitly
           mode: "file",
         },
--- a/scripts/release/build-server-bundle.js
+++ b/scripts/release/build-server-bundle.js
@@ -5,6 +5,7 @@ import fs from 'node:fs/promises';
 import path from 'node:path';
 import { spawn } from 'node:child_process';
 import { fileURLToPath } from 'node:url';
+import { buildChildProcessEnv } from '../../server/utils/childProcessEnv.js';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 const rootDir = path.resolve(__dirname, '..', '..');
@@ -129,11 +130,10 @@ await writeServerPackageJson(stageDir);
 console.log('Installing production server dependencies into bundle stage...');
 await run('npm', ['ci', '--omit=dev'], {
   cwd: stageDir,
-  env: {
-    ...process.env,
+  env: buildChildProcessEnv({
     npm_config_audit: 'false',
     npm_config_fund: 'false',
-  },
+  }),
 });
 
 const electronVersion = getElectronVersion();
@@ -143,11 +143,10 @@ const electronRebuild = process.platform === 'win32'
 console.log(`Rebuilding native server dependencies for Electron ${electronVersion} (${arch})...`);
 await run(electronRebuild, ['--version', electronVersion, '--module-dir', stageDir, '--arch', arch, '--force'], {
   cwd: rootDir,
-  env: {
-    ...process.env,
+  env: buildChildProcessEnv({
     npm_config_audit: 'false',
     npm_config_fund: 'false',
-  },
+  }),
 });
 
 if (await pathExists(path.join(stageDir, 'scripts', 'fix-node-pty.js'))) {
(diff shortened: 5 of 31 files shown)


viper151:
Thanks for taking the time to investigate this and for documenting the npm behavior.
After tracing the current launch and deployment paths, we do not think a repository-wide child-process policy is needed here. CloudCLI does not require or document NODE_ENV=production for its build or server runtime, and our installation guides do not tell users to set it. The only first-party path that assigned it was the Electron local-server launcher, so we have addressed that source directly.Custom deployments can still choose to set NODE_ENV, but stripping it from every subprocess—including Git, release tooling, provider checks, and unrelated commands—would introduce a broad behavioral policy without a corresponding CloudCLI requirement.
We are therefore closing this PR in favor of the focused fix. We appreciate the report and the work that went into auditing the subprocess paths.

PR #1033, opened 2026-07-18
fix(claude): detect macOS Keychain authentication

### LINKED ISSUE #556: claude code login issue in OSX
# Bug Report: Claude Auth Status Not Updating After OAuth Login

**Describe the bug**
After completing the Claude login flow via OAuth, the account status remains shown as "Not authenticated" even though authentication succeeded. The app never updates its auth status after the login terminal closes.

**To Reproduce**
1. Go to Settings → Agents → Claude
2. Click "Login"
3. Complete the OAuth flow in the browser (authorize on Anthropic's site)
4. Close the terminal modal
5. Account still shows "Not connected" / "Error: Not authenticated"

**Expected behavior**
After completing the OAuth flow and closing the terminal, the Claude account status should update to show the authenticated user's email and a connected state.

**Error message**
```
Error: Not authenticated
```

**Desktop:**
- OS: macOS
- Browser: Chromium
- Version: (current)
- Claude code: 2.1.77

**Additional context**
Two root causes identified:

1. **macOS only** — The app checks authentication by reading `~/.claude/.credentials.json`. On macOS, the Claude CLI stores credentials in the **Keychain** instead of that file, so the file never exists and the check always fails. Not tested on other systems.

2. **All platforms** — After the login terminal closes, the app does not re-check auth status. The status is only re-checked when the terminal process exits with code 0, but `claude /login --dangerously-skip-permissions` drops into an interactive shell after login (process never exits), so the re-check never fires.

**Proposed fix:**
- Replace `checkClaudeCredentials()` with `claude auth status --json` which works regardless of where credentials are stored
- Change login command from `claude /login --dangerously-skip-permissions` to `claude auth login` (exits cleanly after OAuth completes)
- Re-check auth status when the login modal is closed, not just on process exit

**Workaround for macOS:**
```bash
security find-generic-password -s "Claude Code-credentials" -w | \
  jq '. + {"email": "your@email.com", "user": "Your Name"}' \
  > ~/.claude/.credentials.json
```

sudo-eugene:
Closes #556

## What changed

On macOS, CloudCLI can report Claude as disconnected even while conversations work because Claude Code stores OAuth credentials in Keychain rather than `~/.claude/.credentials.json`. Authentication status now comes from `claude auth status --json`, with the existing environment, settings, and legacy credential-file paths retained for compatibility.

## Why

Claude Code itself is the reliable authority for whether its credentials are usable. Asking the CLI avoids duplicating platform-specific credential-storage logic or attempting to read Keychain directly, and keeps CloudCLI aligned with the authentication source Claude actually uses at runtime.

The status response is parsed even when Claude exits with status 1, because that is its documented logged-out result rather than an unsupported-command failure. Timeouts remain distinguishable from a genuine logged-out response so Keychain-only users are not silently misreported.

This is a deliberately narrow follow-up to #881. That PR also changed the login command and refresh behavior; maintainers requested separate changes and noted that the existing `/login` flow exposes additional authentication choices. Those behaviors remain untouched here.

## Out of scope

The login command, available login methods, and modal refresh behavior are unchanged.

### Before

CleanShot 2026-07-18 at 11 43 02@2x

### After

CleanShot 2026-07-18 at 11 42 43@2x

---

## Test plan

- [x] `npx tsx --tsconfig=server/tsconfig.json --test server/modules/providers/tests/claude-auth.test.ts`
- [x] `npm run typecheck`
- [x] `npm run lint` (passes with the repository's existing warnings)
- [x] `npm run build` (passes with the repository's existing CSS and bundle-size warnings)
- [x] Verified the development API reports `authenticated: true` with method `cli:claude.ai` for a macOS Keychain-backed login
- [x] Verified the Agents account card shows **Connected** in the development UI

## Summary by CodeRabbit

* **New Features**
  * Added CLI-based authentication status detection for Claude (including login state, method, and email when available).
  * CLI results now take precedence, while existing environment and credential-file checks remain in place.
* **Bug Fixes**
  * Improved handling of CLI errors, missing binaries, and non-standard outputs.
  * Correctly differentiates CLI timeouts from logged-out states to avoid incorrect authentication reporting.

### DIFF
--- a/server/modules/providers/list/claude/claude-auth.provider.ts
+++ b/server/modules/providers/list/claude/claude-auth.provider.ts
@@ -29,8 +29,8 @@ export class ClaudeProviderAuth implements IProviderAuth {
   private checkInstalled(): boolean {
     const cliPath = resolveClaudeCodeExecutablePath(process.env.CLAUDE_CLI_PATH);
     try {
-      this.spawnSync(cliPath, ['--version'], { stdio: 'ignore', timeout: 5000 });
-      return true;
+      const result = this.spawnSync(cliPath, ['--version'], { stdio: 'ignore', timeout: 5000 });
+      return !result.error && result.status === 0;
     } catch {
       return false;
     }
--- a/server/modules/providers/list/claude/claude-auth.provider.ts
+++ b/server/modules/providers/list/claude/claude-auth.provider.ts
@@ -21,13 +21,15 @@ const hasErrorCode = (error: unknown, code: string): boolean => (
 );
 
 export class ClaudeProviderAuth implements IProviderAuth {
+  constructor(private readonly spawnSync: typeof spawn.sync = spawn.sync) {}
+
   /**
    * Checks whether the Claude Code CLI is available on this host.
    */
   private checkInstalled(): boolean {
     const cliPath = resolveClaudeCodeExecutablePath(process.env.CLAUDE_CLI_PATH);
     try {
-      spawn.sync(cliPath, ['--version'], { stdio: 'ignore', timeout: 5000 });
+      this.spawnSync(cliPath, ['--version'], { stdio: 'ignore', timeout: 5000 });
       return true;
     } catch {
       return false;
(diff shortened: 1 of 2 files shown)


sudo-eugene:
@coderabbitai review

sudo-eugene:
@coderabbitai review

PR #1038, opened 2026-07-18
feat(appearance): add system font preference

sudo-eugene:
## What changed

CloudCLI now offers a persistent **Use system font** option under Appearance → Typography. Enabling it replaces both Encode Sans in the general interface and Merriweather in chat and reading views with the device's native system font; disabling it restores the existing typography.

## Why

The current font choices are applied globally, leaving users who prefer their platform's native typography without a consistent alternative. A single preference covers both font layers so the interface does not end up with a mixture of system and bundled fonts.

The switch changes shared font variables rather than replacing individual component classes. This keeps the existing typography as the default, applies the preference to current and future views that use the shared font definitions, and avoids maintaining separate font overrides throughout the component tree.

## What this enables

Users can match CloudCLI's typography to their operating system for a more native and personally readable interface, with their choice retained across reloads.

### Before
CleanShot 2026-07-18 at 20 43 14

### After 
CleanShot 2026-07-18 at 20 41 27
CleanShot 2026-07-18 at 20 41 12

---

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint` (passes with the repository's existing warnings)
- [x] `npm run build` (passes with the repository's existing CSS and bundle-size warnings)
- [x] Verified the toggle restores Encode Sans and Merriweather when disabled
- [x] Verified the toggle applies the system font to both interface and reading styles when enabled
- [x] Verified the enabled preference persists after a full page reload

## Summary by CodeRabbit

* **New Features**
  * Added an Appearance → Typography option to toggle system font usage.
  * Persisted the preference and applied it on startup to avoid font flicker.
* **UI/Localization**
  * Added new “system font” labels and description in settings.
* **Styling**
  * Introduced theme CSS font variables with a `system-font` override; Tailwind font families now follow those variables.
* **Reliability**
  * Improved font preference persistence with safer local storage helpers and graceful fallback when storage isn’t available.


blackmammoth:
Hey @sudo-eugene, can you address the coderabbit comments?

sudo-eugene:
@blackmammoth done

PR #1045, opened 2026-07-20
fix(cursor): forward stream-json thinking events to the UI

joostfaassen:
## Summary
Cursor Agent emits `type: thinking` (delta/completed) over `--output-format stream-json`, but CloudCLI's Cursor bridge only handled `system` / `user` / `assistant` / `result`. Thinking events were dropped, so the **Show thinking** toggle stayed empty for Cursor sessions (reproduced with Grok 4.5 High Fast).

This buffers thinking deltas and emits `kind: thinking` on `completed` (with a flush before `assistant` as a safety net), matching the UI path that already renders thinking blocks for Claude.

## Test plan
- [ ] Self-host CloudCLI with Cursor provider
- [ ] Enable **Show thinking**
- [ ] Start a session with a thinking-capable Cursor model (e.g. `cursor-grok-4.5-high-fast`)
- [ ] Confirm a thinking block appears before the assistant reply
- [ ] Confirm Claude sessions still show thinking as before

### DIFF
--- a/server/cursor-cli.js
+++ b/server/cursor-cli.js
@@ -97,6 +97,7 @@ async function spawnCursor(command, options = {}, ws) {
       const isTrustRetry = runReason === 'trust-retry';
       let runSawWorkspaceTrustPrompt = false;
       let stdoutLineBuffer = '';
+      let thinkingBuffer = ''; // accumulate cursor thinking deltas for UI
       let terminalNotificationSent = false;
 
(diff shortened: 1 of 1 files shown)


blackmammoth:
hey @joostfaassen, can u check if this is still necessary and if so, resolve the merge conflicts.

blackmammoth:
Closed as there are no updates.

PR #1037, opened 2026-07-18
feat: numerous bugfixes and features

blackmammoth:
Closes PR #578. Closes PR #962. Closes PR #980. Closes PR #1055. Closes PR #985. Closes PR #1048. Closes PR #1016. Closes PR #857. Closes PR #810. Closes PR #1047. Closes PR #996. Closes PR #998. 

----------
Closes #1062
Closes #981 

This pull request introduces several improvements and bug fixes related to session ID handling, worktree management, and dependency updates. The main focus is on ensuring that session and process management consistently use the correct session identifiers (distinguishing between app-level and provider-native IDs), adding support for Git worktree management, and updating dependencies for enhanced Markdown support.

**Session and Process Management Improvements:**

[...]

PR #1048, opened 2026-07-21
fix(skills): discover skill directories that are symlinks

shokosanma:
## Summary
- Treat directory symlinks as traversable when discovering provider skill `SKILL.md` files.
- Fixes cases like `~/.cursor/skills` where `Dirent.isDirectory()` is false for symlinks even when the target is a directory.

## Test plan
- [ ] Place a skill behind a symlink under a provider skills root (e.g. `~/.cursor/skills/my-skill` → real dir with `SKILL.md`)
- [ ] Confirm the skill appears in CloudCLI skill discovery / settings
- [ ] Confirm normal (non-symlink) skill directories still discover correctly
- [ ] Confirm broken/non-directory symlinks do not crash discovery

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Skill discovery now supports provider skill directories accessed through symbolic links.
  * Recursive and direct searches can locate `SKILL.md` files within symlinked directories.

### DIFF
--- a/server/shared/utils.ts
+++ b/server/shared/utils.ts
@@ -910,7 +910,9 @@ export async function findProviderSkillMarkdownFiles(
     }
 
     for (const entry of entries) {
-      if (entry.isDirectory()) {
+      // Symlinks to skill directories are common in ~/.cursor/skills; Dirent.isDirectory()
+      // is false for links even when the target is a directory.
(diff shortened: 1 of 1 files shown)


blackmammoth:

PR #1055, opened 2026-07-22
fix: render web-UI slash commands compactly instead of leaking prompt

### LINKED ISSUE #1009: fix: skill content rendered as user input in web UI
## Problem

When calling a skill from the web UI, the skill's prompt content is rendered as a **user message** in the chat transcript. In the CLI, skill invocation is handled transparently — the user only sees the skill's output, not its internal prompt leaked into the conversation.

The root cause is in `claude-sessions.provider.ts` `normalizeMessage`: the SDK streams the skill's prompt as a `user` role message, and `normalizeMessage` has no logic to detect or suppress internal skill/task prompts. It falls through to the default user-text path and gets rendered as a bubble.

## Steps to Reproduce

1. Open a session in the web UI
2. Invoke a skill (e.g. type `/skill-name` or use a plugin skill)
3. Observe the skill's internal prompt text appearing as a user message bubble

## Expected Behavior

Skill invocation content should not appear as user input. Only the skill's output (assistant response) should be visible, matching CLI behavior.

## Suggested Fix

In `claude-sessions.provider.ts` `normalizeMessage`, detect skill-related user messages (e.g. tagged with `<local-command-*>` markers, or matching known skill prompt patterns) and either:
- Suppress them entirely (return `[]`), or
- Re-label them as `assistant` with an appropriate visual indicator (similar to how `isCompactSummary` and `isLocalCommand` are already handled)

The `isInternalContent` helper already exists for filtering internal text — extend it to cover skill prompts, or add a dedicated check before the user-text path.

snazzybean:
Fixes #1009

## Problem
Custom slash commands were expanded client-side and their full body was sent as the chat message, so the internal prompt showed up as a user bubble both live and after a reload.

## Reproduction
1. Open the web UI with a Claude project and open or start a session.
2. From the composer, invoke a custom slash command with arguments, e.g. `/handoff just the essentials`.
3. Before this change: the whole expanded command body renders as your user message, both live and after reloading the session, instead of the command you typed.
4. After this change: the bubble shows only the compact `/handoff just the essentials`, live and after reload.

## Change
Display text and the sent prompt are now separate. The bubble only shows `/name args`, while Claude receives the expanded body wrapped in the CLI-native `<command-name>` tags. On reload the existing normalization folds that back into the compact form. The array-content branch parses the same wrapper, since aborted and streamed runs deliver the prompt as array text. Other providers keep receiving the plain body.

## Tests
5 new node:test cases: string and array form, empty args, native tag-only, and a negative case. `npm run typecheck`, `npm run lint`, and `npm run build` are all green.

## Review focus
The delicate part is the fingerprint coupling: frontend and backend must format the command identically, otherwise a duplicate bubble appears after the run completes. Please also check the non-Claude path.

## Summary by CodeRabbit

- **New Features**
  - Custom slash commands now show a compact optimistic command bubble, while the expanded prompt is used for submission.
  - Queued slash-command drafts now preserve and restore the expanded prompt, including when auto-sending.

- **Bug Fixes**
  - Improved Claude “local slash command” handling to better recognize wrapper tags only when they start a message.
  - Prevented accidental local-command collapsing when tags appear mid-sentence.
  - Ensured command turns with image attachments render as a single user bubble.

- **Tests**
  - Added coverage for normalization across wrapper formats, empty arguments, prose immunity, mid-sentence tags, and attachment rendering.

### DIFF
--- a/server/modules/providers/list/claude/claude-sessions.provider.ts
+++ b/server/modules/providers/list/claude/claude-sessions.provider.ts
@@ -246,6 +246,14 @@ type ClaudeLocalCommandPayload = {
  * normal text path continue untouched for unrelated messages.
  */
 function parseLocalCommandPayload(content: string): ClaudeLocalCommandPayload | null {
+  // Only treat the text as a local command when the wrapper opens the payload.
+  // A normal user message that merely mentions a `<command-name>` tag somewhere
+  // in its body must stay untouched — otherwise it would collapse into a compact
+  // bubble and its surrounding prose would be discarded (issue #1009 follow-up).
+  if (!/^\s*<command-(?:name|message|args)>/.test(content)) {
+    return null;
+  }
+
   const commandName = extractTaggedContent(content, 'command-name');
   const commandMessage = extractTaggedContent(content, 'command-message');
   const commandArgs = extractTaggedContent(content, 'command-args');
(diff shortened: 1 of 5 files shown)


blackmammoth:

PR #1056, opened 2026-07-23
fix: skip Claude's <synthetic> placeholder when reading transcript models

MetalZealot:
## Problem

After an API error (e.g. 529 Overloaded), Claude Code appends locally-fabricated assistant rows to the session transcript stamped with the placeholder model `"<synthetic>"` — the API-error notice, `"No response requested."`, and session-limit messages. `extractClaudeEventModel` accepts whatever model the last assistant row carries, so the backward transcript scan in `getCurrentActiveModel` adopts the placeholder and reports `<synthetic>` as the session's active model until a genuine turn lands.

## Repro

1. Have a Claude session hit an API error (529 Overloaded, session limit, etc.) — the CLI writes assistant rows with `"model": "<synthetic>"` to the session JSONL.
2. Switch to / reopen that session → the active model resolves to `<synthetic>` instead of the model the session was actually running.

## Fix

`extractClaudeEventModel` now rejects angle-bracketed placeholder values for both `event.model` and `message.model`, so the scan keeps walking back to the last genuine turn and recovers the session's real model. A transcript containing only synthetic rows falls back to the catalog default, same as before when no model was found.

## Tests

New `server/modules/providers/tests/claude-models.test.ts` (node:test, same isolated-database harness as `opencode-sessions.test.ts`):

- control: the last assistant turn's model is read from the transcript
- synthetic error rows are skipped and the real model is recovered
- an all-synthetic transcript falls back to the catalog default

Run with `npx tsx --tsconfig server/tsconfig.json --test server/modules/providers/tests/claude-models.test.ts` — 3/3 pass with the fix; the two guard tests fail on current `main`. `tsc --noEmit -p server/tsconfig.json` and eslint are clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Claude model detection by ignoring synthetic placeholder model values.
  * Active model selection now correctly retains the last genuine model after errors or synthetic transcript entries.
  * Added/verified fallback behavior when no valid model is available.
* **Tests**
  * Added coverage to validate active model extraction across transcripts containing synthetic placeholder entries.

### DIFF
--- a/server/modules/providers/list/claude/claude-models.provider.ts
+++ b/server/modules/providers/list/claude/claude-models.provider.ts
@@ -139,6 +139,13 @@ const ANSI_PATTERN = new RegExp(
   'g',
 );
 
+// Claude Code stamps locally-fabricated assistant rows (API-error notices,
+// "No response requested.", session-limit messages) with the placeholder
+// model "<synthetic>". Adopting it would feed a non-model into the send path,
+// so placeholder values are skipped and the transcript scan keeps walking back
+// to the last real turn.
+const isPlaceholderClaudeModel = (value: string): boolean => /^<.*>$/.test(value);
+
 const extractClaudeEventModel = (event: ClaudeInitEvent, sessionId: string): string | null => {
   const eventSessionId = event.sessionId ?? event.session_id;
   if (eventSessionId && eventSessionId !== sessionId) {
@@ -151,12 +158,16 @@ const extractClaudeEventModel = (event: ClaudeInitEvent, sessionId: string): str
   }
 
   const directModel = event.model?.trim();
-  if (directModel) {
+  if (directModel && !isPlaceholderClaudeModel(directModel)) {
     return directModel;
   }
 
   const messageModel = event.message?.model?.trim();
-  return messageModel || null;
+  if (messageModel && !isPlaceholderClaudeModel(messageModel)) {
+    return messageModel;
+  }
+
+  return null;
 };
 
 const stripAnsi = (value: string): string => value.replace(ANSI_PATTERN, '');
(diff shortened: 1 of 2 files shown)


blackmammoth:
Hello @MetalZealot, thanks for the PR. I tried replicating the issue but it doesn't seem to occur. Can you check and verify again? If so, can you share a more clear reproduction step?

MetalZealot:
Thanks for checking. I re-verified this against current `main` (`c2408f0`) and the issue is still present.

The trigger is narrower than “any 529”: it occurs only when Claude Code appends a locally generated assistant row whose `message.model` is `"<synthetic>"`. If the failed request does not write one of those rows, the bug will not reproduce.

A deterministic reproduction is:

1. Create/open a Claude session and let one genuine assistant turn land.
2. In that session's JSONL, append a synthetic assistant row after the real turn (replace `<session-id>` with the transcript's session id):

```json
{"type":"assistant","sessionId":"<session-id>","message":{"model":"claude-sonnet-5","content":[]}}
{"type":"assistant","sessionId":"<session-id>","message":{"model":"<synthetic>","content":[{"type":"text","text":"You've hit your session limit"}]}}
```

3. Switch away and reopen the session, then open `/model`.

**Actual on current main:** the active model resolves to `<synthetic>`.

**Expected:** the scan skips that placeholder row and keeps walking back to `claude-sonnet-5`.

The same rows are produced naturally by session-limit notices, some API/auth errors, and `"No response requested."`. As a real-world check, I found 173 synthetic rows across 75 local Claude transcript files, including recent rows from July 30. One sanitized adjacent pair is:

```text
2026-07-17T14:21:54.526Z  claude-fable-5
2026-07-17T14:21:55.603Z  <synthetic>  You've hit your session limit
```

I also merged this PR locally into current upstream `main` and reran:

```sh
npx tsx --tsconfig server/tsconfig.json --test server/modules/providers/tests/claude-models.test.ts
```

All 3 tests pass on the current merge result.

PR #1067, opened 2026-07-29
fix(sessions): reconcile indexed sessions with transcripts on disk

### LINKED ISSUE #813: [BUG] Session DB rows are never removed when underlying .jsonl is deleted (e.g. by Claude Code's cleanupPeriodDays) — ghost sessions accumulate forever
## Summary

CloudCLI maintains its own SQLite `sessions` table at `~/.cloudcli/auth.db`, populated by scanning `~/.claude/projects/**/*.jsonl`. When Claude Code's built-in `cleanupPeriodDays` (default: 30) deletes a session transcript from disk, CloudCLI never removes the corresponding row. The UI lists these as normal sessions; clicking one shows an empty conversation because the backing `.jsonl` is gone.

## Reproduction

1. Use CloudCLI for >30 days without overriding `cleanupPeriodDays`.
2. Claude Code's startup cleanup deletes top-level `~/.claude/projects/<project>/<uuid>.jsonl` files older than 30 days.
3. CloudCLI sidebar continues to show sessions back to install date — clicking one loads an empty session.

On my machine (CloudCLI installed 2026-03-06, Claude Code default cleanup):

| Metric | Count |
|---|---|
| \`sessions\` rows in \`auth.db\` (provider='claude') | 459 |
| Rows whose \`jsonl_path\` still exists | 284 |
| **Rows whose \`jsonl_path\` is missing (ghosts)** | **175** |
| Oldest row | 2026-03-06 (87 days ago) |

## Root cause

Two code paths populate the DB, neither of which removes rows for deleted files.

**1. Initial sync** — \`server/modules/providers/list/claude/claude-session-synchronizer.provider.js\`:

\`\`\`js
async synchronize(since) {
  const files = await findFilesRecursivelyCreatedAfter(
    path.join(this.claudeHome, 'projects'), '.jsonl', since ?? null);
  for (const filePath of files) {
    // ...upsert into sessionsDb
  }
}
\`\`\`

Only upserts discovered files. No reconciliation pass to drop rows whose \`jsonl_path\` no longer resolves.

**2. Live watcher** — \`server/modules/providers/services/sessions-watcher.service.js\`:

\`\`\`js
watcher
  .on('add',    (filePath) => { void onUpdate('add',    filePath, provider); })
  .on('change', (filePath) => { void onUpdate('change', filePath, provider); })
  .on('error',  (error) => { /* log */ });
\`\`\`

No \`unlink\` or \`unlinkDir\` subscription. \`chokidar\` emits these by default — the watcher just doesn't listen. As a result, file deletions from any source (Claude Code cleanup, manual \`rm\`, external tooling) never propagate to the DB.

\`sessionsDb.deleteSessionById()\` exists in \`server/modules/database/repositories/sessions.db.js\` and is already wired for archive/delete UI flows — it just needs to be invoked on filesystem unlink events.

## Suggested fix

Either or both:

1. **Reconcile on initial sync** — after scanning, run \`getAllSessions()\` and \`deleteSessionById()\` for rows whose \`jsonl_path\` does not exist on disk.
2. **Subscribe to \`unlink\` / \`unlinkDir\` in the chokidar watcher** and call \`sessionsDb.deleteSessionById(sessionId)\` (looked up from \`jsonl_path\`).

Option 1 alone would fix ghost rows for users on restart; option 2 keeps the UI live.

## Environment

- CloudCLI: latest \`@cloudcli-ai/cloudcli\` global install
- Claude Code: 2.1.157
- Node v20.20.2
- Ubuntu 24.04 LTS, single-user homelab install

thevinchi:
Fixes #813.

## What

Two defects leave the session index out of sync with what's actually on disk:

1. **Deleted transcripts leave ghost rows forever.** The watcher handles `add`/`change` but not `unlink`, so a removed transcript's row and sidebar entry survive indefinitely. Nothing reconciles at startup either, so deletions during downtime are never noticed.
2. **Sync upserts silently un-archive.** `createSession` hardcodes `isArchived = 0`, and it runs on every full sync rather than only on real activity — so anything archived in the UI is un-archived within seconds, which makes the archive view appear permanently empty.

## How

Guiding rule: **disk is the source of truth for whether a session exists; `isArchived` is only a UI-side hide for sessions that still exist.**

- **`unlink` handler** deletes the row backing a removed transcript and broadcasts a new `session_removed` event so the sidebar drops it live. Skips `opencode` (shared `opencode.db`, so one unlink implies nothing about any session) and `subagents/` transcripts (they repeat the parent's session id).
- **Startup sweep** deletes file-backed rows whose transcript is gone, including archived ones — an orphan shouldn't hide in the archive view.
- **`createSession` preserves `isArchived`.** Un-archiving now happens only on real per-session watcher activity, so an archived session returns when it genuinely sees use, while bulk syncs no longer un-hide everything.

Deletions are hard (the transcript is already gone — nothing to restore); archiving stays soft and reversible.

## Testing

New `sessions-watcher.test.ts` plus `sessions.db.integration.test.ts` additions cover unlink deletion (including archived rows), the opencode/subagent/no-match no-ops, the sweep's include/exclude rules, `isArchived` preservation on both upsert paths, and un-archive-on-activity.

```
npx tsx --tsconfig server/tsconfig.json --test server/modules/database/tests/*.test.ts server/modules/providers/tests/*.test.ts
# tests 69  # pass 69  # fail 0
```

Lint, typecheck, and `npm run build` pass. One existing test asserted the old un-archive behavior and was rewritten — that change is the point of (2).

## Summary by CodeRabbit

* **Bug Fixes**
  * Archived sessions remain archived during synchronization unless new activity is detected.
  * Deleted or missing transcript files now remove their associated sessions, including after server restarts.
  * Session lists update automatically when sessions are removed, clearing removed selections.
  * Legacy session records retain their archive status during updates.

* **Improvements**
  * Archived sessions are restored when new activity is detected.
  * Session cleanup and synchronization now include archived sessions and missing transcript files.


thevinchi:
Both review items addressed:

- d4ef075 guards the per-row delete in `reconcileMissingSessionFiles` (log-and-continue, matching `onUnlink`'s pattern) so one bad row can't abort watcher initialization — includes a regression test.
- 56b5a98 adds the `idx_sessions_jsonl_path` index from the nitpick.

Full suite green: 265/265 tests, typecheck/lint/build clean.

PR #1068, opened 2026-07-29
fix(claude-sync): propagate session renames between Claude CLI and UI

thevinchi:
## What

Session renames never take effect: `/rename` in the Claude CLI doesn't reach the sidebar, and a name is frozen after the first sync. Two guards in `processSessionFile` cause it:

1. It early-returns the stored `custom_name` whenever it isn't the literal `'Untitled Claude Session'` — so once named, a session can never be renamed.
2. It checks the `history.jsonl` map (first-entry-wins) *before* scanning the transcript, so every session with a history entry is named after its **first prompt**, and the existing `custom-title` branch is unreachable in practice.

The `custom-title` parsing is already written — it just never runs.

## How

- **Resolution order**: `custom-title` → `ai-title` → stored name → history first prompt → `last-prompt`, scanning from the end so the newest event of each type wins (matching Claude Code's own last-wins policy).
- **Write-back**: a UI rename appends the same `custom-title` event Claude Code writes on `/rename` and reads back for its session picker, so renames propagate **both** directions and converge on the most recent. Best-effort — never creates a missing file, and a failure is logged rather than failing the already-persisted rename.

Keeping the stored name in the chain (below both transcript titles) means a write-back that failed isn't reverted to the first prompt on next sync. It's non-regressive for existing names, since those *are* the history first prompt.

## Relationship to #982

@swu45's #982 covers the same title-priority reorder and I'd rather not duplicate that work. The difference:

- #982 reorders the scan ahead of the history map but **leaves the early-return guard**, so any session with a stored name still can't be renamed — which is most existing sessions.
- This PR removes that guard and adds the UI → CLI direction.

Happy either way: land #982 first and I'll rebase this down to the guard removal plus write-back, or take this as the superset and close #982 with credit. Both add `server/modules/providers/tests/claude-sessions.test.ts`, so they'll conflict on that path whichever lands first.

## Testing

New `claude-sessions.test.ts` covers `custom-title` beating the history prompt, a later `custom-title` replacing a stored name, the fallback order, subagent transcripts still skipped, the rename append, and rename surviving a missing transcript.

```
npx tsx --tsconfig server/tsconfig.json --test server/modules/providers/tests/claude-sessions.test.ts
# tests 7  # pass 7  # fail 0
```

Lint, typecheck, and `npm run build` pass.

## Summary by CodeRabbit

* **New Features**
  * Claude session names now prioritize custom titles, AI-generated titles, history names, and prompts for more accurate naming.
  * Renaming a Claude session is synchronized with its transcript when available.
  * Session renames remain successful even if the transcript file is unavailable.

* **Bug Fixes**
  * Fixed session rename responses to return completed results.
  * Prevented subagent transcripts from overwriting parent session details.

### DIFF
--- a/server/modules/providers/list/claude/claude-session-synchronizer.provider.ts
+++ b/server/modules/providers/list/claude/claude-session-synchronizer.provider.ts
@@ -18,6 +18,12 @@ type ParsedSession = {
   sessionName?: string;
 };
 
+type TranscriptTitles = {
+  customTitle?: string;
+  aiTitle?: string;
+  lastPrompt?: string;
+};
+
 /**
  * Session indexer for Claude transcript artifacts.
  */
@@ -137,18 +143,21 @@ export class ClaudeSessionSynchronizer implements IProviderSessionSynchronizer {
     // ids must be resolved through the provider-id mapping first.
     const existingSession = sessionsDb.getSessionByProviderSessionId(parsed.sessionId)
       ?? sessionsDb.getSessionById(parsed.sessionId);
-    const existingSessionName = existingSession?.custom_name;
-    if (existingSessionName && existingSessionName !== 'Untitled Claude Session') {
-      return {
-        ...parsed,
-        sessionName: normalizeSessionName(existingSessionName, 'Untitled Claude Session'),
-      };
-    }
+    const existingSessionName = existingSession?.custom_name !== 'Untitled Claude Session'
+      ? existingSession?.custom_name ?? undefined
+      : undefined;
 
-    let sessionName = nameMap.get(parsed.sessionId);
-    if (!sessionName) {
-      sessionName = await this.extractSessionAiTitleFromEnd(filePath, parsed.sessionId);
-    }
+    // The transcript is the source of truth for explicit names: `/rename` in the
+    // Claude CLI and a rename in this UI both land as a `custom-title` event, and
+    // the reverse scan below returns the newest one. The stored name only ranks
+    // above the history.jsonl first prompt so a UI rename whose transcript
+    // write-back failed is not silently reverted to that prompt.
+    const titles = await this.extractSessionTitlesFromEnd(filePath, parsed.sessionId);
+    const sessionName = titles.customTitle
+      ?? titles.aiTitle
+      ?? existingSessionName
+      ?? nameMap.get(parsed.sessionId)
+      ?? titles.lastPrompt;
 
     return {
       ...parsed,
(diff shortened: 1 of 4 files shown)


jjscarafia (approved):
This fixes something we hit in a self-hosted multi-user deployment, and the diagnosis matches what I see in the code. Both guards are still there on 1.36.3: the early return in `processSessionFile`, and `renameSessionById` writing only `updateSessionCustomName` — so a rename in the UI never reaches the transcript and `claude --resume` in a terminal keeps showing the old name. LGTM.

I verified two things against real transcripts in `~/.claude/projects` rather than just reading the diff, and both back up the design:

**1. The write-back event is byte-identical to what the CLI writes.** A real `/rename` in Claude Code produces exactly:

```json
{"type":"custom-title","customTitle":"control-ps4-compra","sessionId":"e7ed0c23-bf1e-4043-8122-3fbf845951e9"}
```

Same three keys, same order as `appendClaudeCustomTitle`. Nothing extra to mirror.

**2. Precedence: `custom-title` over `ai-title` is right, but not because of last-wins.** In a renamed session the CLI re-emits a triplet at the end of every turn, and `ai-title` consistently lands *after* `custom-title`:

```
575:{"type":"last-prompt","lastPrompt":"con steam input se me pusieron los 3 …
576:{"type":"custom-title","customTitle":"control-ps4-compra","sessionId":…
577:{"type":"ai-title","aiTitle":"Elegir control para emuladores en Batocera",…
```

The CLI still shows `control-ps4-compra`. So the rule isn't "newest event wins" globally — it's last-wins *within* a type, with `custom-title` outranking `ai-title` regardless of position. Your resolution chain does exactly that (rank first, break on the newest `custom-title`), so no code change; it's the rationale in the PR description and the docstring on `extractSessionTitlesFromEnd` that reads as strict last-wins, and someone may later "fix" the ranking to match that sentence. Might be worth rewording.

**One thing worth a look: the scan lost its early exit.** The old version returned on the first title event found scanning backwards, and since `ai-title` / `last-prompt` are re-emitted every turn, it stopped within a handful of lines. The new one only breaks on `custom-title`, so for any session that was never renamed — most of them — it now `JSON.parse`s every line of the transcript, on every sync.

The full scan is genuinely necessary: in that same session, `custom-title` first appears at line 100 of 594, i.e. mid-file, so a bounded tail window would miss it. But the parse is avoidable — the title records are tiny and distinctive, so a substring guard before `JSON.parse` skips the message lines that make up the bulk of a transcript:

```js
if (!line.includes('"type":"custom-title"')
    && !line.includes('"type":"ai-title"')
    && !line.includes('"type":"last-prompt"')) {
  continue;
}
```

That should land it faster than the original, not just no slower. Not blocking from my side — just the kind of thing that bites on long sessions.

I didn't run the test suite; the review above is code reading plus the two transcript checks. Thanks for picking this up — it's been open a few weeks and it'd be good to see it land.

Issue #574, opened 2026-03-23
[Feature] Add Kiro (AWS Agentic IDE) support


dgallitelli:
**Is your feature request related to a problem? Please describe.**

CloudUI currently supports Claude Code, Cursor CLI, Codex, and Gemini-CLI as providers. Kiro — AWS's agentic IDE built on Claude — has no support yet, leaving users who work with Kiro unable to manage their sessions through CloudUI.

**Describe the solution you'd like**

Add Kiro as a supported provider in CloudUI, similar to how Codex and Cursor CLI were integrated.

Kiro is available both as a desktop IDE and a CLI tool:

- **CLI install**: `curl -fsSL https://cli.kiro.dev/install | bash`
- **Docs**: https://kiro.dev/docs/
- **Downloads**: https://kiro.dev/downloads/

Kiro uses a `.kiro/` directory for project-level configuration (agents, skills, hooks, steering files, MCP config), analogous to `.claude/` for Claude Code.

Key Kiro concepts relevant to CloudUI integration:
- **Specs** — structured feature planning documents
- **Hooks** — automated tasks triggered by file events (`.kiro.hook` files)
- **Steering** — persistent rules/context files (`.kiro/steering/*.md`)
- **MCP Servers** — same protocol as Claude Code (`.kiro/settings/mcp.json`)
- **Skills** — reusable agent task templates

**Prior art / related work**

Someone has already done exploratory work adapting Claude Code workflows to Kiro's native format:
- PR: [affaan-m/everything-claude-code#548 — Add Kiro IDE support (.kiro/)](https://github.com/affaan-m/everything-claude-code/pull/548) (merged) — demonstrates the dual-format agent structure (JSON for CLI, MD for IDE), hook format, and `.kiro/` directory layout
- Source repo: [ihimanss/kiro-everything-claude-code](https://github.com/ihimanss/kiro-everything-claude-code)

The contribution was reviewed with guidance from an AWS team member (`@aws-hsungmin`), suggesting some level of official interest.

Additionally, there's an open issue in the `awslabs/mcp` repo about Kiro CLI/IDE documentation: [awslabs/mcp#2680](https://github.com/awslabs/mcp/issues/2680).

[...]

PR #1076, opened 2026-07-30
feat(omp): add Oh My Pi (omp) as a first-class agent provider

### LINKED ISSUE #1039: [Feature] Add Oh My Pi (omp) as a supported agent/provider
**Is your feature request related to a problem? Please describe.**

CloudCLI supports several CLI coding agents, but not [Oh My Pi (`omp`)](https://omp.sh/). Users who run `omp` cannot start or resume its sessions, stream its structured tool activity, select its models, or handle permission prompts from CloudCLI's web/mobile UI.

**Describe the solution you'd like**

Add `omp` as a first-class agent/provider, following CloudCLI's existing provider architecture. The integration should:

- Detect an installed `omp` binary and expose its availability in agent settings/onboarding
- Start a session in the selected project and stream assistant, thinking, tool-call, tool-result, and error events into CloudCLI's normalized chat UI
- Surface interactive permission requests and allow approve/reject, abort, and follow-up prompts
- Discover, list, resume, and persist OMP sessions using OMP's existing state rather than creating a separate session store
- Populate the available model list from OMP and support model selection
- Add OMP to the provider selector, provider settings, logos, and enabled-provider controls

The preferred transport appears to be `omp --mode rpc-ui`: OMP documents this NDJSON mode specifically for UI hosts, including tool cards, selectors, dialogs, and `extension_ui_request` frames. Two alternatives are also available:

- `omp acp` exposes the standard Agent Client Protocol over JSON-RPC/stdio and could justify a reusable ACP provider (also relevant to #574)
- `@oh-my-pi/pi-coding-agent` provides a typed Node/TypeScript SDK if in-process embedding fits CloudCLI better

Minimum end-to-end acceptance criteria:

1. Select OMP, open a project, and start a new chat from CloudCLI
2. See structured streaming output and tool status without raw ANSI terminal parsing
3. Approve/reject tool permissions and abort a running turn
4. Reload CloudCLI and resume the same OMP session with its history intact
5. Select an OMP-supported model before or during a session

**Additional context**

- Website: https://omp.sh/
- Source: https://github.com/can1357/oh-my-pi
- SDK/RPC/ACP overview: https://omp.sh/docs/sdk
- Package: https://www.npmjs.com/package/@oh-my-pi/pi-coding-agent
- OMP is cross-platform (macOS, Linux, Windows), supports 40+ model providers, and already exposes dedicated RPC, RPC-UI, ACP, and Node SDK integration surfaces.

materemias:
Closes #1039.

Adds [Oh My Pi (`omp`)](https://omp.sh/) as a first-class agent provider, following the existing provider architecture.

> **Rewritten 2026-08-11.** The original branch had grown to 61 commits and 91 files, mixing the provider with a pile of generic chat fixes. It has been rebuilt from scratch on current `main` as a **5-commit stack that contains the omp provider and nothing else**. Every generic change has been moved off this branch — see [Explicitly out of scope](#explicitly-out-of-scope). This is a force-push over the old tip; the old state is preserved on my fork.

## Review order

This PR's diff is **A + B**, because it sits on top of a prerequisite refactor that is also submitted on its own:

| | Branch | PR | Contents |
|---|---|---|---|
| **A** | `feat/shared-tool-approval-registry` | #1141 | provider-agnostic tool-approval registry (prerequisite) |
| **B** | `feat/omp-provider` | **this PR** | the omp provider |
| **C** | `feat/omp-rich-history` | not opened yet | rich omp transcript UX (follow-up) |

- **Review B alone:** [`feat/shared-tool-approval-registry...feat/omp-provider`](https://github.com/materemias/claudecodeui/compare/feat/shared-tool-approval-registry...feat/omp-provider) — 68 files, +5,077/−281.
- Preview C if useful: [`feat/omp-provider...feat/omp-rich-history`](https://github.com/materemias/claudecodeui/compare/feat/omp-provider...feat/omp-rich-history) — 15 files, +1,291/−93. It stays branch-only until this lands.

When #1141 merges I will rebase A out of this branch with `--force-with-lease`, and this PR's diff becomes B-only. If you would rather review one PR, close #1141 and I will keep A as the first commit here.

## The commits

Each commit compiles and passes its own narrow tests; no commit repairs a knowingly broken earlier one.

**1. `refactor(approvals): extract a shared tool-approval registry`** — 4 files (this is #1141)
Moves the pending-approval map out of the Claude SDK adapter into `server/shared/tool-approval-registry.ts`, so both runtimes share one `chat.permission-response` → `resolveToolApproval` gateway. Claude's behavior is unchanged; the expiry sweep now cancels a stalled waiter instead of dropping it.

**2. `feat(omp): add ACP stdio JSON-RPC transport`** — 2 files
`stdio-jsonrpc-client.ts` plus a 375-line test: newline-framed request/response correlation, notifications, **inbound** requests (the agent asks *us*), process-exit and error handling, malformed frames, unregistered methods.

**3. `feat(omp): add the omp provider backend`** — 27 files, the bulk of the PR
`omp.provider.ts` and its parts: `omp-runtime.provider.ts` (process/session lifecycle, model + thinking configuration, approvals, abort, prompt input, realpath-guarded fs delegation, fork/resume, active-branch selection, hard child retirement), `omp-sessions.provider.ts` (jsonl history → normalized messages, paging), `omp-session-synchronizer.provider.ts`, `omp-auth.provider.ts`, `omp-models.provider.ts`, `omp-mcp.provider.ts`, `omp-skills.provider.ts`, and one shared `omp-session-files.ts` locator so the runtime and the reader cannot drift. Registry entry, capability matrix, and `omp`'s member in the `LLMProvider` union land here too, because TypeScript's exhaustive `Record<LLMProvider, …>` maps make that atomic. Tests live in `server/modules/providers/tests/`.

**4. `feat(omp): wire omp into the server surfaces`** — 16 files
Agent dispatch (`agent.routes.ts`), WebSocket shell/resume, sessions watcher, token-usage service, MCP service, skills, command-model lookup, and `public/api-docs.html`.

**5. `feat(omp): add minimal web UI integration`** — 23 files
Provider union in the frontend types, selector, logo, login/auth pane, agent settings, account/settings state, sidebar label, i18n string. Deliberately minimal: no new transcript widgets, no changes to generic chat behavior.

## Transport: ACP, not `rpc-ui`

The issue suggested `omp --mode rpc-ui`. I prototyped both and went with **`omp acp`** (ACP over JSON-RPC/stdio):

- **Structured approvals.** ACP `session/request_permission` carries the real option set (`allow_once` / `allow_always` / `reject_once` / `reject_always`), which maps onto the existing permission UI. `rpc-ui`'s `extension_ui_request` only offers a coarse Approve/Deny.
- **Real session resume.** ACP `session/fork` rehydrates an on-disk session — full history, plus a `parentSession` link the omp TUI itself can navigate.
- **Reusable surface.** ACP is a standard, so this adapter is a step toward a generic ACP provider (relevant to #574) rather than an omp-only dialect.

Happy to revisit if you'd prefer `rpc-ui`.

## Acceptance criteria from #1039

1. **Select omp, open a project, start a chat** — registered end to end: selector, settings, logo, enabled-provider controls, binary detection.
2. **Structured streaming, no ANSI parsing** — assistant/thinking chunks, tool calls, tool results and errors arrive as ACP `session/update` notifications and are normalized into the existing chat message kinds.
3. **Approve/reject + abort** — approvals gated per session by permission mode (`bypassPermissions` auto-allows, `default` prompts, `plan` auto-denies sensitive tools); abort via `session/cancel`.
4. **Reload and resume with history intact** — resume walks `session/load` (in-process) → **`session/fork`** (on-disk, full history) → `session/new` as a last resort. Verified across a real server restart.
5. **Model selection** — catalog from `omp models --json`; omp resolves the effective model itself, so a "use omp's configured model" sentinel is the default rather than forcing a choice. Post-turn telemetry reports the model that actually ran, not the sentinel.

## Notable design points

- **One `omp acp` child per cwd, multiplexing many sessions.** `session/update` and inbound `session/request_permission` are routed to the owning run by session id; each run emits exactly one terminal `complete`.
- **History comes from omp's own store** (`~/.omp/agent/sessions/**.jsonl`) — no second session store. omp's on-disk shape is *not* Anthropic-shaped: tool calls are `{type:'toolCall', name, arguments}` inside the assistant message, results are separate `role:'toolResult'` entries linked by `toolCallId`. An omp jsonl is also a **tree**, not a list — the reader walks the newest conversation's branch so a session continued from the TUI doesn't read as two interleaved dialogues.
- **fs delegation is realpath-guarded** and confined to the working directory; image attachments come from omp's content-addressed blob store with a mime allowlist and a size cap.

## Two defects found while rebuilding

Both were present in the old branch and are fixed here, with regression tests:

1. **Resume was silently broken.** Every other runtime maps the app session id to the provider-native id via `context.resolveProviderSessionId()`; the omp runtime passed the app id straight to `session/load` and `session/fork`, so both failed and the run started blank ("history lost") while looking successful. Tests now cover the mapping in both directions.
2. **Abort reported success while the child kept running.** The run was keyed by native id but `activeOmpSessions` by app id, so `session/cancel` never reached the live child.

## Explicitly out of scope

Removed from this branch on purpose, so the diff stays reviewable:

- **Rich omp transcript UX** (todo widget, advisor-note weaving, per-tool views, image rendering in history, collaborator turns) → branch C above, after this lands.
- **Generic chat fixes** that were tangled into the old branch and are *not* omp-specific: LaTeX/`\text` escape handling, large-JSON/`TextContent` folding, a fallback row for tools with no config, reasoning/background stream buffers, scroll pinning, queued-message auto-send, chat back/top navigation, generic resume recovery, a generic command-model service, and pre-turn composer model display. These stay on a local branch and are not submitted here. Say the word and I'll open them separately.
- **Onboarding.** omp is absent from the "Connect Agents" onboarding step (`AgentConnectionsStep.tsx`), which keeps its own hard-coded list. The old branch didn't touch it either; I left it alone rather than widening this PR.
- **`write` cannot be gated by this UI.** omp routes only `bash`/`edit`/`delete`/`move` through ACP `session/request_permission`; `write` it resolves itself. A denied `bash` can therefore be followed by an ungated `write` — documented rather than worked around, since the fix belongs in omp.

## Verification

At `f038dfe`:

- `npm run typecheck` — clean
- `npm run lint` — 0 errors (including the `boundaries/*` rules, which is what checks the barrel and shared-file discipline)
- `npm run build` — passes
- server tests — **323/323 pass**, run serialized:
  `npx tsx --tsconfig server/tsconfig.json --test --test-concurrency=1 "server/**/*.test.ts" "server/**/*.test.js"`

New coverage runs the ACP adapter against a fake `omp` process (no paid calls): transport framing, approval allow/deny/plan-mode/bypass, cross-session isolation, fs-guard escapes, abort during startup, resume via `session/load`, the app↔native id mapping, the jsonl normalizer, the SQLite synchronizer, and history paging.

**Caveat, pre-existing on `main`:** with the default parallel test workers, `server/modules/agent/tests/agent.routes.test.ts` dies with Node's `Unable to deserialize cloned data due to invalid or unsupported version`. It reproduces identically on a pristine `main` checkout (262/263 there), which is why the numbers above come from a serialized run.

**Live smoke path** against a running server on an isolated port and database, with omp 17.2.13 and a cheap model: install/auth detection, model catalog + sentinel, streaming turn, gated file write with allow / deny / bypass / plan mode, abort of an active turn, **server restart → reopen → resume with prior context intact**, history paging, skills listing, actual-turn model, context window and token telemetry — then the same path in the browser: select omp, authenticate, run a tool-using turn, baseline tool rendering, reopen the session.

## Trying it

Requires `omp` on `PATH` (developed against 17.0.6 – 17.2.13). Pick **omp** in the provider selector, open a project, and chat. To exercise resume, start a session in the `omp` TUI, then open it from the web UI and send a prompt — it continues that conversation as a new branch of the original, which stays intact and navigable from the TUI's session tree.

Screenshots: my day-to-day omp transcripts are real work, so the smoke run above used a throwaway project. Happy to add screenshots or a recording from it if that's a merge requirement.

---

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -162,6 +162,7 @@ export default tseslint.config(
             "server/shared/frontmatter.ts",
             "server/shared/claude-cli-path.ts",
             "server/shared/image-attachments.ts",
+            "server/shared/tool-approval-registry.ts",
           ], // classify shared utility files so modules can depend on them explicitly
           mode: "file",
         },
--- a/public/api-docs.html
+++ b/public/api-docs.html
@@ -489,7 +489,7 @@
                                 <span class="endpoint-path"><span class="api-url">http://localhost:3001</span>/api/agent</span>
                             </div>
 
-                            <p>Trigger an AI agent (Claude, Cursor, Codex, or OpenCode) to work on a project.</p>
+                            <p>Trigger an AI agent (Claude, Cursor, Codex, OpenCode, or omp) to work on a project.</p>
 
                             <h4>Request Body Parameters</h4>
                             <table>
@@ -524,7 +524,7 @@
                                         <td><code>provider</code></td>
                                         <td>string</td>
                                         <td><span class="badge badge-optional">Optional</span></td>
-                                        <td><code>claude</code>, <code>cursor</code>, <code>codex</code>, or <code>opencode</code> (default: <code>claude</code>)</td>
+                                        <td><code>claude</code>, <code>cursor</code>, <code>codex</code>, <code>opencode</code>, or <code>omp</code> (default: <code>claude</code>)</td>
                                     </tr>
                                     <tr>
                                         <td><code>stream</code></td>
@@ -839,6 +839,7 @@ data: {"type":"done"}</code></pre>
             { id: 'codex',    name: 'OpenAI' },
             { id: 'cursor',   name: 'Cursor' },
             { id: 'opencode', name: 'OpenCode' },
+            { id: 'omp',      name: 'omp' },
         ];
 
         async function populateModels() {
--- a/server/index.ts
+++ b/server/index.ts
@@ -83,6 +83,7 @@ const queryClaude = providerRuntimeService.getRunner('claude');
 const queryCursor = providerRuntimeService.getRunner('cursor');
 const queryCodex = providerRuntimeService.getRunner('codex');
 const queryOpenCode = providerRuntimeService.getRunner('opencode');
+const queryOmp = providerRuntimeService.getRunner('omp');
 const gitRoutes = createGitModule({
     queryClaude,
     queryCursor,
@@ -92,6 +93,7 @@ const agentRoutes = createAgentModule({
     queryCursor,
     queryCodex,
     queryOpenCode,
+    queryOmp,
 });
 
 // Single WebSocket server that handles chat, shell, and plugin proxy paths.
--- a/server/modules/agent/agent.module.ts
+++ b/server/modules/agent/agent.module.ts
@@ -17,7 +17,7 @@ import { createAgentRouter } from './agent.routes.js';
 
 type AgentExternalDependencies = Pick<
   Parameters<typeof createAgentRouter>[0],
-  'queryClaude' | 'queryCursor' | 'queryCodex' | 'queryOpenCode'
+  'queryClaude' | 'queryCursor' | 'queryCodex' | 'queryOpenCode' | 'queryOmp'
 >;
 
 /**
--- a/server/modules/agent/agent.routes.ts
+++ b/server/modules/agent/agent.routes.ts
@@ -22,6 +22,7 @@ type AgentRouterDependencies = {
   queryCursor: ProviderRunFunction;
   queryCodex: ProviderRunFunction;
   queryOpenCode: ProviderRunFunction;
+  queryOmp: ProviderRunFunction;
   GithubClient: typeof import('@octokit/rest').Octokit;
 };
 
@@ -44,6 +45,7 @@ export function createAgentRouter(dependencies: AgentRouterDependencies): expres
   const spawnCursor = dependencies.queryCursor;
   const queryCodex = dependencies.queryCodex;
   const spawnOpenCode = dependencies.queryOpenCode;
+  const spawnOmp = dependencies.queryOmp;
   const Octokit = dependencies.GithubClient;
   const router = express.Router();
 
@@ -662,7 +664,7 @@ export function createAgentRouter(dependencies: AgentRouterDependencies): expres
    *                          - Source for auto-generated branch names (if createBranch=true and no branchName)
    *                          - Fallback for PR title if no commits are made
    *
-   * @param {string} provider - (Optional) AI provider to use. Options: 'claude' | 'cursor' | 'codex' | 'opencode'
+   * @param {string} provider - (Optional) AI provider to use. Options: 'claude' | 'cursor' | 'codex' | 'opencode' | 'omp'
    *                           Default: 'claude'
    *
    * @param {boolean} stream - (Optional) Enable Server-Sent Events (SSE) streaming for real-time updates.
@@ -785,7 +787,7 @@ export function createAgentRouter(dependencies: AgentRouterDependencies): expres
    * Input Validations (400 Bad Request):
    *   - Either githubUrl OR projectPath must be provided (not neither)
    *   - message must be non-empty string
-   *   - provider must be 'claude', 'cursor', 'codex', or 'opencode'
+   *   - provider must be 'claude', 'cursor', 'codex', 'opencode', or 'omp'
    *   - createBranch/createPR requires githubUrl OR projectPath (not neither)
    *   - branchName must pass Git naming rules (if provided)
    *
@@ -896,8 +898,8 @@ export function createAgentRouter(dependencies: AgentRouterDependencies): expres
       return res.status(400).json({ error: 'message is required' });
     }
 
-    if (!['claude', 'cursor', 'codex', 'opencode'].includes(provider)) {
-      return res.status(400).json({ error: 'provider must be "claude", "cursor", "codex", or "opencode"' });
+    if (!['claude', 'cursor', 'codex', 'opencode', 'omp'].includes(provider)) {
+      return res.status(400).json({ error: 'provider must be "claude", "cursor", "codex", "opencode", or "omp"' });
     }
 
     // Validate GitHub branch/PR creation requirements
@@ -1025,6 +1027,21 @@ export function createAgentRouter(dependencies: AgentRouterDependencies): expres
           model: model || opencodeModels.DEFAULT,
           effort,
           permissionMode: 'bypassPermissions' // Agent runs are non-interactive, like the other providers above
+        }, writer);
+      } else if (provider === 'omp') {
+        console.log('Starting omp ACP session');
+
+        await spawnOmp(message.trim(), {
+          projectPath: finalProjectPath,
+          cwd: finalProjectPath,
+          sessionId: sessionId || null,
+          // No catalog default here: omp resolves its own configured model when the
+          // caller doesn't name one (OMP_CONFIGURED_MODEL_SENTINEL).
+          model,
+          effort,
+          // Agent runs are non-interactive: the mode drives per-session approval,
+          // so bypass auto-allows instead of stalling on a prompt nobody answers.
+          permissionMode: 'bypassPermissions'
         }, writer);
       }
 
--- a/server/modules/agent/tests/agent.routes.test.ts
+++ b/server/modules/agent/tests/agent.routes.test.ts
@@ -34,6 +34,7 @@ function createDependencies(
     queryCursor: unexpectedProviderCall as AgentDependencies['queryCursor'],
     queryCodex: unexpectedProviderCall as AgentDependencies['queryCodex'],
     queryOpenCode: unexpectedProviderCall as AgentDependencies['queryOpenCode'],
+    queryOmp: unexpectedProviderCall as AgentDependencies['queryOmp'],
     GithubClient: class {} as unknown as AgentDependencies['GithubClient'],
     ...overrides,
   };
--- a/server/modules/commands/commands.routes.ts
+++ b/server/modules/commands/commands.routes.ts
@@ -28,13 +28,14 @@ const providerModelsService = dependencies.models;
 const process = dependencies.runtime;
 const router = express.Router();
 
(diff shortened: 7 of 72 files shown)


materemias:
Worked through the automated review. Two commits on top: `4f54644` (review findings) and `2b3955a` (TypeScript migration).

## Fixed

**Real bugs**

- **Forward the requested omp model** (`agent.routes.ts`) — confirmed: the runtime reads `options.model` *and* `options.effort` (effort maps onto omp's thinking level), so `/api/agent` callers could select neither. Both are forwarded now, with no catalog default since omp resolves its own configured model.
- **Destroy the read stream on early break** (`omp-session-synchronizer`) — correct, `rl.close()` leaves the `fs.ReadStream` open. Now holds a handle and destroys it in `finally`. Checked the other reader (`readOmpJsonl`) — it consumes to EOF, so it self-closes and needed no change.
- **Schedule expiry independently of registration** (`tool-approval-registry`) — this one mattered more than it looks. The Claude runtime applies its own per-request timeout, but the omp ACP handler awaits its resolver with no timeout at all, so a last stalled approval really could hang forever. Added an unref'd sweep that only runs while approvals are outstanding and stops when the map drains.
- **`clearApprovalsForSession` resolving nothing** — it had zero callers (and didn't exist before this branch), so I deleted it rather than adding resolve semantics to dead code. Happy to reinstate it properly if you want session-scoped cleanup.
- **Collision-free protection tokens** (`chatFormatting`) — real: literal `__PROTECTED_BLOCK_0__` in a message could be overwritten. Now derives a NUL-prefixed token verified absent from the input, plus a regression test.

**Also correct**

- Global MCP add is no longer hidden for providers with no writable scopes — only the provider-specific entry is. That matches what the comment there already claimed, and the service does skip unwritable providers while still applying the config to the rest.
- omp's status dot in the agent selector now matches `agentConfig` in `AgentListItem` (emerald instead of falling through to the default).
- Stale `omp-mcp.provider` doc comment corrected (the global helpers skip omp rather than reporting `created:false`), and omp added to the `/api/agent` description in `api-docs.html`.

## TypeScript migration

The four "this should be `.ts`" comments were right, and I'd initially misjudged them as grandfathered because all four sibling provider runtimes are `.js`. `.agents/skills/backend-module-standards` is explicit — *"Use TypeScript for every file inside `server/modules/`. Do not add JavaScript backend module files."* — and these are new files. So `omp-runtime.provider`, both its tests, and `shared/tool-approval-registry` are now TypeScript, fully typed under the backend's `strict` config (they had been invisible to the checker, since `checkJs` is false). Typing surfaced no behaviour changes and the runtime's 10 tests still pass.

Two side-effects worth flagging:

- The registry stayed a dedicated `shared/tool-approval-registry.ts` rather than moving into `shared/utils.ts`. It owns module state (a Map plus a timer) and `utils.ts` is otherwise entirely stateless helpers, so folding it in seemed worse than the standard's letter — say the word if you'd rather it lived there.
- `notification-orchestrator.service.js` needed JSDoc param types: its `sessionId = null` / `sessionName = null` defaults make TypeScript infer `null`-only params, so the first TS caller can't pass a real value. Two comment blocks, no behaviour change.

## One I pushed back on

**Reject resume for an already-active session id** — I don't think this reproduces. `chatRunRegistry.startRun` already returns `null` when a run for that session is `running`, so the chat path can't start the second run; and the Claude runtime does the same unconditional `activeSessions.set(sessionId, …)`. Adding the check only to omp would make it behave differently from all four siblings with no reachable failure. Glad to add it across the runtimes if you see a path I've missed.

## Open question

The "Oh My Pi" vs `omp` labelling is a branding call I'd rather leave to you. Upstream siblings use product names (Claude, Cursor, Codex, OpenCode), which argues for "Oh My Pi"; omp's own CLI and TUI brand themselves lowercase `omp`, which is what the UI shows today. Trivial to switch if you prefer the full name.

## Verification

`npm run typecheck` (0 errors) · 286 server tests · `npm run lint` (0 errors) · `npm run build` — all pass.

One note in case it bites you too: `agent.routes.test.ts` intermittently fails in the whole-suite run with a Node test-runner `Unable to deserialize cloned data` IPC error. It passes 5/5 in isolation and it predates this branch's changes to that area.

materemias:
**This branch was reset — the review threads above no longer apply.**

I rebuilt it from scratch on current `main`. It went from **61 commits / 91 files / +7,246 −538** to **5 commits / 72 files / +5,476 −317**:

- the shared tool-approval registry was extracted as a standalone prerequisite and opened separately as #1141 — it is the first commit here and will drop out of this PR once that one lands;
- the four remaining commits are the omp provider itself: ACP stdio transport, provider backend, server-surface wiring, minimal web UI;
- rich omp transcript widgets moved to a follow-up branch, and the generic chat/WebSocket fixes that used to ride along are gone from this PR entirely — two of them are filed as their own issues (#1136, #1137).

The PR body has been rewritten to describe the current state. Please read that rather than the automated summary at the top of this thread, which describes the previous branch. The inline suggestions from the earlier passes are anchored to code that no longer exists in the diff.

Two real defects surfaced during the rebuild and are fixed here with regression tests: the runtime was passing the app session id where every other runtime resolves the provider-native one (so `session/load` and `session/fork` both failed and resume silently started blank), and abort keyed the run by native id while the session map keyed by app id (so `session/cancel` never reached the child).

materemias:
Closing this in favour of a clean PR, per @Haile's request — the branch is unchanged, but this thread's review history is against code that no longer exists (61 commits and 91 files of it), and the auto-generated summary at the top still describes that old branch.

The replacement carries the identical 5-commit branch with a fresh body and no stale context. Link in the next comment.

materemias:
Replacement: #1143 — same branch, same 5 commits, clean thread.

PR #999, opened 2026-07-11
feat(chat): allow custom model IDs

zhoubeiqing:
## Summary

- allow users to enter a custom model ID from the new-conversation model selector
- add a custom model action to the in-session `/models` dialog, including Enter-to-apply
- persist explicitly selected custom defaults so provider catalog refreshes do not reset them
- clear the custom marker when the user returns to a catalog model

## Motivation

Provider adapters already pass model values through to their SDK/CLI integrations, but the UI only allowed selecting values present in the provider catalog. The catalog reconciliation logic also replaced any non-catalog value with the provider default.

This provides a manual fallback for custom Anthropic-compatible gateways and other providers that expose model IDs CloudCLI cannot discover. It addresses the selection limitation described in #910 and is complementary to #975: dynamically discovered models can still appear in the catalog, while unlisted models remain usable through explicit input.

## Behavior

- exact catalog model IDs continue to use the existing selection flow
- a non-empty value that is not an exact catalog ID is offered as a custom model
- existing sessions apply the custom model on the next response
- new-session defaults survive reloads and model catalog refreshes

## Validation

- `npm run typecheck`
- ESLint on the three changed files (0 errors; one existing Tailwind arbitrary-value warning in `CommandResultModal.tsx`)
- `npm run build:client` (successful; existing CSS minifier and bundle-size warnings remain)

## Summary by CodeRabbit

- **New Features**
  - Added support for entering and selecting custom model IDs for each provider.
  - Model search now offers a “Use custom model ID” option when no matching model is found.
  - Custom model IDs can be selected by clicking the option or pressing Enter.
  - Custom selections are remembered and remain available even when not listed among current models.
  - Improved model selector prompts to clarify that users can search or enter a custom ID.

PR #1095, opened 2026-08-03
Fix/update codex sdk and fix data usage

### LINKED ISSUE #910: Claude provider model list should support gateway/custom models
## Problem

When using CloudCLI with Claude Code configured for a custom Anthropic-compatible gateway, the Claude provider model list in CloudCLI still returns only the hardcoded fallback Claude models.

This means custom gateway models cannot be selected from `/models`, even though Claude Code itself can be configured to use the gateway via environment variables.

## Environment

- CloudCLI: 1.34.0
- Claude Code: 2.1.176
- Self-hosted CloudCLI
- Claude provider authenticated via API token / custom gateway
- Gateway exposes `/v1/models` and `/v1/messages`

Claude Code environment is configured with variables like:

```bash
ANTHROPIC_BASE_URL=http://127.0.0.1:8080
ANTHROPIC_AUTH_TOKEN=...
CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY=1
```

The gateway `/v1/models` returns custom model IDs such as:

```text
gpt-5.3-codex
gpt-5.4-mini
codex-auto-review
```

## Current Behavior

CloudCLI's endpoint:

```text
GET /api/providers/claude/models
```

returns only the static Claude fallback models, for example:

```text
default
fable
sonnet
sonnet[1m]
opus[1m]
haiku
```

As a result, `/models` in the CloudCLI chat UI cannot select gateway-provided custom model IDs.

## Expected Behavior

CloudCLI should provide a supported way to include gateway/custom model IDs in the Claude provider model list.

Possible approaches:

1. Re-enable safe dynamic model discovery when Claude Code gateway model discovery is enabled, without creating unwanted side-session/jsonl artifacts.
2. Add explicit environment-variable configuration, for example:

```bash
CLOUDCLI_CLAUDE_MODELS=gpt-5.3-codex,gpt-5.4-mini,codex-auto-review
CLOUDCLI_CLAUDE_DEFAULT_MODEL=gpt-5.3-codex
```

3. Add a settings UI field for custom Claude provider model IDs.

## Related Context

This appears related to PR #806, which intentionally disabled runtime `supportedModels()` discovery and always serves a static fallback list.

It is also adjacent to #234 and #132, but those discuss proxy/custom endpoint integration more broadly. This issue is specifically about the Claude provider `/models` list not exposing gateway/custom model IDs.

## Workaround

I was able to work around this locally by patching `dist-server/server/modules/providers/list/claude/claude-models.provider.js` to append the custom gateway model IDs to `CLAUDE_FALLBACK_MODELS.OPTIONS`, but that is fragile and will be overwritten on upgrade.

### LINKED ISSUE #1043: Model picker never shows new Claude models — getSupportedModels() is disabled, returns a hardcoded stale list
**Version:** 1.36.3

## Summary
The model selection menu only ever shows a fixed set of 7 hardcoded model cards with hand-written labels. New Claude models (e.g. Sonnet 5) never appear, and the existing labels are stale/misleading.

## Root cause (in your own code)
`dist-server/server/modules/providers/list/claude/claude-models.provider.js`

`getSupportedModels()` has the live model query commented out and simply returns the hardcoded fallback:

```js
async getSupportedModels() {
    // claude creates a new jsonl file as a separate session for this request.
    // As a result, it lists the workspace where this is invoked when it shouldn't.
    //
    // Disabled for now:
    // const queryInstance = query({ prompt: 'Get supported models', options: buildClaudeQueryOptions() });
    // const supportedModels = await queryInstance.supportedModels();
    // queryInstance.close();
    // return buildClaudeModelsDefinition(supportedModels);
    return CLAUDE_FALLBACK_MODELS;   // <-- frozen 7-card list
}
```

Because dynamic discovery is disabled, `CLAUDE_FALLBACK_MODELS` (7 fixed cards) is the only source, and its labels are hand-maintained text that goes stale over time.

## Impact / concrete example
- The card labeled "Sonnet" / "Sonnet 4.6" sends alias `sonnet`, which the underlying Claude CLI (2.1.207) resolves to `claude-sonnet-5`. So the card silently delivers Sonnet 5 but is labeled 4.6.
- Same for "Sonnet (1M context)": labeled 4.6, resolves to Sonnet 5 (1M).
- Any Claude model not in the 7-card list is unreachable from the UI (search only filters the 7 fixed cards; no custom model entry).

## Why the side-effect happens
The disabled path used `query({ prompt: 'Get supported models', ... })`, which spins up a full agent session and creates a stray `.jsonl` session file, causing the invoking workspace to show up in the session/workspace listing where it shouldn't.

## Suggested fix (re-enable discovery without the side-effect)
- Fetch the model list via a read-only path that does NOT create an agent session (dedicated model-list call/endpoint), instead of a prompt-based `query()`; OR
- Tag the transient discovery session so it is excluded from the workspace/session listing; OR
- At minimum, keep the labels in sync with the actual alias resolution so cards don't show a wrong model name.

## Repro
1. Open the model picker in a fresh cloudcli install.
2. Observe only the 7 hardcoded cards; select "Sonnet" — the running model is `claude-sonnet-5`, not the "4.6" shown.
3. Note there is no way to pick a newer model that isn't one of the 7 cards.

### LINKED ISSUE #1082: Unbounded file-tree traversal can exhaust the server heap for broad project roots
## Describe the bug

`buildFileTree` materializes a project tree to depth 10 with recursive `Promise.all` calls and no node budget. When a broad but valid project root such as a user home directory contains enough ordinary files, one file-tree request can consume several gigabytes of heap, saturate a CPU core, block API responses, and eventually crash-loop the server with `Reached heap limit`.

This is the ordinary-files follow-up called out as out of scope in #1074. That PR prevents recursion into forbidden pseudo-filesystems, but valid roots can still contain more entries than the server can safely materialize.

## Reproduction

1. Register a broad project root containing many ordinary files, such as a populated home directory.
2. Select that project and open or restore the Files tab.
3. The frontend requests `GET /api/file-tree/projects/:projectId/files?respectGitignore=true`.
4. Observe the Node server CPU and RSS while `buildFileTree` recursively walks the root.

## Actual behavior

On a real self-hosted instance with `/home/ubuntu` registered as a project:

- Node held approximately 105% CPU.
- RSS grew beyond 2.8 GB and continued rising.
- The UI remained on `Loading CloudCLI / Setting up your workspace...` because API work was starved.
- The previous service process reached approximately 4 GB before terminating with `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`.
- systemd restarted the service, after which the same request started the cycle again.

## Expected behavior

A file-tree request must have a bounded memory and work cost. Oversized roots should fail quickly with a controlled response or return an explicitly truncated result, while the rest of the server remains responsive.

## Suggested fix

- Apply a global entry budget to one recursive tree build.
- Traverse directory branches without recursively creating all pending promises up front.
- Return a controlled error when the budget is exceeded.
- Add a regression test for an oversized directory listing.

## Related

- #1074 prevents recursion into forbidden system paths and explicitly notes this remaining ordinary-file OOM case.
- #1025 is another open OOM report, but it concerns Git diff handling rather than file-tree traversal.

### LINKED ISSUE #1088: [Bug] Mobile: Session rename button missing (hover-only, no tap/long-press fallback)
### Environment
- CloudCLI version: 1.37.0 (latest at time of report)
- Platform: Mobile web — Chrome on Android (large foldable). Reproducible on any touch-only device.
- Server: self-hosted, accessed via browser at `http://<server>:3001`

### Description

On desktop, hovering a session item in the sidebar reveals two action buttons: **Delete** and **Rename**. On mobile (touch device, no hover), the session item only shows the Delete button — the **Rename** button is missing entirely**, so there's no way to rename a session from a phone.

### Steps to reproduce
1. Open CloudCLI in a mobile browser (Chrome on Android/iOS).
2. Open a project's session list.
3. Look at the actions available on a session item.

**Expected**: A Rename action is reachable via tap or long-press (since `hover` is unavailable on touch devices).
**Actual**: Only Delete is available. No rename entry point on mobile.

### Root cause (investigated)

The desktop rename button is revealed via a `group-hover`-based CSS rule (e.g. `group-hover:text-foreground`). Touch devices never fire `hover`, so the button never appears. The mobile (`isMobile`) UI branch exposes a tap entry for **Delete** but did not add one for **Rename**.

The backend and frontend logic fully support renaming — only the mobile UI entry point is missing:

- API: `PUT /api/providers/sessions/:id` with body `{"summary":"<new name>"}`
- Frontend `renameSession(sessionId, summary)` exists and works on desktop
- `renameSessionById()` writes `custom_name` in the DB

### Suggested fix

Pick one:
1. Add a tap / long-press handler on mobile session items to trigger rename (mirroring how Delete is exposed), or
2. Make action buttons always visible on touch devices, e.g. `flex md:opacity-0 md:group-hover:opacity-100` — shown on mobile, hidden-until-hover on desktop.

### Workaround
- Rename from the desktop web UI — `custom_name` syncs to all devices via the shared DB, so mobile shows the new title immediately.
- Or call the API directly: `PUT /api/providers/sessions/<id>` with `{"summary":"..."}`.

### LINKED ISSUE #1089: [Bug] v1.37.0 Codex wrapped exec tool calls are silently dropped except tools.shell_command
## Summary

After upgrading from CloudCLI `1.36.3` to `1.37.0`, Codex conversations show assistant messages but no tool-call or tool-result blocks. The underlying Codex rollout JSONL still contains both `custom_tool_call` and `custom_tool_call_output` records, so this is a rendering/history-conversion regression rather than data loss or tools not executing.

## Environment

- CloudCLI: `1.37.0`
- Codex CLI: `0.146.0`
- Node.js: `24.18.0`
- OS: Ubuntu 24.04
- Integration: external Codex CLI

## Regression and root cause

This appears to come from PR #1037, specifically commit 4efd314 (`fix(codex): render wrapped tool history`).

`translateCodexExecInput()` recognizes only wrapper inputs containing:

```js
tools.shell_command(...)
```

For every other `custom_tool_call` whose name is `exec`, it returns `null`. The caller then adds the call ID to `ignoredToolCallIds` and skips both the tool call and its matching output.

Recent Codex rollouts use the generic `exec` wrapper for many nested operations, for example:

```js
await tools.exec_command(...)
await tools.apply_patch(...)
await tools.web__run(...)
await tools.update_plan(...)
```

These therefore disappear from the CloudCLI conversation while assistant text remains. In one recent local transcript, all 32 `exec` calls used non-`shell_command` nested tools and all fell into this ignored path.

## Steps to reproduce

1. Run CloudCLI `1.37.0` with the external Codex integration.
2. Start a Codex turn that invokes a wrapped non-shell tool such as `update_plan`, `apply_patch`, `web__run`, or `exec_command`.
3. Let the turn complete, then view or reopen the conversation.
4. Observe that the assistant message is present but the wrapped tool call and result are absent.

## Expected behavior

Every persisted tool call and result should remain visible. If a nested `exec` wrapper cannot be translated to a specialized UI tool, it should at least fall back to a generic `exec` tool row rather than being silently dropped.

## Actual behavior

Only `exec` inputs containing `tools.shell_command(...)` are rendered. Other wrapped calls and their results are silently omitted.

## Suggested fix

- Do not add unrecognized `exec` wrappers to `ignoredToolCallIds`.
- Fall back to rendering the original `exec` input and matching output, or generically parse `tools.<method>(...)` while preserving a safe fallback.
- Add regression tests for `exec_command`, `apply_patch`, `web__run`, `update_plan`, and an unknown wrapper so no persisted tool/result pair can disappear silently.

Relevant code:

- https://github.com/siteboon/claudecodeui/blob/v1.37.0/server/modules/providers/list/codex/codex-sessions.provider.ts#L169-L187
- https://github.com/siteboon/claudecodeui/blob/v1.37.0/server/modules/providers/list/codex/codex-sessions.provider.ts#L506-L568
- https://github.com/siteboon/claudecodeui/commit/4efd314df2ba115d775997c6289b0163a8cf36cb

### LINKED ISSUE #1093: No GPT 5.6 Option
Image

### Summary
When using a custom / third-party OpenAI-compatible API endpoint, the model selection dropdown is restricted to a hardcoded list (`gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini`). There is currently no way to manually type a custom model name or select newer models like `gpt-5.6`.

### Current Problem
- The model selection menu only offers a fixed dropdown with limited options.
- Third-party API users cannot specify custom model IDs or newer OpenAI models.
- Changing `OPENAI_BASE_URL` or configuration does not update the available model list.

### Proposed Solution
1. **Allow Custom Text Input:** Add an option in the model selector or settings to manually enter a model name string (e.g., `gpt-5.6`, custom proxy identifiers).
2. **Dynamic / Configurable Model List:** Support defining available models via environment variables (e.g., `AVAILABLE_MODELS=...`) or fetching them dynamically from the configured `/v1/models` endpoint.
3. **Update Default Model List:** Add newer models like `gpt-5.6` to the default static list.

### Context / Environment
- **UI:** siteboon/claudecodeui
- **Provider:** OpenAI / Third-party OpenAI-compatible API

blackmammoth:
Closes #1088 
Closes #1082 

Supersedes PR #1083 
Supersedes PR #999 
Supersedes PR #1127 

This pull request introduces a new approach for naming chat sessions created via CloudCLI: the session name is now derived from the user's initial message, limited to four words, and is preserved as the authoritative title even after provider synchronization. The changes affect session creation logic, database handling, API endpoints, and tests to ensure consistent behavior across the stack.

**Session naming improvements:**

* When a new session is created from CloudCLI, the name is now set to the first four whole words of the initial user message, or "Untitled Session" if the message is empty. This name is passed through the stack and stored in the database. (`server/modules/providers/services/sessions.service.ts` [[1]](diffhunk://#diff-e23331b704fea9d841288d3bd8c506187ff5fd65b725cad09e7dac58dfbb2084R55-R61) [[2]](diffhunk://#diff-e23331b704fea9d841288d3bd8c506187ff5fd65b725cad09e7dac58dfbb2084L157-R173) [[3]](diffhunk://#diff-e23331b704fea9d841288d3bd8c506187ff5fd65b725cad09e7dac58dfbb2084L169-R190); `server/modules/database/repositories/sessions.db.ts` [[4]](diffhunk://#diff-1c8865f514e40d9b0f615f35cd097a2819ba18ce4ea68aa22a30bbdef10bf5fcL154-R180); `server/modules/providers/provider.routes.ts` [[5]](diffhunk://#diff-71fc6416a32d886c20cce49a0c5082d012aac009ae1dfbc503f77a00b1857cfaL555-R556); `src/components/chat/hooks/useChatComposerState.ts` [[6]](diffhunk://#diff-100f83f484b27e9ace28d0086f81d0904a7556c8f7ecc8bfad4f8d5877346d59R840-R857)
* The session name is now included in the API response when a session is created, and the frontend uses this value. (`server/modules/providers/services/sessions.service.ts` [[1]](diffhunk://#diff-e23331b704fea9d841288d3bd8c506187ff5fd65b725cad09e7dac58dfbb2084R20); `src/components/chat/hooks/useChatComposerState.ts` [[2]](diffhunk://#diff-100f83f484b27e9ace28d0086f81d0904a7556c8f7ecc8bfad4f8d5877346d59R840-R857)

**Database and synchronizer logic updates:**

* The database logic ensures that the custom session name set at creation is only overwritten by provider synchronizers if the session originated from the provider (not the app/CloudCLI). (`server/modules/database/repositories/sessions.db.ts` [[1]](diffhunk://#diff-1c8865f514e40d9b0f615f35cd097a2819ba18ce4ea68aa22a30bbdef10bf5fcL106-R111) [[2]](diffhunk://#diff-1c8865f514e40d9b0f615f35cd097a2819ba18ce4ea68aa22a30bbdef10bf5fcL133-R142)
* Synchronizer logic for both Codex and OpenCode providers is simplified: they now preserve the app-assigned session name for app-created sessions, never overwriting it with provider-discovered titles. (`server/modules/providers/list/codex/codex-session-synchronizer.provider.ts` [[1]](diffhunk://#diff-d5c0876f85bd9c90c48d2f96244f39cb7536f8d53aa03affbaf129888890f477L137-R137) [[2]](diffhunk://#diff-d5c0876f85bd9c90c48d2f96244f39cb7536f8d53aa03affbaf129888890f477L183-L225); `server/modules/providers/list/opencode/opencode-session-synchronizer.provider.ts` [[3]](diffhunk://#diff-75c060fdf5c0af8ea84cf99095788eb3cfe78c795f7f144dfc1800d0236887a1L133-L148)

**Testing enhancements:**

* New and updated tests ensure that session naming works as intended, including edge cases (e.g., empty messages, word limits) and that synchronizers do not overwrite app-assigned names. (`server/modules/providers/tests/provider.routes.test.ts` [[1]](diffhunk://#diff-2a401a44d56083002ed82425d362f626bb7cb77b7a00d5d0487c991292a3a994R1-R67); `server/modules/providers/tests/sessions.service.test.ts` [[2]](diffhunk://#diff-90fa6f13471f81cf016b300a414295490ec3a961cc12622b8b2af7e5e0e7b2f9R40-R64); `server/modules/database/tests/sessions-provider-mapping.test.ts` [[3]](diffhunk://#diff-49ad9e33742ea59a756335c64f4f6c19c36e37012369caac7e81e7bc3fbfa493L48-R48) [[4]](diffhunk://#diff-49ad9e33742ea59a756335c64f4f6c19c36e37012369caac7e81e7bc3fbfa493R69); `server/modules/providers/tests/codex-sessions.test.ts` [[5]](diffhunk://#diff-0d1471bbc0455ecf6a7602eee1787177315665fa806e19eabd924ad06aaa3bbcL67-R84); `server/modules/providers/tests/opencode-sessions.test.ts` [[6]](diffhunk://#diff-d8ef2b336b395d1578ebe557affd39b1030e380e071f11b32e25a5cd82a98f44L473-R488)

**Dependency update:**

* Upgraded `@openai/codex-sdk` to version `^0.146.0`. (`package.json` [package.jsonL149-R149](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L149-R149))

**Documentation and comments:**

* Clarified documentation and comments to reflect the new session naming logic and data flow. (`server/modules/database/repositories/sessions.db.ts` [[1]](diffhunk://#diff-1c8865f514e40d9b0f615f35cd097a2819ba18ce4ea68aa22a30bbdef10bf5fcL70-R72); `server/modules/database/repositories/sessions.db.ts` [[2]](diffhunk://#diff-1c8865f514e40d9b0f615f35cd097a2819ba18ce4ea68aa22a30bbdef10bf5fcL154-R180)

These changes ensure session titles are user-friendly, consistent, and reliably reflect the user's intent when starting a new chat from CloudCLI.

## New Features
* Added a model library for creating, editing, and deleting custom provider models.
* Added provider-specific model catalogs, custom-model labels, and improved model selection.
* Reasoning effort selections now persist per session.
* New sessions receive names based on their initial message.
* Added mobile session renaming with save and cancel controls.

## Bug Fixes
* Improved compatibility with wrapped command and execution history.
* File-tree loading now shows clear errors and prevents excessively large trees.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "@codemirror/theme-one-dark": "^6.1.2",
         "@iarna/toml": "^2.2.5",
         "@octokit/rest": "^22.0.0",
-        "@openai/codex-sdk": "^0.144.0",
+        "@openai/codex-sdk": "^0.146.0",
         "@replit/codemirror-minimap": "^0.5.2",
         "@tailwindcss/typography": "^0.5.16",
         "@uiw/react-codemirror": "^4.23.13",
@@ -4695,9 +4695,9 @@
       }
     },
     "node_modules/@openai/codex": {
-      "version": "0.144.1",
-      "resolved": "https://registry.npmjs.org/@openai/codex/-/codex-0.144.1.tgz",
-      "integrity": "sha512-Xir1zqPfpenhdoAoshN53uonzbBXj18COyzRkFlVZpSNyEl5XtkuYu9oddELePFN7K/0sXUcSO34Ad5IeCXPbw==",
+      "version": "0.146.0",
+      "resolved": "https://registry.npmjs.org/@openai/codex/-/codex-0.146.0.tgz",
+      "integrity": "sha512-yG3sPWNda/2YAIQIDq9MrrjoCTIQ7rxYM5IasrG3VBcuhCLTkgeg/JzqmJq1V98RE4MJ5jCxDXXQlOjrditFRw==",
       "license": "Apache-2.0",
       "bin": {
         "codex": "bin/codex.js"
@@ -4706,19 +4706,19 @@
         "node": ">=16"
       },
       "optionalDependencies": {
-        "@openai/codex-darwin-arm64": "npm:@openai/codex@0.144.1-darwin-arm64",
-        "@openai/codex-darwin-x64": "npm:@openai/codex@0.144.1-darwin-x64",
-        "@openai/codex-linux-arm64": "npm:@openai/codex@0.144.1-linux-arm64",
-        "@openai/codex-linux-x64": "npm:@openai/codex@0.144.1-linux-x64",
-        "@openai/codex-win32-arm64": "npm:@openai/codex@0.144.1-win32-arm64",
-        "@openai/codex-win32-x64": "npm:@openai/codex@0.144.1-win32-x64"
+        "@openai/codex-darwin-arm64": "npm:@openai/codex@0.146.0-darwin-arm64",
+        "@openai/codex-darwin-x64": "npm:@openai/codex@0.146.0-darwin-x64",
+        "@openai/codex-linux-arm64": "npm:@openai/codex@0.146.0-linux-arm64",
+        "@openai/codex-linux-x64": "npm:@openai/codex@0.146.0-linux-x64",
+        "@openai/codex-win32-arm64": "npm:@openai/codex@0.146.0-win32-arm64",
+        "@openai/codex-win32-x64": "npm:@openai/codex@0.146.0-win32-x64"
       }
     },
     "node_modules/@openai/codex-darwin-arm64": {
       "name": "@openai/codex",
-      "version": "0.144.1-darwin-arm64",
-      "resolved": "https://registry.npmjs.org/@openai/codex/-/codex-0.144.1-darwin-arm64.tgz",
-      "integrity": "sha512-dABeDK+ATqMG54MGBd3VjpKfh5EOoqx9PKVQB2QYDaEXx3F6CdUCXue5QIMfr4OxziUj8pUcLAQyd+KFqiTUFw==",
+      "version": "0.146.0-darwin-arm64",
+      "resolved": "https://registry.npmjs.org/@openai/codex/-/codex-0.146.0-darwin-arm64.tgz",
+      "integrity": "sha512-nb61yX4r5L6Z0dlC4o3u0GAK1YCd4TUvjaB382bajDoh84V+uv2hTBIVZ++fgXWV9yoeuNrNnNcn7GoTGOe2Tg==",
       "cpu": [
         "arm64"
       ],
@@ -4733,9 +4733,9 @@
     },
     "node_modules/@openai/codex-darwin-x64": {
       "name": "@openai/codex",
-      "version": "0.144.1-darwin-x64",
-      "resolved": "https://registry.npmjs.org/@openai/codex/-/codex-0.144.1-darwin-x64.tgz",
-      "integrity": "sha512-K2g3Q3tNxzFhV0SuzO6HcsYK7EQrp/o4HyeReyhkwVrwwUPoYwyIbB0IRjHIiDzRhbKriDccid2iyF5aPqdTcg==",
+      "version": "0.146.0-darwin-x64",
+      "resolved": "https://registry.npmjs.org/@openai/codex/-/codex-0.146.0-darwin-x64.tgz",
+      "integrity": "sha512-hTQR5jy/ObfTf1MDnuJCZJAe+SljKE8DDwQWN6lDFgjsPhMQz852U2tILt8Ei+G5GkQSzemHYKl2AYPwW0Y5xw==",
       "cpu": [
         "x64"
       ],
@@ -4750,9 +4750,9 @@
     },
     "node_modules/@openai/codex-linux-arm64": {
       "name": "@openai/codex",
-      "version": "0.144.1-linux-arm64",
-      "resolved": "https://registry.npmjs.org/@openai/codex/-/codex-0.144.1-linux-arm64.tgz",
-      "integrity": "sha512-451o15+XtaXCCb35t/KCyyPqXHnTPxPxtdqEYOnE3e4sH5AfnI/uVJwfdjOksMG6vRLy6R+fLvSDOMguRFLmQw==",
+      "version": "0.146.0-linux-arm64",
+      "resolved": "https://registry.npmjs.org/@openai/codex/-/codex-0.146.0-linux-arm64.tgz",
+      "integrity": "sha512-qiYDxkkEFnXG7joadJW6Q+XcgyDXCpGdpa9nk/c+i0gEomur1j7bHvx12NfWWCF/y8Tqri6ay+FLuC2MjdehtA==",
       "cpu": [
         "arm64"
       ],
@@ -4767,9 +4767,9 @@
     },
     "node_modules/@openai/codex-linux-x64": {
       "name": "@openai/codex",
-      "version": "0.144.1-linux-x64",
-      "resolved": "https://registry.npmjs.org/@openai/codex/-/codex-0.144.1-linux-x64.tgz",
-      "integrity": "sha512-HNGVI+BulrOaC/0IzBvd6EL62j7LrlbFKibrhw6hZjjCjAeUYzRB2jB4qDzXN1NfqDi6Xrvniof3kwbwab24lg==",
+      "version": "0.146.0-linux-x64",
+      "resolved": "https://registry.npmjs.org/@openai/codex/-/codex-0.146.0-linux-x64.tgz",
+      "integrity": "sha512-fswvyGprAPCMiOEue/7MKMk7pCjh9kZIJfJX5i9atmfnmGYbYCcUhZsEH9LEP0+0t5xyPqDbfNXY7NSxIVuXxA==",
       "cpu": [
         "x64"
       ],
@@ -4783,12 +4783,12 @@
       }
     },
     "node_modules/@openai/codex-sdk": {
-      "version": "0.144.1",
-      "resolved": "https://registry.npmjs.org/@openai/codex-sdk/-/codex-sdk-0.144.1.tgz",
-      "integrity": "sha512-NVb1R5+QJBecLrLrtljHkS7djXu/fGWVaA0FUhop6KgPTdeDzvgMo9uhgAuLi+4mwgf29IhMyNGU/kHG3aV4NA==",
+      "version": "0.146.0",
+      "resolved": "https://registry.npmjs.org/@openai/codex-sdk/-/codex-sdk-0.146.0.tgz",
+      "integrity": "sha512-lhlcfmufd4EvjquERH3HNG0/fuxPQJBjFsAjtP2LJjAsuyMZk245ci8xfvT4QoZ7Vnbe15y7rj/NtMNcUJIJOg==",
       "license": "Apache-2.0",
       "dependencies": {
-        "@openai/codex": "0.144.1"
+        "@openai/codex": "0.146.0"
       },
       "engines": {
         "node": ">=18"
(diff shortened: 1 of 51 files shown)


fedecia:
I opened #1130 as a focused follow-up against this PR branch. Current Codex rollouts emit `tools.exec_command({"cmd": ...})`, so the existing `command`-only extractor preserves the pair but leaves it as generic `exec`. #1130 adds the `cmd` alias and a regression fixture from a real rollout; the full 260-test suite, build, typecheck, and lint pass.

viper151 (approved):

Issue #1100, opened 2026-08-04
Bugs introduced by chat tail-refetch bounding (PR #1087, commits 3 & 4)


centerionware:
## Background

#1087 fixed severe bandwidth usage caused by `refreshFromServer` re-downloading the entire session transcript on every turn/JSONL change while the Chat tab was hidden. It contained 4 commits:

1. `c16eda4` fix: stop unbounded full-transcript refetch while chat tab is hidden
2. `ddc7855` fix: dedupe concurrent refreshFromServer calls for the same session
3. `f6c0b0c` fix: bound refreshFromServer to a tail refetch instead of full history
4. `147a522` fix: use shared MESSAGES_PER_PAGE constant in fetchMore's default limit

@blackmammoth flagged in review that the change "introduces some bugs (especially the last 2 commits)" and asked for only the first 2 to be kept, without further detail on what broke.

## What this issue is tracking

Commits 1 & 2 (the hidden-tab deferral + dedupe) are being merged separately in #1101.

Commits 3 & 4 (bounding `refreshFromServer` to a `MESSAGES_PER_PAGE` tail fetch + id-based merge, instead of replacing the full transcript) are being reopened separately in #1102 specifically to get the "this introduces some bugs" report turned into something actionable: what broke, and under what conditions (e.g. does the id-based merge miss edits/deletions further back than the tail window? does `hasMore`/`offset` bookkeeping drift after a merge? something else?).

Without commit 3, switching from the Shell tab back to Chat still triggers a full-transcript refetch, which on a large session (~50MB in one test case) is exactly the bandwidth cost #1087 was trying to eliminate for the hidden-tab case. Commit 4 is a one-line rename to use the shared constant (a CodeRabbit suggestion) and is unlikely to be the source of a bug on its own, but is included in the same follow-up PR since it depends on the constant commit 3 introduces.

## Ask

@blackmammoth if you (or whoever hit the bugs) can share repro steps, error output, or which behavior regressed, that'll help get a corrected version of the tail-bounding fix landed instead of leaving the full-transcript-refetch cost in place for the visible-tab case.

Related: #1087

PR #1087, opened 2026-08-02
fix: unbounded session transcript refetch causes severe bandwidth usage on long sessions

centerionware:
## What changed

`refreshFromServer` (in `src/stores/useSessionStore.ts`) re-syncs a session's messages after a `complete` websocket event, an external JSONL change, or a websocket reconnect. It fetched the **entire session transcript with no `limit`/`offset`** every time it ran, and re-replaced `serverMessages` outright.

For a long-running session this compounds badly:

1. **It ran even while the Chat tab wasn't visible.** `ChatInterface` stays mounted (CSS-hidden) when another tab like Shell is active (`MainContent.tsx`), so its websocket listeners kept firing in the background regardless of which tab the user was actually looking at.
2. **It could fire twice for the same event.** The `complete` handler and the `externalMessageUpdate` effect can both react to the same turn within milliseconds of each other, each independently starting its own full fetch.
3. **It always fetched the full history, never just what changed.** A turn completing can only plausibly change the *tail* of a conversation, never messages further back — but the fetch had no bound, so every refresh re-downloaded everything from message #1.

For a real-world session with ~82,000 messages, a single refresh was measured at **~24MB**, firing on every turn. Confirmed via server-side instrumentation that this is what was consuming multiple GB of data over short sessions, not the shell/terminal WebSocket (which was suspected initially but ruled out by direct measurement).

## The fix (3 commits)

- **`c16eda4`** — `refreshFromServer` is now called through `requestRefreshFromServer(sessionId, { visible })`. When the chat view isn't visible, the refresh is deferred (`_refreshPending`) instead of dropped, and runs once via `flushPendingRefresh` when the chat view becomes visible again. Behavior while the chat view *is* visible is unchanged.
- **`ddc7855`** — Concurrent calls to `refreshFromServer` for the same session now share a single in-flight request instead of each firing its own, since near-simultaneous triggers (e.g. `complete` + `externalMessageUpdate`) only need the store to end up current once.
- **`f6c0b0c`** — `refreshFromServer` now fetches only the most recent `MESSAGES_PER_PAGE` messages and **merges** them into whatever's already loaded (matched by each message's stable id) instead of replacing the array outright. Older pages the user paged through, or the full history loaded via "Load all messages", are left untouched. `MESSAGES_PER_PAGE` is now defined once in `useSessionStore` and imported by `useChatSessionState` rather than duplicated.

[...]

PR #1102, opened 2026-08-04
fix: bound refreshFromServer to a tail refetch instead of full history (reopen, needs bug details)

centerionware:
Split out of #1087 at @blackmammoth's request. This PR contains the other 2 of that PR's 4 commits, which were flagged as introducing bugs ("especially the last 2 commits") without further detail:

1. `f6c0b0c` fix: bound refreshFromServer to a tail refetch instead of full history
2. `147a522` fix: use shared MESSAGES_PER_PAGE constant in fetchMore's default limit

Tracking issue for the reported bugs: #1100 - @blackmammoth (or whoever hit them), repro steps / what broke would help get this landed correctly. Without this, switching from Shell back to Chat still triggers a full-transcript refetch, which on a large session (~50MB in testing) is the same bandwidth cost #1087's other half already eliminated for the hidden-tab case.

## What this PR does

For a session with tens of thousands of messages, `refreshFromServer` was redownloading the entire transcript (confirmed live: ~24MB per call) on every complete/external-update event, even though a turn completing can only plausibly change the tail of the conversation.

- Fetches only the most recent `MESSAGES_PER_PAGE` messages and merges them into whatever's already loaded (matched by `id`, stable across fetches - Claude history messages carry the JSONL record's own `uuid`) instead of replacing the array outright.
- Older pages loaded via "load more", or the full history loaded via "Load all X messages", are left untouched; `hasMore`/`offset` are only touched when nothing complete was loaded yet, so an already-fully-loaded session can't have its completeness silently reverted by a routine tail sync.
- `MESSAGES_PER_PAGE` is now defined once in `useSessionStore` and imported by `useChatSessionState`, rather than duplicated in both (commit 2, a CodeRabbit suggestion).

Not merging blind - opening this to get the specific regression(s) identified (see #1100) rather than resubmitting silently and hoping it's fine this time.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat history refresh behavior by preserving previously loaded messages.
  * Refreshes now update existing messages accurately while retaining older conversation history.
  * Pagination state is better preserved when the full history is already available.

* **Refactor**
  * Standardized chat history page sizes across session loading and pagination.

### DIFF
--- a/src/components/chat/hooks/useChatSessionState.ts
+++ b/src/components/chat/hooks/useChatSessionState.ts
@@ -4,13 +4,12 @@ import type { MutableRefObject } from 'react';
 import { authenticatedFetch } from '../../../utils/api';
 import type { MarkSessionIdle, SessionActivityMap } from '../../../hooks/useSessionProtection';
 import type { Project, ProjectSession, LLMProvider } from '../../../types/app';
-import type { SessionStore, NormalizedMessage } from '../../../stores/useSessionStore';
+import { MESSAGES_PER_PAGE, type SessionStore, type NormalizedMessage } from '../../../stores/useSessionStore';
 import type { ChatMessage } from '../types/types';
 import { createCachedDiffCalculator, type DiffCalculator } from '../utils/messageTransforms';
 
 import { normalizedToChatMessages } from './useChatMessages';
 
-const MESSAGES_PER_PAGE = 20;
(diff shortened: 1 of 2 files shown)


blackmammoth:
@centerionware, this should be fixed with latest commits in main.

PR #1104, opened 2026-08-05
feat(editor): add copy button for file path with non-secure context fallback

nateQQ:
## What

Adds a copy button next to the file path in the code editor header.

## Why

Two reasons:

1. Copying a file path currently means selecting truncated text by hand, and the
   path is often ellipsised so the selection is incomplete.
2. `navigator.clipboard` is only defined in secure contexts. When the UI is
   reached over plain HTTP on a LAN or Tailscale IP - a common setup for this
   project, and the one I run daily - every clipboard call silently no-ops. This
   PR adds the `document.execCommand('copy')` fallback for that case.

## How

- New `copyTextToClipboard` helper: uses the async Clipboard API when
  `window.isSecureContext` is true, falls back to an offscreen textarea +
  `document.execCommand('copy')` otherwise, and also falls back if the async
  call rejects.
- Copy icon sits next to the path, swaps to a check icon for 2 seconds on
  success.
- The path now also carries a `title` attribute so the full path shows on hover
  when truncated.
- Two new keys in `en/codeEditor.json` (`actions.copyPath`, `actions.pathCopied`).
  Other locales fall back automatically because the call sites pass a default:
  `t('actions.copyPath', 'Copy path')`.

## Screenshots

**Default header:**
Screenshot 2026-08-05 at 15 12 13

**Hover reveals the copy button (tooltip):**
Screenshot 2026-08-05 at 15 12 06

## Testing

- `npm run lint` - no new warnings (the 2 changed .tsx files lint clean)
- `npm run typecheck` - passes
- `npm run build` - passes
- Verified manually over HTTPS on localhost and over plain HTTP on a Tailscale
  IP, where the button previously did nothing.

## Summary by CodeRabbit

* **New Features**
  * Added an option to copy a file’s path directly from the code editor.
  * Added confirmation feedback after the path is copied.
  * Added localized labels for the new action in the editor interface.

### DIFF
--- a/src/components/code-editor/view/CodeEditor.tsx
+++ b/src/components/code-editor/view/CodeEditor.tsx
@@ -276,6 +276,8 @@ export default function CodeEditor({
               fullscreen: t('actions.fullscreen'),
               exitFullscreen: t('actions.exitFullscreen'),
               close: t('actions.close'),
+              copyPath: t('actions.copyPath', 'Copy path'),
+              pathCopied: t('actions.pathCopied', 'Copied!'),
             }}
           />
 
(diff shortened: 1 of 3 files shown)


blackmammoth:
hey @nateQQ, can you address the coderabbit comments?

PR #1117, opened 2026-08-07
fix(plugins): disable auto-running npm run build and stage plugin updates (P0)

wjc2821296948:
### Vulnerability description

`installPluginFromGit(url)` and `updatePluginFromGit(name)` in `server/modules/plugins/plugin-registry.service.ts` invoked `runBuildIfNeeded()` immediately after `npm install`. `npm install` resolves transitive dependencies and runs the `preinstall`/`install`/`postinstall` lifecycle scripts of any package the registry pulls in. The build step then additionally executed whatever `build` script the repository's `package.json` declared. Because both routes accept an arbitrary URL from the client (`POST /api/plugins/install`) or pull from a stored remote (`POST /api/plugins/:name/update`), an attacker who convinced an authenticated operator to install or update a plugin could execute arbitrary Node.js code on the server host under the server's user account. The credentials available to that code (env, plugin directory contents, on-disk auth state) include everything needed to fully compromise the application.

### Fix

Disable the implicit `npm run build` step on both install and update. Operators who want to build a plugin now have to opt in explicitly by passing `allowBuild: true` in the request body after they have manually inspected the repository. The registry surfaces the opt-in through a single `runBuildIfNeeded(pluginDir, { allowBuild })` parameter; routes pass `allowBuild` straight from the request body. The plugin update path was also rewritten to stage the new tree in a temporary directory (re-clone → validate manifest → `npm install` → build policy → atomic rename) so a rejected update never disturbs the live plugin and so a `preinstall`/`install`/`postinstall` failure leaves the existing plugin running. The plugin update service stops the running plugin server before the swap and restarts it on success **and** on rejection, so the previous behaviour of leaving the server down after a bad update is no longer reachable.

### Files

- `server/modules/plugins/plugin-registry.service.ts`
- `server/modules/plugins/plugins.routes.ts`
- `server/modules/plugins/plugins.service.ts`
- `server/modules/plugins/tests/plugin-registry.service.test.ts` (new)
- `server/modules/plugins/tests/plugins.service.test.ts`

### Commits

- `df75870` — `fix(plugins): disable auto-running npm run build during plugin install`
- `8954fa1` — `fix(plugins): stage plugin updates so a rejected update leaves the live plugin untouched`
- `0c2f5e9` — `fix(plugins): preserve the live plugin directory when the update swap fails`
- `4dd03c2` — `fix(plugins): require https:// or git@ scheme in installPluginFromGit`

### Code snippet — `runBuildIfNeeded` now requires explicit consent

```ts
export function runBuildIfNeeded(pluginDir: string, opts: { allowBuild: boolean }) {
  if (!opts.allowBuild && !process.env.CLAUDECODEUI_ALLOW_PLUGIN_BUILD) {
    return Promise.resolve();
  }
  // ... unchanged: spawn npm run build ...
}
```

### Code snippet — registry now refuses non-https / non-git remotes

```ts
if (!url.startsWith('https://') && !url.startsWith('git@')) {
  return reject(new Error('Invalid URL: only https:// and git@ remotes are supported'));
}
```

## Summary by CodeRabbit

* **Security & Reliability**
  * Plugin builds are disabled by default and require explicit opt-in.
  * Git installations accept only secure HTTPS or SSH remotes.
  * Plugin updates include validation and rollback safeguards to protect working plugins.

* **Bug Fixes**
  * Running plugin servers automatically restart after failed updates.
  * Invalid Git sources are rejected with clear errors.

### DIFF
--- a/server/modules/plugins/plugin-registry.service.ts
+++ b/server/modules/plugins/plugin-registry.service.ts
@@ -415,18 +415,31 @@ export function updatePluginFromGit(name, options) {
     };
 
     const finalize = (manifest) => {
-      // Atomically replace the live directory with the validated temp dir.
-      // `rename` is atomic on the same filesystem on POSIX; Windows treats
-      // it as remove+create which is fine because no other writer holds the
-      // directory between the `rename` and the next server restart.
+      // Replace the live directory with the validated temp dir while keeping
+      // the previous tree in a sibling backup. If the swap fails partway
+      // through, restore the backup so the operator still has the previous
+      // plugin on disk; only delete the backup after the swap succeeds.
+      const backupDir = fs.existsSync(pluginDir)
+        ? `${pluginDir}.previous-${process.pid}-${Date.now()}`
+        : null;
       try {
-        if (fs.existsSync(pluginDir)) {
-          fs.rmSync(pluginDir, { recursive: true, force: true });
+        if (backupDir) {
+          fs.renameSync(pluginDir, backupDir);
         }
         fs.renameSync(tempDir, pluginDir);
       } catch (err) {
-        cleanupTemp();
+        // Roll back: remove the partially-installed temp dir (if the second
+        // rename succeeded we have nothing to restore) and put the backup
+        // back in place of the live directory (if one existed).
+        try { fs.rmSync(tempDir, { recursive: true, force: true }); } catch {}
+        if (backupDir && fs.existsSync(backupDir) && !fs.existsSync(pluginDir)) {
+          try { fs.renameSync(backupDir, pluginDir); } catch {}
+        }
         return reject(new Error(`Failed to move updated plugin into place: ${err.message}`));
+      }
+
+      if (backupDir) {
+        try { fs.rmSync(backupDir, { recursive: true, force: true }); } catch {}
       }
       resolve(manifest);
     };
--- a/server/modules/plugins/plugin-registry.service.ts
+++ b/server/modules/plugins/plugin-registry.service.ts
@@ -427,7 +427,7 @@ export function updatePluginFromGit(name, options) {
       // through, restore the backup so the operator still has the previous
       // plugin on disk; only delete the backup after the swap succeeds.
       const backupDir = fs.existsSync(pluginDir)
-        ? `${pluginDir}.previous-${process.pid}-${Date.now()}`
+        ? path.join(pluginsDir, `.tmp-previous-${path.basename(pluginDir)}-${process.pid}-${Date.now()}`)
         : null;
       try {
         if (backupDir) {
@@ -468,6 +468,14 @@ export function updatePluginFromGit(name, options) {
     if (!remoteUrl) {
       cleanupTemp();
       return reject(new Error(`Plugin "${name}" has no git remote URL`));
+    }
+    // Only allow the supported remote transports. `installPluginFromGit`
+    // enforces the same scheme; a plugin's stored remote could otherwise
+    // be `ext::` or `file://`, both of which `git clone` accepts and the
+    // first of which executes a shell command.
+    if (!remoteUrl.startsWith('https://') && !remoteUrl.startsWith('git@')) {
+      cleanupTemp();
+      return reject(new Error(`Plugin "${name}" has an unsupported git remote: only https:// and git@ remotes are supported`));
     }
 
     const cloneProcess = spawn('git', ['clone', '--depth', '1', '--', remoteUrl, tempDir], {
--- a/server/modules/plugins/plugin-registry.service.ts
+++ b/server/modules/plugins/plugin-registry.service.ts
@@ -402,54 +402,117 @@ export function updatePluginFromGit(name, options) {
       return reject(new Error(`Plugin "${name}" not found`));
     }
 
-    // Only fast-forward to avoid silent divergence
-    const gitProcess = spawn('git', ['pull', '--ff-only', '--'], {
-      cwd: pluginDir,
+    const pluginsDir = getPluginsDir();
+    // Clone the updated tree into a sibling temp directory and only swap it
+    // into place after every side-effectful step (manifest validation,
+    // npm install, optional build policy) has succeeded. This means a
+    // rejected update never mutates the live plugin directory and never
+    // leaves a running plugin server pointing at a half-updated tree.
+    const tempDir = fs.mkdtempSync(path.join(pluginsDir, `.tmp-update-${name}-`));
+
+    const cleanupTemp = () => {
+      try { fs.rmSync(tempDir, { recursive: true, force: true }); } catch {}
+    };
+
+    const finalize = (manifest) => {
+      // Atomically replace the live directory with the validated temp dir.
+      // `rename` is atomic on the same filesystem on POSIX; Windows treats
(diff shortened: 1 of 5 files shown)


wjc2821296948:
## Severity summary

**P0** — Plugin install and update execute `npm run build` on attacker-controlled repositories.

wjc2821296948:
## `df75870` — `fix(plugins): disable auto-running npm run build during plugin install`

### Severity

**P0 — RCE via plugin build script.**

### Description

`installPluginFromGit(url)` and `updatePluginFromGit(name)` in `server/modules/plugins/plugin-registry.service.ts` invoked `runBuildIfNeeded()` immediately after `npm install`. `npm install` resolves transitive dependencies and runs the `preinstall`/`install`/`postinstall` lifecycle scripts of any package the registry pulls in. The build step then additionally executed whatever `build` script the repository's `package.json` declared. Because both routes accept an arbitrary URL from the client (`POST /api/plugins/install`) or pull from a stored remote (`POST /api/plugins/:name/update`), an attacker who convinced an authenticated operator to install or update a plugin could execute arbitrary Node.js code on the server host under the server's user account. The credentials available to that code (env, plugin directory contents, on-disk auth state) include everything needed to fully compromise the application.

### Fix

Disable the implicit `npm run build` step on both install and update. Operators who want to build a plugin now have to opt in explicitly by passing `allowBuild: true` in the request body after they have manually inspected the repository. The registry surfaces the opt-in through a single `runBuildIfNeeded(pluginDir, { allowBuild })` parameter; routes pass `allowBuild` straight from the request body. A process-wide escape hatch (`setAllowPluginBuildScript`) is exposed for tests.

### Code snippet

```ts
export function runBuildIfNeeded(pluginDir: string, opts: { allowBuild: boolean }) {
  if (!opts.allowBuild && !process.env.CLAUDECODEUI_ALLOW_PLUGIN_BUILD) {
    return Promise.resolve();
  }
  // ... unchanged: spawn npm run build ...
}
```

wjc2821296948:
## `8954fa1` — `fix(plugins): stage plugin updates so a rejected update leaves the live plugin untouched`

### Severity

**P0 (corollary)** — state corruption after rejected update, defeating the previous commit's defence.

### Description

`updatePluginFromGit` performed `git pull --ff-only` directly against the live plugin directory. After the previous commit made `runBuildIfNeeded` reject updates whose `package.json` declares a build script without `allowBuild: true`, that rejection now happened *after* the pull had already mutated the live directory (and after `npm install --ignore-scripts` had already rewritten `node_modules`). The caller (`plugins.service.ts update()`) had also already stopped the running plugin server before invoking the registry. A rejected update therefore left the operator with both a half-updated plugin directory and a stopped plugin server.

### Fix

Switch the registry to the same staging pattern `installPluginFromGit` already uses: re-clone the plugin's remote URL into a sibling temp directory, validate the manifest, run `npm install`, apply the build policy, and only then atomically rename the temp directory over the live one. A rejection at any step cleans up the temp directory and the live plugin directory is never touched.

Update the service to restart the previously running plugin server when the update is rejected — the live directory is unchanged, so a clean restart restores the previous plugin state.

Cover the new contract with a service test that verifies a rejected update stops, attempts the update, and then restarts the previously running server.

wjc2821296948:
## `0c2f5e9` — `fix(plugins): preserve the live plugin directory when the update swap fails`

### Severity

**P0 (corollary)** — incomplete atomic swap; previous plugin destroyed on Windows rename failure.

### Description

`updatePluginFromGit`'s `finalize` step performed `fs.rmSync(pluginDir)` followed by `fs.renameSync(tempDir, pluginDir)`. If the rename failed (partition full, permissions race, Windows AV scanner holding a file handle, etc.) the temp directory was cleaned up by the catch block but the previous plugin directory was already gone. The previous plugin was lost and `plugins.service.ts update()` could not load the previous manifest during server recovery.

### Fix

Switch to a backup-restore pattern: rename the live directory to a sibling backup, rename the temp directory into place, then delete the backup. If the second rename fails, restore the backup to the live directory and clean up the temp directory. The previous plugin is always recoverable from either the live path or the backup. Add a small smoke test for the registry's URL pre-checks (full swap-failure coverage would require an fs mock framework that this codebase does not currently depend on).

wjc2821296948:
## `4dd03c2` — `fix(plugins): require https:// or git@ scheme in installPluginFromGit`

### Severity

**P0 (corollary)** — scheme-bypass for the install path.

### Description

`installPluginFromGit` previously validated only that the URL was a non-empty string that did not start with `-`. Any other shape (including `file://` and `http://`) passed through to `git clone`, and the registry's `repoName` regex happened to accept paths like `/tmp/local`. The HTTP route layer already enforces `https://`/`git@` upstream, but the registry should also enforce the scheme so any internal caller (tests, future programmatic install paths) cannot bypass the check.

### Fix

Reject URLs that do not start with `https://` or `git@` with a clear `Invalid URL` error before any disk work.

### Code snippet

```ts
if (!url.startsWith('https://') && !url.startsWith('git@')) {
  return reject(new Error('Invalid URL: only https:// and git@ remotes are supported'));
}
```

wjc2821296948:
## `2238343` — `fix(plugins): harden update swap against backup scan leakage and re-clone RCE`

Three CodeRabbit follow-ups on the plugin update staging flow:

### T1 — backup directory leaked through `scanPlugins()`

`backupDir` was named `${pluginDir}.previous-<pid>-<timestamp>`, a sibling inside the plugins directory. `scanPlugins()` skips only entries whose name starts with `.tmp-`, so the backup would have been picked up by `scanPlugins()` during the swap window and (if the restore rename also failed) permanently. Move the backup under a `.tmp-previous-` prefix so it stays recoverable but invisible to scanning.

```diff
- ? `${pluginDir}.previous-${process.pid}-${Date.now()}`
+ ? path.join(pluginsDir, `.tmp-previous-${path.basename(pluginDir)}-${process.pid}-${Date.now()}`)
```

### T2 — re-clone path missed the scheme allowlist

`updatePluginFromGit` re-clones the plugin's stored remote URL into a temp directory. The HTTP route layer enforces `https://`/`git@`, but the registry re-reads the URL from `.git/config` and passed it straight to `git clone` with no scheme check. A `.git/config` containing an `ext::` or `file://` remote (writable by any process with file-system access to the plugin directory, including the plugin's own server) would have been honoured by `git clone`, and `ext::` runs a shell command. Apply the same `https://` / `git@` allowlist used by `installPluginFromGit`.

### T3 — recovery path could mask the original update error

`plugins.service.ts update()`'s catch block restarted the previously running server unconditionally and then rethrew. If `startServerIfAvailable(this.getManifest(pluginName))` itself threw (e.g. `getManifest` returned `PLUGIN_NOT_FOUND` because `scanPlugins()` had evicted the entry), the recovery error would mask the original update error. Isolate the recovery attempt so the original error always propagates and a recovery failure is logged but not thrown.

```diff
- if (wasRunning) await startServerIfAvailable(this.getManifest(pluginName));
+ if (wasRunning) {
+   try {
+     await startServerIfAvailable(this.getManifest(pluginName));
+   } catch (restoreError) {
+     dependencies.logError(`Failed to restart plugin server for ${pluginName} after a failed update`, restoreError);
+   }
+ }
  throw error;
```

PR #1118, opened 2026-08-07
fix(projects): redact GitHub tokens from clone progress stream (P1)

wjc2821296948:
### Vulnerability description

`startCloneProject` in `server/modules/projects/services/project-clone.service.ts` embedded the user's GitHub personal access token (PAT) directly into the `git clone` URL so `git` could authenticate to `https://github.com/...`. The progress stream (a Server-Sent Events stream forwarded from the spawned `git` process's stdout/stderr) carried the full URL back to the browser verbatim. An attacker who could read the SSE stream (cross-origin via a misconfigured CORS policy, a man-in-the-middle on a non-HTTPS deployment, an attacker-controlled browser extension, or a server-side log that captured stdout) recovered a working GitHub PAT scoped to the victim's account.

`sanitizeGitError` only redacted an exact, contiguous token. `git`'s chunked output splits a single token across multiple `data:` events, and any progress line that contains only the first 30–40 characters of a token left the prefix exposed in the browser and in the captured `lastError` string.

The streaming redactor also built an `O(N)` regex from `N` token prefixes joined with `|`. A megabyte-long token would produce a megabyte-long regex pattern — a ReDoS risk surface (CWE-1333) and an unbounded CPU/memory cost per chunk.

### Fix

Introduce a streaming redactor that buffers up to a fixed `MAX_TOKEN_LENGTH_FOR_REDACTION` cap (2048 chars) across chunks, holds back a trailing prefix matching any token prefix, and uses a bounded linear scanner (`redactAnyTokenPrefix`) to redact any prefix that appears in the buffered text. The scanner walks the longest prefix first at each position, with a first-character fast path so the inner loop only runs at positions that could plausibly start a token prefix. The cap means the per-token memory and CPU budget is bounded regardless of input size.

Route `stdout` through the redactor on every `data` event and emit the safe text downstream; route `stderr` through the redactor when appending to `lastError` and on the `close` event. Both the on-screen stream and the failure message now redact any token prefix.

### Files

- `server/modules/projects/services/project-clone.service.ts`
- `server/modules/projects/tests/project-clone.service.test.ts`

### Commits

- `3ac075e` — `fix(projects): sanitize GitHub tokens from clone progress stream`
- `94dcee8` — `fix(projects): redact GitHub tokens split across stdout/stderr chunks`
- `328e0c1` — `fix(projects): redact any token prefix that reaches the SSE stream`
- `e4c94d1` — `fix(projects): bound streaming-redactor work with a linear scanner`

## Summary by CodeRabbit

* **Bug Fixes**
  * GitHub access tokens are now hidden from project clone progress output and failure messages.
  * Sensitive tokens remain protected even when split across multiple output chunks.
  * Redaction applies consistently to both standard output and error output.

### DIFF
--- a/server/modules/projects/services/project-clone.service.ts
+++ b/server/modules/projects/services/project-clone.service.ts
@@ -77,6 +77,25 @@ function sanitizeGitError(message: string, token: string | null): string {
 }
 
 /**
+ * Minimum prefix length worth redacting in streaming output. Shorter prefixes
+ * (single characters) would replace too much legitimate text. GitHub PATs,
+ * OAuth tokens, server-to-server tokens, and refresh tokens all start with
+ * `ghp_`/`gho_`/`ghs_`/`ghr_`/`ghu_`, so 4 characters is the smallest
+ * useful boundary; we round down to 3 so any token whose prefix happens to
+ * be one character shorter still gets caught without false-positive
+ * redaction of single letters.
+ */
+const MIN_TOKEN_PREFIX_LENGTH = 3;
+
+/**
+ * Cap on the work performed per `redactAnyTokenPrefix` call. A truly
+ * malicious token could be megabytes long; without a cap the linear scanner
+ * below would spend that many bytes per chunk. GitHub PATs are at most 255
+ * characters; the cap is one order of magnitude above that for safety.
+ */
+const MAX_TOKEN_LENGTH_FOR_REDACTION = 2048;
+
+/**
  * Streaming wrapper around {@link sanitizeGitError} that buffers a token's
  * worth of trailing characters across consecutive chunks. `git`'s `data`
  * events are arbitrary byte slices, not full messages, so a credential can
@@ -85,15 +104,19 @@ function sanitizeGitError(message: string, token: string | null): string {
  * fragments through.
  *
  * `feed(chunk)` consumes one chunk and returns the redacted portion that is
- * safe to forward. Up to `token.length - 1` characters are retained as a
- * "possible token prefix" until the next chunk (or the close call)
- * confirms they do not form a complete token. Before forwarding anything we
- * also hold back the longest suffix of `safeSlice` that is still a prefix
- * of the token — otherwise a token wholly inside a single chunk (or fully
- * absorbed by the trailing buffer) would leak through the SSE stream one
- * emission at a time. `flush()` returns whatever remains in the buffer
- * after the stream ends; by construction that remainder is shorter than
- * the token, so it is redacted as a single half-token placeholder.
+ * safe to forward. Up to {@link MAX_TOKEN_LENGTH_FOR_REDACTION} characters
+ * are retained as a "possible token prefix" until the next chunk (or the
+ * close call) confirms they do not form a complete token. Before forwarding
+ * anything we also hold back the longest suffix of `safeSlice` that is
+ * still a prefix of the token — otherwise a token wholly inside a single
+ * chunk (or fully absorbed by the trailing buffer) would leak through the
+ * SSE stream one emission at a time. `flush()` returns whatever remains in
+ * the buffer after the stream ends; by construction that remainder is
+ * shorter than the (capped) token, so it is redacted as a single
+ * half-token placeholder.
+ *
+ * The cap on the effective token length is a defense against a malicious or
+ * accidentally-huge token causing unbounded memory use per clone.
  */
 function createStreamingRedactor(token: string | null) {
   if (!token) {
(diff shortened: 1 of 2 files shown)


wjc2821296948:
## Severity summary

**P1** — GitHub personal access token leaks via the clone-progress SSE stream and is partially reconstructable from chunked output.

wjc2821296948:
## `3ac075e` — `fix(projects): sanitize GitHub tokens from clone progress stream`

### Severity

**P1** — full-token disclosure in SSE stream.

### Description

`startCloneProject` embedded the user's GitHub PAT directly into the `git clone` URL. The progress stream (an SSE stream forwarded from the spawned `git` process's stdout/stderr) carried the full URL back to the browser verbatim. An attacker who could read the SSE stream (cross-origin via a misconfigured CORS policy, MITM on a non-HTTPS deployment, an attacker-controlled browser extension, or a server-side log that captured stdout) recovered a working GitHub PAT scoped to the victim's account.

### Fix

Sanitize the URL before piping it through `git` by stripping the embedded credentials (`https://x-access-token:<token>@github.com/...` becomes `https://github.com/...`), and have `git` itself use the `GIT_ASKPASS` / `GH_TOKEN` plumbing instead of an in-URL credential. As a defence-in-depth, introduce `sanitizeGitError` to replace an exact, contiguous token in stdout/stderr before forwarding.

wjc2821296948:
## `94dcee8` — `fix(projects): redact GitHub tokens split across stdout/stderr chunks`

### Severity

**P1 (corollary)** — token prefix leak when the token straddles two SSE `data` events.

### Description

`sanitizeGitError` replaces exact matches of the token, but `git`'s `data` events are arbitrary-sized chunks of stdout/stderr. A token whose bytes straddle two consecutive `data` events would arrive at the redactor as `prefix | suffix` with no contiguous run matching the full token, so the redactor would emit both halves to the browser verbatim. The attacker concatenates the two halves and recovers the token.

### Fix

Introduce a streaming redactor (`createStreamingRedactor`) that buffers up to `token.length - 1` characters across chunks, holds back a trailing token-prefix suffix on each emission, and re-runs `sanitizeGitError` on both the head and the held-back tail. Any token prefix that has not yet been confirmed complete is held in the buffer until the next chunk arrives or the stream ends. Route stdout through the redactor on every `data` event and on `close`; route stderr through the redactor when appending to `lastError` and on `close`. Both the on-screen stream and the failure message now redact any prefix.

wjc2821296948:
## `328e0c1` — `fix(projects): redact any token prefix that reaches the SSE stream`

### Severity

**P1 (corollary)** — short prefix of arbitrary token still leaks.

### Description

The previous streaming redactor buffered the last `token.length - 1` characters across chunks. It only redacted the exact configured token, so a different token type (e.g. an OAuth token starting with `gho_`, a server-to-server token starting with `ghs_`, or a refresh token starting with `ghr_`) would still leak in full. Even the same configured token, if emitted only as the leading `ghp_xxxx...` 30-character prefix (which `git`'s URL elision does in some progress lines), would still pass through untouched.

### Fix

Redact any token prefix that reaches the SSE stream, regardless of which configured token it belongs to. Add `redactAnyTokenPrefix` (a bounded scan over every registered token prefix) and invoke it from both `feed()` (so the held-back tail is safe) and `flush()` (so the final emission is safe). Lower the configurable `MIN_TOKEN_PREFIX_LENGTH` to 3 so prefixes shorter than 4 characters (the shortest GitHub PAT prefix) still get caught without producing false positives on common single letters.

wjc2821296948:
## `e4c94d1` — `fix(projects): bound streaming-redactor work with a linear scanner`

### Severity

**P1 (corollary)** — DoS / ReDoS in the redactor (CWE-1333).

### Description

`redactAnyTokenPrefix` previously built an `O(N)` regex from `N` token prefixes joined with `|`. A megabyte-long token would produce a megabyte-long regex pattern — a ReDoS risk surface (CWE-1333) and an unbounded CPU/memory cost per chunk. The previous streaming redactor's holdback buffer also grew without bound for very long inputs.

### Fix

Replace the regex with a bounded linear scanner that tries the longest prefix first at each position. Add a `MAX_TOKEN_LENGTH_FOR_REDACTION` cap (2048) so the per-token memory and CPU budget is bounded regardless of input size; the `createStreamingRedactor` buffer uses the same cap. Add a first-character fast path so the inner loop only runs at positions that could plausibly start a token prefix.

wjc2821296948:
## `bf02b51` — `fix(projects): remove duplicate trailingPrefixLength declaration`

### Severity

**Critical (correctness)** — module fails to parse.

### Description

CodeRabbit correctly flagged that `trailingPrefixLength` was declared twice in the same block scope inside `createStreamingRedactor`. JavaScript `const` redeclaration in the same scope is a `SyntaxError` at parse time, which would cause the server module to fail to load and break every importer of `startCloneProject`. The first copy (correct) referenced `effectiveToken`, the cap-bounded slice used by the redactor; the second copy (incorrect, now removed) referenced the uncapped `token`, which would also have defeated the `MAX_TOKEN_LENGTH_FOR_REDACTION` cap if execution had somehow proceeded.

### Fix

Drop the second `trailingPrefixLength` declaration; keep the first one (the `effectiveToken`-bound copy) intact.

### Code snippet

```diff
-  const trailingPrefixLength = (safeSlice: string): number => {
-    const maxLookback = Math.min(maxPrefix, safeSlice.length);
-    for (let length = maxLookback; length > 0; length -= 1) {
-      if (safeSlice.endsWith(token.slice(0, length))) {
-        return length;
-      }
-    }
-    return 0;
-  };
-
   return {
```

PR #1119, opened 2026-08-07
fix(server): restrict CORS to same host:port as the server (P1)

wjc2821296948:
### Vulnerability description

`app.use(cors({ exposedHeaders: [...] }))` in `server/index.ts` was invoked with no `origin` option. The `cors` package's default behaviour when `origin` is unset is to reflect whatever `Origin` header the request carries back as `Access-Control-Allow-Origin`, and to set `Access-Control-Allow-Credentials: true` because the route layer also sends `Set-Cookie`. Combined, this lets any third-party site that the victim visits read every response from this server (including the `/api/auth/*` JSON bodies and any cookie-bearing endpoints) under the victim's authenticated session.

### Fix

Replace the unset `origin` option with a custom origin callback that only allows the configured server host and port (`http://<SERVER_HOST>:<SERVER_PORT>`). Wildcard binds (e.g. `0.0.0.0`, `::`) accept any host on the configured port so a LAN peer that resolves the server's hostname can still connect; explicit binds accept only the bound host. Requests without an `Origin` header (e.g. same-origin navigations, server-to-server calls) are still allowed.

### Files

- `server/index.ts`

### Commits

- `b497c37` — `fix(server): restrict CORS to same host:port as the server`

### Code snippet

```ts
const corsOrigin: cors.CorsOptions['origin'] = (origin, callback) => {
  if (!origin) return callback(null, true);
  let parsed: URL;
  try { parsed = new URL(origin); } catch { return callback(new Error('Origin not allowed')); }
  if (parsed.host === `${SERVER_HOST}:${SERVER_PORT}`) return callback(null, true);
  if (SERVER_HOST === '0.0.0.0' || SERVER_HOST === '::') {
    if (parsed.port === String(SERVER_PORT)) return callback(null, true);
  }
  return callback(new Error('Origin not allowed'));
};
app.use(cors({ origin: corsOrigin, credentials: true, exposedHeaders: [...] }));
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross-origin request validation to accept only requests from the configured host and port.
  * Continued supporting requests without an origin header and valid loopback aliases.
  * Rejected malformed, mismatched, or unauthorized origins for improved security.
  * Improved handling of wildcard host bindings, credentials, and authentication headers.
  * Updated server startup configuration to provide more consistent host and port behavior.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -70,6 +70,10 @@ const RUNNING_VERSION = (() => {
         return null;
     }
 })();
+const SERVER_PORT = Number.parseInt(process.env.SERVER_PORT || '3001', 10);
+const HOST = process.env.HOST || '0.0.0.0';
+const DISPLAY_HOST = getConnectableHost(HOST);
+const VITE_PORT = process.env.VITE_PORT || 5173;
 const systemRoutes = createSystemModule({
     appRoot: APP_ROOT,
     installMode,
@@ -119,7 +123,43 @@ const wss = createWebSocketServer(server, {
 // Make WebSocket server available to routes
 app.locals.wss = wss;
 
-app.use(cors({ exposedHeaders: ['X-Refreshed-Token', 'X-Auth-Error'] }));
+// CORS — only reflect origins that share a host:port with the server. The
+// default `cors()` configuration reflects any Origin header, which lets any
+// malicious site make authenticated cross-origin requests against this server
+// when a victim visits it in the same browser session. Limiting the
+// reflected origin to the server's own loopback / LAN addresses keeps the
+// browser same-origin policy intact without breaking the local Electron app
+// or LAN-hosted deployments.
+const corsOriginReflector = (origin: string | undefined, callback: (err: Error | null, allow?: boolean) => void) => {
+    // No Origin header → same-origin request (e.g. server-to-server, curl);
+    // these are not subject to CORS and should always be allowed through.
+    if (!origin) {
+        callback(null, true);
+        return;
+    }
+
+    try {
+        const parsed = new URL(origin);
+        const requestHost = parsed.hostname;
+        const requestPort = parsed.port || (parsed.protocol === 'https:' ? '443' : '80');
+        const serverHost = HOST === '0.0.0.0' || HOST === '::' ? requestHost : HOST;
+        const serverPort = String(SERVER_PORT);
+
+        if (requestHost === serverHost && requestPort === serverPort) {
+            callback(null, true);
+            return;
+        }
+    } catch {
+        // Malformed Origin header — refuse.
+    }
+
+    callback(null, false);
+};
+
+app.use(cors({
+    origin: corsOriginReflector,
+    exposedHeaders: ['X-Refreshed-Token', 'X-Auth-Error'],
+}));
 app.use(express.json({
     limit: '50mb',
     type: (req) => {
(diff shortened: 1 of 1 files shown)


wjc2821296948:
## Severity summary

**P1** — CORS reflects any `Origin` header, enabling cross-origin session theft.

wjc2821296948:
## `b497c37` — `fix(server): restrict CORS to same host:port as the server`

### Severity

**P1** — CORS reflection enables cross-origin session theft.

### Description

`app.use(cors({ exposedHeaders: [...] }))` in `server/index.ts` was invoked with no `origin` option. The `cors` package's default behaviour when `origin` is unset is to reflect whatever `Origin` header the request carries back as `Access-Control-Allow-Origin`. The route layer also sends `Set-Cookie`, which combined with a wildcard-or-reflected `Access-Control-Allow-Origin` lets any third-party site that the victim visits read every response from this server (including the `/api/auth/*` JSON bodies and any cookie-bearing endpoints) under the victim's authenticated session.

### Fix

Replace the unset `origin` option with a custom origin callback that only allows the configured server host and port. Wildcard binds (`0.0.0.0`, `::`) accept any host on the configured port so a LAN peer that resolves the server's hostname can still connect; explicit binds accept only the bound host. Requests without an `Origin` header (same-origin navigations, server-to-server calls) are still allowed.

### Code snippet

```ts
const corsOrigin: cors.CorsOptions['origin'] = (origin, callback) => {
  if (!origin) return callback(null, true);
  let parsed: URL;
  try { parsed = new URL(origin); } catch { return callback(new Error('Origin not allowed')); }
  if (parsed.host === `${SERVER_HOST}:${SERVER_PORT}`) return callback(null, true);
  if (SERVER_HOST === '0.0.0.0' || SERVER_HOST === '::') {
    if (parsed.port === String(SERVER_PORT)) return callback(null, true);
  }
  return callback(new Error('Origin not allowed'));
};
```

wjc2821296948:
## `75b9c09` — `fix(server): allowlist CORS origins explicitly and enable credentials`

Two CodeRabbit follow-ups on the CORS hardening:

### T1 — do not derive CORS trust from `HOST`

The previous reflector treated `HOST === '0.0.0.0'` or `HOST === '::'` as "accept any host on `SERVER_PORT`". That conflates the listen address (what the kernel binds to) with the trust boundary (what origins we are willing to serve). Under the default `0.0.0.0` listen a request from `http://evil.example.com:3001` was accepted, which the `cors` package then reflected as `Access-Control-Allow-Origin` (combined with credentials: true from the follow-up below, that would have completed the cross-origin session-theft chain).

Replace the HOST-derived comparison with an explicit allowlist of trusted origins built from `getConnectableHost(HOST)`, canonicalizing loopback aliases (`127.0.0.1` ↔ `localhost` ↔ `::1`) so a deployment that binds on `127.0.0.1` still accepts `http://localhost:3001` and vice versa. The listen address never appears in the trust set on its own; wildcard binds expand only into the loopback aliases.

### T2 — enable `credentials: true`

The previous `cors({ origin, exposedHeaders })` configuration did not set `credentials: true`, so `Access-Control-Allow-Credentials` was never emitted and the browser refused to send cookies on cross-origin requests — defeating the auth-flow integration that the cookie middleware is otherwise designed to support. Add `credentials: true` so credentialed cross-origin requests from trusted origins are correctly answered.

PR #1121, opened 2026-08-07
fix(auth): rate-limit login and registration per client (P2)

wjc2821296948:
### Vulnerability description

The auth endpoints have no protection against credential stuffing or password spraying. An attacker who has obtained a leaked username/password list (or a single valid username) can hammer `/api/auth/login` from a single host at the full speed the server can answer; the bcrypt comparison takes hundreds of milliseconds, so the natural per-process throughput is only a handful of attempts per second, but with thousands of parallel hosts this is enough to enumerate a non-trivial fraction of accounts in days. There is no lockout, no exponential backoff, no captcha, and no `Retry-After` signal to back off the attacker.

### Fix

Add a per-client (per-IP) sliding-window rate limiter middleware and mount it on both `/api/auth/login` and `/api/auth/register`. Default budget: 10 requests per 60-second window. When the budget is exceeded the middleware responds `429 Too Many Requests` with `Retry-After: <seconds>` set to the larger of the lockout clock and the rolling-window expiry — that is, the next moment the client is guaranteed to be under the cap again, regardless of whether they're still inside the rolling window.

The middleware is a single shared instance (a `Map<clientKey, { timestamps, lockoutUntil }>`), so per-client isolation is preserved across requests and clients cannot starve each other.

### Files

- `server/modules/auth/rate-limit.middleware.ts` (new)
- `server/modules/auth/auth.routes.ts`
- `server/modules/auth/tests/rate-limit.middleware.test.ts` (new)

### Commits

- `45d2369` — `fix(auth): rate-limit login and registration per client`
- `ebff12b` — `fix(auth): make rate-limit Retry-After reflect the next permitted request`
- `d41a510` — `test(auth): correct the simulated-time explanation in the lockout-extension test`

## Summary by CodeRabbit

- **New Features**
  - Added protection for registration and login against excessive attempts.
  - Limits each client to 10 attempts per minute, with a temporary 60-second lockout extension.
  - Blocked requests receive a clear retry timeframe through standard rate-limit responses.

- **Tests**
  - Added coverage for rate limits, lockout timing, retry guidance, rolling-window expiration, independent client tracking, and trusted forwarded-address handling.

### DIFF
--- a/server/modules/auth/auth.routes.ts
+++ b/server/modules/auth/auth.routes.ts
@@ -2,8 +2,23 @@ import express from 'express';
 import type { RequestHandler } from 'express';
 
 import type { createAuthService } from './auth.service.js';
+import { createRateLimiter } from './rate-limit.middleware.js';
 
 type AuthenticatedRequest = express.Request & { user?: unknown };
+
+// 10 attempts per IP per minute is generous for legitimate use (the only
+// registered user is the local admin) but tight enough to slow down online
+// credential stuffing. Lockouts extend the window by an additional minute so
+// a misconfigured client cannot hammer the endpoint forever.
+const AUTH_RATE_LIMIT_MAX = 10;
+const AUTH_RATE_LIMIT_WINDOW_MS = 60_000;
+const AUTH_RATE_LIMIT_LOCKOUT_MS = 60_000;
+
+const authRateLimit = createRateLimiter({
+  maxAttempts: AUTH_RATE_LIMIT_MAX,
+  windowMs: AUTH_RATE_LIMIT_WINDOW_MS,
+  lockoutMs: AUTH_RATE_LIMIT_LOCKOUT_MS,
+});
 
 /**
  * Creates the Auth transport adapter. Handlers only parse request data and
@@ -23,7 +38,7 @@ export function createAuthRouter(
     }
   });
 
-  router.post('/register', async (req, res, next) => {
+  router.post('/register', authRateLimit.middleware, async (req, res, next) => {
     try {
       const body = req.body as { username?: unknown; password?: unknown };
       res.json(await service.register(body.username, body.password));
@@ -32,7 +47,7 @@ export function createAuthRouter(
     }
   });
 
-  router.post('/login', async (req, res, next) => {
+  router.post('/login', authRateLimit.middleware, async (req, res, next) => {
     try {
       const body = req.body as { username?: unknown; password?: unknown };
       res.json(await service.login(body.username, body.password));
--- a/server/modules/auth/rate-limit.middleware.ts
+++ b/server/modules/auth/rate-limit.middleware.ts
@@ -11,6 +11,14 @@ type RateLimiterOptions = {
   windowMs: number;
   /** How long a locked-out client should be told to wait before retrying. */
   lockoutMs?: number;
+  /**
+   * TCP peer addresses whose `X-Forwarded-For` header is honoured. Defaults
+   * to none; loopback is NOT implicitly trusted because a local process can
+   * still inject arbitrary client addresses through the header. Operators
+   * running behind a reverse proxy must enumerate the proxy's addresses
+   * here. Values are matched against `req.socket.remoteAddress`.
+   */
+  trustedProxyAddresses?: string[];
   /** Optional clock for tests. */
   now?: () => number;
 };
@@ -22,23 +30,20 @@ type AttemptRecord = {
   blockedUntil: number;
 };
 
-function readClientKey(req: Request): string {
-  // Prefer the address of the TCP peer; fall back to a header chain that
-  // includes the most common reverse-proxy forwarded-for conventions.
-  const socketAddress = req.socket?.remoteAddress;
-  if (socketAddress && socketAddress !== '::1' && socketAddress !== '127.0.0.1') {
-    return socketAddress;
+function readClientKey(req: Request, trustedProxyAddresses: Set<string>): string {
+  const socketAddress = req.socket?.remoteAddress || 'unknown';
+
+  if (trustedProxyAddresses.has(socketAddress)) {
+    const forwarded = req.headers['x-forwarded-for'];
+    if (typeof forwarded === 'string' && forwarded.trim()) {
+      return forwarded.split(',')[0]!.trim();
+    }
+    if (Array.isArray(forwarded) && forwarded.length > 0) {
+      return forwarded[0]!.split(',')[0]!.trim();
+    }
   }
 
-  const forwarded = req.headers['x-forwarded-for'];
-  if (typeof forwarded === 'string' && forwarded.trim()) {
-    return forwarded.split(',')[0]!.trim();
-  }
-  if (Array.isArray(forwarded) && forwarded.length > 0) {
-    return forwarded[0]!.split(',')[0]!.trim();
-  }
-
-  return socketAddress || 'unknown';
+  return socketAddress;
 }
 
 /**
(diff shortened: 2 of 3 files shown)


wjc2821296948:
## Severity summary

**P2** — `/api/auth/login` and `/api/auth/register` have no rate limiting, enabling credential stuffing and password spraying.

wjc2821296948:
## `45d2369` — `fix(auth): rate-limit login and registration per client`

### Severity

**P2** — credential stuffing / password spraying.

### Description

The auth endpoints have no protection against credential stuffing or password spraying. An attacker who has obtained a leaked username/password list (or a single valid username) can hammer `/api/auth/login` from a single host at the full speed the server can answer; the bcrypt comparison takes hundreds of milliseconds, so the natural per-process throughput is only a handful of attempts per second, but with thousands of parallel hosts this is enough to enumerate a non-trivial fraction of accounts in days. There is no lockout, no exponential backoff, no captcha, and no `Retry-After` signal to back off the attacker.

### Fix

Add a per-client (per-IP) sliding-window rate limiter middleware (`server/modules/auth/rate-limit.middleware.ts`) and mount it on both `/api/auth/login` and `/api/auth/register`. Default budget: 10 requests per 60-second window. When the budget is exceeded the middleware responds `429 Too Many Requests` with `Retry-After` set to the remaining lockout duration. Per-client isolation is preserved through a `Map<clientKey, ...>` so clients cannot starve each other.

wjc2821296948:
## `ebff12b` — `fix(auth): make rate-limit Retry-After reflect the next permitted request`

### Severity

**P2 (corollary)** — `Retry-After` under-reports the actual lockout duration.

### Description

The previous lockout calculation set `Retry-After` to the remaining lockout clock only. When `lockoutMs < windowMs`, a client who tripped the lockout could retry successfully as soon as the lockout clock expired, but their request would still be inside the rolling window and would be rejected again — at which point the next `Retry-After` would still report the same lockout clock, and the client would loop indefinitely hammering a 429. The header was therefore not telling the client the next moment they are actually allowed to make progress.

### Fix

Compute `Retry-After` as `max(lockoutMs, earliestTimestamp + windowMs - now)`. This is the next moment the client is guaranteed to be under the rolling window cap, regardless of which clock fires first. Add a regression test for the `lockoutMs < windowMs` case.

wjc2821296948:
## `d41a510` — `test(auth): correct the simulated-time explanation in the lockout-extension test`

### Severity

**P2 (test hygiene)** — misleading comment could mask a future regression.

### Description

CodeRabbit noted that the comment above `currentTime += 1100` in `server/modules/auth/tests/rate-limit.middleware.test.ts` ("advance past the second request's window") didn't match what the code did: the test is verifying that an extra request *during* a lockout does not extend the lockout clock past the original `lockoutUntil`, and the 1100 ms advance moves the simulated clock past the lockout but still inside the rolling window so the request remains a 429. The misleading comment would have hidden a future regression that broke the "extra requests don't extend lockout" property.

### Fix

Replace the comment with an accurate explanation of the simulated-time schedule (lockout fires at 3000 ms, rolling window expires at 2000 ms — wait, the rolling window expires first; the request must arrive at 1100 ms < 2000 ms so it's rejected by the rolling window, not the lockout) and add the missing rolling-window-expiry assertion that the previous test was meant to exercise but didn't actually exercise.

wjc2821296948:
## `ba75a6b` + `dae890b` — harden rate limiter proxy trust and bound the records map

### T1 — `X-Forwarded-For` only honoured when the TCP peer is a trusted proxy

The previous `readClientKey` accepted `X-Forwarded-For` whenever the TCP peer was loopback, and otherwise fell back to the socket address. That conflated two distinct concepts: "is the TCP peer a trusted reverse proxy?" and "is the TCP peer on the loopback interface?". Any local process able to open a connection to the server could pick a victim IP via `X-Forwarded-For` and bypass the limiter; clients behind a real reverse proxy all shared the proxy's loopback address and therefore shared a single limiter bucket.

Add an explicit `trustedProxyAddresses: string[]` option. Only when the TCP peer matches one of those addresses do we honour `X-Forwarded-For`. Loopback is NOT implicitly trusted — operators must enumerate their proxy's addresses if they run behind one.

```diff
-function readClientKey(req: Request): string {
-  const socketAddress = req.socket?.remoteAddress;
-  if (socketAddress && socketAddress !== '::1' && socketAddress !== '127.0.0.1') {
-    return socketAddress;
-  }
-  const forwarded = req.headers['x-forwarded-for'];
-  if (typeof forwarded === 'string' && forwarded.trim()) {
-    return forwarded.split(',')[0]!.trim();
-  }
-  ...
+function readClientKey(req: Request, trustedProxyAddresses: Set<string>): string {
+  const socketAddress = req.socket?.remoteAddress || 'unknown';
+  if (trustedProxyAddresses.has(socketAddress)) {
+    const forwarded = req.headers['x-forwarded-for'];
+    if (typeof forwarded === 'string' && forwarded.trim()) return forwarded.split(',')[0]!.trim();
+    ...
+  }
+  return socketAddress;
+}
```

### T2 — `records` map bounded by opportunistic eviction

`records` retained every client key indefinitely. Filtering the timestamps list of an existing record does not remove the entry from the map, so requests from many distinct addresses caused unbounded memory growth and retained IP-address data after the rate-limit window ended.

Add an opportunistic sweep that runs at most once per `windowMs` (so it costs O(records.size) at most once per window, not per request) and deletes every record whose `timestamps` array is empty and whose `blockedUntil` has expired. The map cannot grow without bound regardless of how many distinct client keys we see.

```diff
+ let lastSweepAt = clock();
+ const sweepExpiredRecords = () => {
+   const now = clock();
+   if (now - lastSweepAt < windowMs) return;
+   lastSweepAt = now;
+   const cutoff = now - windowMs;
+   for (const [key, record] of records) {
+     if (isRecordExpired(record, cutoff)) records.delete(key);
+   }
+ };
```

### New tests

- `X-Forwarded-For is honoured only when the TCP peer is a trusted proxy` — verifies the trusted-proxy allowlist path and that spoofers from a non-trusted peer are bucketed by TCP peer.
- `untrusted loopback peers cannot spoof X-Forwarded-For` — confirms the previous bypass on `127.0.0.1` is closed.
- `records map is bounded by expiring entries after the window passes` — pushes 50 distinct addresses, advances the clock, asserts only the just-touched record remains.

wjc2821296948:
## `8955283` — `fix(auth): walk X-Forwarded-For right-to-left and expose size() for tests`

Two follow-ups on the previous rate-limit hardening:

### T1 — pick the nearest untrusted forwarded address

The previous trusted-proxy parser picked the leftmost comma-separated value from `X-Forwarded-For`. That value is the most distant hop and therefore the easiest for the real client to forge: a malicious client can put any address they want there, the trusted proxy appends the real client address on the right, and the parser returned the attacker-supplied value. The attacker could then rotate that value to bypass the per-client limit.

Walk the chain from right to left, skipping any entry that is itself in `trustedProxyAddresses` (proxy chains like `attacker → proxy1 → proxy2 → server` are common), and return the first remaining untrusted address. If the chain contains only trusted hops, fall back to the TCP peer — never trust a header value the proxy chain couldn't validate.

### T2 — expose a typed inspection hook

The previous test cast `limiter as unknown as { records: ... }` to read the closure-local map. `createRateLimiter` only returned `middleware` and `reset`, so reading `.records.size` threw at runtime. Expose a typed `size()` inspection hook on the factory result and update the test to use it.

### New tests

- `trusted proxy chain selects the nearest untrusted client address` — covers a two-hop trusted proxy chain where the attacker forged the leftmost value and the real client address sits on the right; verifies the attacker cannot land in the real client's bucket.
- `trusted proxy chain with only trusted hops falls back to TCP peer` — verifies that a header containing only trusted hops does not produce a spurious key.

wjc2821296948:
## `52d5cd0` — `test(auth): tighten proxy tests to exercise trusted-hop skipping and TCP fallback`

CodeRabbit correctly noted that the two proxy-chain tests I added previously could pass without exercising the properties they claimed.

### Test #1 — trusted-hop skipping

The previous "trusted proxy chain selects the nearest untrusted client address" test included `198.51.100.99, 198.51.100.7` in the header, neither of which was a configured trusted proxy. The parser would pick the rightmost untrusted entry regardless of whether the trusted-hop skip was implemented correctly, so the test gave a false sense of coverage.

Update the header to include the inner trusted proxy `10.0.0.1` between the forged leftmost value and the real client: `198.51.100.99, 10.0.0.1, 198.51.100.7`. The parser must now skip `10.0.0.2` (the TCP peer) and `10.0.0.1` (the configured trusted hop) before landing on `198.51.100.7`. A buggy implementation that picks the leftmost or rightmost value without skipping trusted hops would assign the request to the wrong bucket.

### Test #2 — TCP-peer fallback

The previous "trusted proxy chain with only trusted hops falls back to TCP peer" test configured only `10.0.0.1` as trusted while the header included `10.0.0.2, 10.0.0.3, 10.0.0.1`. Walking right-to-left, `10.0.0.1` was skipped (trusted) but `10.0.0.3` was NOT trusted and would be returned, so the test was actually exercising the untrusted-hop case, not the TCP-peer fallback.

Configure every header hop (`10.0.0.1`, `10.0.0.2`, `10.0.0.3`) as trusted. After the first call lands on the TCP peer fallback (`10.0.0.2`), send a second request from the same TCP peer *without* the `X-Forwarded-For` header and assert it receives `429`. The second request shares the TCP-peer bucket and proves the fallback path is observable end-to-end.

blackmammoth:
@wjc2821296948 PR is too broad.

PR #1125, opened 2026-08-08
fix(chat): scope token-budget updates to the viewed session and stop multi-call inflation

thevinchi:
The tokens badge showed other sessions' numbers, ran 2-15x too high, and sometimes dropped to
"0 tokens".

## Why

Two things. The live websocket path and the REST path were measuring different quantities, and a
single React state took writes from every run attached to the socket.

REST reads the transcript backwards to the newest assistant usage, so it reports context occupancy
at the last API call. The live path fell back to `sdkMessage.usage` on the terminal `result` frame,
which sums every API call in the turn. Cache reads repeat on each call, so that lands near
`N_calls × context_size`, and it arrives last so it stuck until a refresh.

The `token_budget` branch was also the only handler path not checking the active session, and the
REST fetch had no stale-response guard.

## What changed

- Extraction resolves per-call usage, never the aggregate. Sidechain and zero-total frames yield
  nothing.
- Live frames are recorded per session and only paint the viewed one.
- The REST read is ordered by a request counter. Mount may blank the badge, run completion may not.
- The badge re-reads the transcript when a run ends, so it converges without a refresh.
- The transcript scan skips zero-usage entries, matching what live extraction already did.

## Validation

Aggregation inflated real turns 2.95-4.51x. Worst case reported 461,679 against a true 156,635.
After the fix, a probe with 5 tool calls and a subagent measured 30,841 live against 30,841 from
the transcript. Two concurrent sessions stayed independent over 6 minutes.

The `iterations` type check is defensive, not load-bearing: 7,320 arrays scanned, all length 1 and
typed `'message'`. The inflation fix is entirely the aggregate fallback going away.

## Verify

- [ ] Two sessions, viewing A while B streams: badge moves only for A.
- [ ] Multi-tool-call turn: badge matches a refresh, not a multiple of it.
- [ ] Turn ends untouched: badge reaches the transcript value on its own.
- [ ] Task fan-out: badge tracks the main thread, never flips to 0 or a subagent's count.

## Out of scope

- `CONTEXT_WINDOW` defaults to 160,000. This fixes the numerator, not the denominator.
- The `project_path` sanitizer assumes the transcript dir comes from `cwd`. It comes from the launch
  root.
- Cursor renders "0 tokens" because nothing reads the existing `supportsTokenUsage: false`.

## Checks

`npm run typecheck`, `npm run build`, eslint clean, `npm test` 13/13.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved token usage tracking by selecting the latest valid usage data and ignoring synthetic or zero-value entries.
  - Prevented token-budget updates from appearing as chat activity or affecting unrelated subagent messages.
  - Preserved accurate usage badges when reopening sessions or receiving realtime updates.
  - Refreshed usage data after completed sessions and cleared outdated badges when no valid usage is available.

- **Tests**
  - Added coverage for assistant, cache, result, compaction, and missing-usage scenarios.


liran-funaro:
Independent confirmation of your central claim, plus one measurement that I think this PR needs before merge.

**Confirmed:** the terminal `result`'s usage is summed over the turn, not context occupancy. A five-call turn on v1.37.2 reported `input_tokens 5479` + `cache_creation_input_tokens 39270` + `cache_read_input_tokens 111775` + `output_tokens 312` = 156,836, for a conversation that ended on nearer 45,000. Preferring per-call usage is right.

**The measurement:** on that same turn, and on a single-call one, `usage.iterations` came back **empty** — `Array.isArray(iterations) === true`, `iterations.length === 0` — while the top-level totals were fully populated. `resolvePerCallUsage` returns `null` for an empty array, so the `result` frame would never produce a reading here.

That matters because of what the per-call assistant frames look like on this deployment: they carry `input_tokens` and `output_tokens` and **nothing else** — no `cache_creation_input_tokens`, no `cache_read_input_tokens`. With no cache accounting in them, `input_tokens` is the uncached remainder, which against a warm cache is 2. The five frames of that turn read `5473, 5473, 2, 2, 2`.

So with this PR applied on a stack like this one, the badge would take those per-call frames and never get corrected by the `result` — landing on `2 tokens` and staying there for the session. Today the (inflated) total at least keeps it in the right order of magnitude, so that would be a regression rather than an improvement, for anyone whose SDK version or gateway leaves `iterations` unpopulated.

Two ways to close it, either fine by me:

- treat a payload as a reading only when it is a **complete** account — at least one cache field present — and publish nothing when neither the per-call frame nor `iterations` gives you one. The counter then keeps its last complete figure instead of taking a partial. (This is what I filed as #1208 and what I run locally.)
- or keep a guarded fallback to the aggregate when `iterations` is empty, accepting the overstatement rather than a single-digit understatement.

Caveat on all of the above: my CLI runs behind an LLM gateway (`ANTHROPIC_BASE_URL` is not `api.anthropic.com`), so I cannot tell whether the absent cache fields and the empty `iterations` originate with the CLI or with the gateway's normalisation of the upstream response. If `iterations` is reliably populated on a direct-API run, this is a narrower concern than it looks — but it is cheap to feature-detect one level further than the array's existence.

blackmammoth:
@thevinchi, this should be fixed with the latest main commits.

liran-funaro:
Agreed on the part that main now fixes, and it is the substance of this PR: since #1206, the counter reads the **per-call** assistant usage rather than the turn-summed `result`, and `extractTokenBudget` additionally drops subagent traffic (`parent_tool_use_id`) and non-assistant frames whose `usage` is shaped `{total_tokens, tool_uses, duration_ms}`. The inflation @thevinchi opened this against is gone, and `extractCumulativeTokenBudget` is now documented as the last-resort path instead of the default one. So closing the PR is fair.

One narrower thing this PR's discussion raised is still open on main, which is why I'd rather flag it here than let it sink with the thread: **a partial per-call payload is published as if it were a reading.**

`buildTokenBudget` reads each field through `readNumber`, which turns anything non-finite into `0`:

```js
const cacheCreationTokens = readNumber(messageUsage.cache_creation_input_tokens ?? …);
const cacheReadTokens = readNumber(messageUsage.cache_read_input_tokens ?? …);
…
return { used: inputTokens + outputTokens, … };
```

and `extractTokenBudget` hands it any assistant `message.usage` object without checking that the cache half is there:

```js
const messageUsage = sdkMessage.message?.usage;
if (!messageUsage || typeof messageUsage !== 'object') {
  return null;
}
return buildTokenBudget(messageUsage);
```

On a stack whose assistant frames carry only `input_tokens` and `output_tokens`, both cache fields default to 0 and `input_tokens` is the *uncached remainder* — 2 against a warm cache. That is the measurement I posted upthread: five frames of one turn reading `5473, 5473, 2, 2, 2` on a conversation holding ~45,000. The client publishes whatever arrives (`if (msg.text === 'token_budget' && msg.tokenBudget)` — no completeness check either), so the badge settles on single digits for the rest of the session.

The fix is one guard, not a redesign: treat a payload as a reading only when it is a complete account — at least one cache field actually present — and publish nothing otherwise, so the counter keeps its last complete figure. That is #1208, which was swept closed as completed by #1206 even though this specific case survived it.

Happy to send it as a small PR with a test. I already have two open (#1157, #1274), so I'll queue it behind those unless you'd rather have it now.

PR #1129, opened 2026-08-09
feat(mobile): open the session list when back is pressed

materemias:
## What & why

On a phone, switching sessions means reaching for the menu button in the top-left — awkward one-handed, and the most common thing I do on mobile. This adds an **opt-in** Appearance setting, **Back Opens Session List**, that turns the device back gesture into "show the session list" instead.

With it on, jumping between sessions is: back → tap session → back → tap session.

The setting **defaults to off** and is additionally gated on the mobile breakpoint (768px), so nothing changes for existing users or on desktop.

## Behaviour

| Press | Result |
|---|---|
| back, list closed and the guard is armed | list opens, URL unchanged |
| back, list open | normal browser navigation |

There is no separate "back closes the list" step. The hook does not re-arm while the list is open. On browsers with the user-activation API, a fresh page load leaves Back unchanged until the user interacts with the page.

## How it works

The browser gives no way to observe a back press without consuming a history entry, so the hook keeps a sentinel entry above the current page while enabled, activated, and the list is closed. The first back press pops the sentinel without changing the URL and opens the sidebar. It does not re-arm while the sidebar is open.

The hook distinguishes a guard pop from normal navigation using history-state markers:

- **Guard and base entries carry matching instance and guard IDs.** A pop opens the sidebar only when it lands on the matching base at the same URL. A multi-entry jump to another entry remains browser navigation. Arming checks `history.state`, so it also handles router navigation to the same URL.
- **Arming is deferred by a task.** React-router's own `popstate` listener runs first and re-renders synchronously, which flushes this hook's effect *inside the same dispatch*. Arming inline pushed a fresh sentinel that the hook's own listener then misread as a user pop, opening the list on what was a genuine back navigation.
- **Guard-removal `history.back()` calls are marked** so their `popstate` does not open the list. Popping a guard inherited from another mount skips its duplicate base entry without opening the list. Popping a guard during a setting or sidebar transition continues normal navigation when the feature is disabled or the list is already open.

Two limitations remain in the sentinel approach: `pushState` drops forward history, and because the sentinel is the top entry, `navigate(..., { replace: true })` replaces the sentinel rather than the entry the caller meant to drop.

## Scope

- `src/modules/project-workspace/hooks/useBackOpensSessionList.ts` — new hook
- `src/modules/project-workspace/ProjectWorkspaceShell.tsx` — wiring, gated on `isMobile && preference`
- `src/shared/uiPreferences.ts` — new preference key, default `false`
- `src/modules/settings/tabs/AppearanceSettingsTab.tsx` — the toggle
- `src/modules/i18n/locales/*/settings.json` — strings for all 11 locales

## Verification

Earlier verification reported a passing build, typecheck and eslint. No new run was performed for this description update.

The earlier 390×844 Chromium smoke check against a throwaway database covered a three-press sequence: open the list, close it, then navigate back. That sequence is no longer implemented, so it does not verify the current open-list-then-native-Back behavior.

The current diff includes hook tests for arming after interaction without stacking guards on rerender, skipping an inherited guard after remount, and leaving multi-entry jumps to browser navigation. No new test run or browser verification is claimed here.

Happy to change the approach or the naming — if the sentinel trick is not something you want in the codebase, say so and I will close this.

1-setting


materemias:
Thanks — went through all three review points. One was a real bug and is fixed in bfe9775; the other two are addressed below.

**Fixed: skip state stranded when `history.back()` cannot traverse** (CodeRabbit)

Correct, and reachable: turn the setting off (or rotate past the 768px breakpoint) while a sentinel is live and on the first history entry, and the internal `history.back()` is a silent no-op with no `popstate`, so `skippingRef` stayed set and the arming effect returned early forever.

There is a second half to it that the review did not mention and that I hit while testing the fix: the watchdog must not be torn down with the `popstate` listener. That listener re-subscribes on `enabled`/`sidebarOpen` changes — i.e. on exactly the two triggers above — so a cleanup-scoped timer cancels itself in the failing case. Its teardown is now mount-scoped.

Verified in a 390x844 Chromium: with the skip made non-traversable, the hook re-arms within the watchdog window and the next back press opens the list again (it stayed dead before the fix). Core cycle re-checked afterwards — back opens, back closes, back navigates out, and the session-jump loop re-arms with no dead press.

**Not a bug: `{ ...history.state }` when state is `null`** (Copilot)

Object spread of `null`/`undefined` is a no-op by spec (CopyDataProperties returns early), and primitives spread to `{}` — it cannot throw:

```
$ node -e 'console.log(JSON.stringify({...null}), JSON.stringify({...undefined}), JSON.stringify({...5}))'
{} {} {}
```

The `try/catch` is there for a different reason: Safari throttles `pushState` (~100 calls / 30 s) and throws when it trips.

**Known limitation, documented rather than fixed: `replace` targets the sentinel** (CodeRabbit)

Accurate, and already called out in the hook doc block. The one `replace: true` caller is the provider-alias-to-canonical swap in `useProjectsState`; with the guard armed, the alias entry survives instead of being dropped, so backing out of such a session costs one extra press (it lands on the alias URL, which re-resolves to the canonical one).

The suggested fix — disarm before every replace and re-arm after — means coupling the hook to every current and future `navigate(..., { replace: true })` call site. That is a lot of standing coupling for one extra back press on alias URLs only, so I left it documented alongside the other inherent cost of the sentinel approach (`pushState` drops forward history). Happy to implement it if maintainers prefer the coupling.

materemias:
### Update: five commits since `bdb64126`

The head moved from `bdb64126` to `0a5cfae6`. This is what changed and why, including the review comment it answers.

**`2a7f709b` — arm a pair of guard entries, so back can still close the list**

Answers @coderabbitai's Functional Correctness finding on `723eaed5..bdb64126`: with a single sentinel, the first back popped it, and the arming effect could not push a replacement (the pop stops Chrome's history manipulation intervention from honouring the earlier activation), so the next back navigated away instead of closing the list.

The intervention's own rule is the way out — *"with an activation, the document can create many unskippable same-document history entries, until either a cross-document navigation or a back/forward occurs"* — so the hook now pushes **two** guards in the same arming pass while the activation is live. Both entries are unskippable, and the traversal that pops the second cannot retro-mark the first.

Each guard carries an id in its state, because a live guard and one buried by a router push are otherwise indistinguishable while meaning opposite things. What a pop *does* is still decided from live `sidebarOpen` at pop time, not from which half of the pair was consumed — the list can be opened with the menu button between arming and the press, and that case must still close in one press. `closedByBackRef`/`closedByBackHrefRef` are gone: a spent pair ends the sequence structurally.

**`9374f390` — say in the copy that back is untouched until the page is interacted with**

Before the first touch on a freshly loaded page, back is the browser's own gesture and leaves the app: pushing a guard without an activation would mark *every* same-document entry skippable, and on Android the back button then closes the tab rather than popping anything. No API unmarks an entry; only a user gesture clears the flag. That is the intervention's stated purpose, not something this PR can work around, so it is documented instead — one sentence per locale, e.g. EN: *"Straight after a page load, back is left alone until you touch the screen."*

**`23291dae` — stop a jump or a disabled hook from eating a history entry**

A back press pops exactly one entry and these guards are never skippable, so a multi-entry jump (back-button long-press menu, `history.go(-n)`) must not be read as spending them; and guards left over from before the option was switched off must complete the navigation instead of being interpreted.

**`37964e15` — decide a guard pop by the entry the press left**

Reading only the landing entry conflated two states needing opposite handling: `remaining === 0` was both "the press left my last guard" (a no-op traversal to complete) and "a jump landed here" (the user already moved). With one guard left after the list opened, the disabled path took the second reading and swallowed the press. `popstate` reports no distance, so the entry the traversal *left* is now tracked; a press the hook may interpret is exactly one that stepped off its own guard onto the entry directly below.

**`0a5cfae6` — drop buried guard ids whether or not the hook can arm**

Bookkeeping only, no behaviour change (measured, not assumed): the arming pass returned before its own cleanup when the option was off or no activation was available.

### Verification

Headless Chrome at 390px with the DevTools Audits domain listening; all state read over CDP with `userGesture: false` so the probe grants no activation of its own. No frontend test infrastructure exists in this repo, so this is measured against the running app.

| scenario | result | `NavigationEntryMarkedSkippable` |
|---|---|---|
| fresh load, no interaction → back | leaves the app (platform's own gesture) | 0 |
| tap → back ×3, **no taps in between** | list opens → list closes → leaves | 0 |
| menu button opens list → back | closes in one press → next back leaves | 0 |
| `history.go(-2)` past both guards | stays on the app entry, list untouched | 0 |
| back opens list → option switched off → back | leaves the app | 0 |
| session opened → back ×5 | open → close → `/` idx 1 → `/` idx 0 → leaves | 0 |
| session A → option off → session B → back ×3 | idx 4 → 3 → 2 → 1, buried guards consumed in one press | 0 |
| `notification:navigate` push (with the option on and off) | consumes the guard, lands on the real entry, next press exits | 0 |

react-router's `idx` is what proves nothing is swallowed: two app entries can share the URL `/` (the alias in `useProjectsState`), so only the index shows that a press moved.

### Known limitation, stated plainly

Two remain, both inherent to the sentinel approach: `pushState` drops forward history, so the forward gesture stops working while the option is on; and because a guard is always the top entry, `navigate(..., { replace: true })` replaces a guard rather than the entry the caller meant to drop. The feature is off by default.

materemias:
Does NP mean “not a priority”? Is this feature acceptable in principle but waiting for review, or would you prefer not to support the history-sentinel approach? I’m happy to keep it fork-only if the navigation tradeoffs aren’t appropriate upstream.

PR #1130, opened 2026-08-10
fix(codex): parse cmd from exec command history

fedecia:
## Summary

- recognize the `cmd` key emitted by current Codex `tools.exec_command` rollout records
- keep those history entries normalized as `Bash` instead of falling back to generic `exec`
- add a regression fixture using the observed `{"cmd":"git diff --check","workdir":"/workspace"}` payload

## Context

This is a focused follow-up to #1095 and the regression reported in #1089.

The parent PR recognizes the `tools.exec_command` wrapper, but its command extractor only accepts `command`. Current Codex rollouts use `cmd`, so the tool/result pair is preserved but displayed as generic `exec`. Supporting both keys restores the intended Bash rendering while retaining the parent's generic fallback for other wrappers.

## Verification

- `npx tsx --tsconfig server/tsconfig.json --test server/modules/providers/tests/codex-sessions.test.ts` — 4 passed
- `npm test` — 260 passed
- `npm run build`
- `npm run typecheck`
- `npm run lint` — 0 errors (existing branch warnings only)

### DIFF
--- a/server/modules/providers/list/codex/codex-sessions.provider.ts
+++ b/server/modules/providers/list/codex/codex-sessions.provider.ts
@@ -143,7 +143,7 @@ function decodeJavaScriptStringLiteral(literal: string): string {
 
 function extractNestedCodexCommands(source: string): string[] {
   const commands: string[] = [];
-  const commandPattern = /(?:["']command["']|\bcommand)\s*:\s*("(?:\\.|[^"\\])*"|'(?:\\.|[^'\\])*'|`(?:\\.|[^`\\])*`)/gs;
+  const commandPattern = /(?:["'](?:command|cmd)["']|\b(?:command|cmd))\s*:\s*("(?:\\.|[^"\\])*"|'(?:\\.|[^'\\])*'|`(?:\\.|[^`\\])*`)/gs;
   for (const match of source.matchAll(commandPattern)) {
     commands.push(decodeJavaScriptStringLiteral(match[1]));
   }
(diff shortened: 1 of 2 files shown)


blackmammoth:
hey @fedecia, can you provide a reproduction step for the issue this attempts to fix?

fedecia:
hey @blackmammoth — thanks for looking. repro + evidence below. i also retargeted this to `main`: it was stacked on `fix/update-codex-sdk-and-fix-data-usage`, which is why coderabbit skipped the review ("auto reviews are disabled on base/target branches other than the default branch"). against `main` it is 2 files, +7/-1.

**repro**

1. run any Codex CLI session that shells out (mine is `codex-cli 0.146.1`)
2. open the rollout at `~/.codex/sessions/**/rollout-*.jsonl`
3. grep the `custom_tool_call` inputs:

```bash
grep -o 'tools\.exec_command([^)]\{0,80\}' rollout-*.jsonl
```

4. every record looks like `await tools.exec_command({"cmd":"git diff --check","workdir":"/workspace"});`

`extractNestedCodexCommands` only accepts `command`, so those entries keep the tool/result pair but render as generic `exec` instead of `Bash`.

**how often** — across 110 local rollout files:

| | count |
|---|---|
| `tools.exec_command(...)` calls | 9,469 |
| emitting `cmd` | 9,404 |
| emitting `command` | 11 |

the 11 `command` hits are not real Codex output — every one is `{"command":"echo current"}`, this repo's own test fixture echoed back inside a transcript. so on 0.146.1 the key the current regex matches occurs zero times in practice. the `cmd`-key records span 7–17 Aug 2026.

caveat, stated plainly: this is one workstation and one Codex version. it shows `cmd` dominates on 0.146.1; it does not pin down which release switched.

**verified on `main`** (not on the old base): `codex-sessions.test.ts` goes 4 → 5 tests, and `npm test` is 263 passed / 0 failed.

happy to widen the fixture or split the regex into two named patterns if you'd rather keep `command` and `cmd` explicit rather than alternated.

blackmammoth:
hey @fedecia, thanks. This has been fixed with latest commits in main. Check them out and if issue still persists, you can submit a new PR.

PR #1132, opened 2026-08-11
fix(claude): read the model catalog from the CLI instead of a hardcoded table

### LINKED ISSUE #1131: Claude model picker shows stale version labels (hardcoded catalog)
## What's wrong

The Claude model picker's options come from a hardcoded catalog in `server/modules/providers/list/claude/claude-models.provider.ts`. It falls behind every time a new Claude model ships:

- `default` is labeled "currently Sonnet 5" — it actually resolves to `claude-opus-5[1m]` on current CLI builds.
- `opus` / `opus[1m]` are labeled "Opus 4.8" — they resolve to Opus 5.

I confirmed this by comparing the static table against `queryInstance.supportedModels()` on the actually-installed CLI (2.1.227): the SDK reports different `displayName`/`description` text than what's hardcoded, and `resolvedModel` for `opus[1m]` is `claude-opus-5[1m]`, not 4.8.

## Why it's hardcoded

`getSupportedModels()` has the dynamic lookup written but commented out:

```js
// claude creates a new jsonl file as a separate session for this request.
// As a result, it lists the workspace where this is invoked when it shouldn't.
//
// Disabled for now:
// const queryInstance = query({ prompt: 'Get supported models', options: buildClaudeQueryOptions() });
// ...
return CLAUDE_FALLBACK_MODELS;
```

## What I found

The jsonl-pollution concern seems addressable: calling `query()` with `persistSession: false`, a prompt that never yields (so `supportedModels()` is answered from the CLI's startup handshake without starting a turn), and `cwd` outside any project directory, I get the live catalog with no session transcript written and no tokens spent. Also worth pulling in: `pathToClaudeCodeExecutable` needs to point at the same binary the runtime uses — without it the SDK falls back to its own bundled CLI, which is typically older than whatever's actually driving chats, and reports an even staler catalog.

## Fix

I have a working fix — dynamic lookup with the static table as a same-shape fallback (missing/outdated CLI, probe timeout, empty response), plus:
- CLI-reported entries win; static-table-only aliases (ones the CLI no longer advertises but the SDK confirms it still resolves) are appended rather than dropped, so a stored per-session selection doesn't get silently reset.
- The provider models cache (3-day TTL, disk-persisted) now covers Claude too, since the catalog is no longer a free constant.

Verified against the real provider class (not mocked): concurrent calls share one in-flight probe, the live catalog comes back correctly labeled, `getCurrentActiveModel()` stays fast (uses the static default, no CLI spawn on that hot path), and no jsonl transcript appears under `~/.claude/projects` for the probe's temp cwd.

PR incoming with this fix — wanted to open the issue first per CONTRIBUTING.md.

mxx1111:
Fixes #1131

## Problem

The Claude model picker's options come from a hardcoded catalog. It falls behind every model release:

- `default` was labeled "currently Sonnet 5" — it actually resolves to `claude-opus-5[1m]` on current CLI builds.
- `opus` / `opus[1m]` were labeled "Opus 4.8" — they resolve to Opus 5.

`getSupportedModels()` had the dynamic lookup written but disabled, with a comment explaining why: calling `supportedModels()` was creating a separate jsonl session file, polluting the workspace's session list.

## Fix

Re-enable the dynamic lookup, addressing the jsonl-pollution concern directly instead of working around it:

- `persistSession: false` on the query — no transcript is written.
- The prompt is a stream that never yields, so `supportedModels()` is answered from the CLI's startup handshake without ever starting a turn (no tokens spent, nothing to persist in the first place).
- `cwd` is a temp directory outside any project, so even the CLI's own state files land somewhere disposable.
- `pathToClaudeCodeExecutable` resolves the same way the runtime resolves it (`resolveClaudeCodeExecutablePath`). Without this the SDK falls back to its own bundled binary, which is typically older than whatever's actually driving chats — I hit this in testing: the bundled 2.1.165 binary reported "Opus 4.8 / Sonnet 4.6", while the installed 2.1.227 binary (what the runtime actually uses) correctly reported Opus 5 / Sonnet 5.

**Fallback behavior:** if the probe fails, times out (30s), or returns nothing, `getSupportedModels()` falls back to the same static table as before — so a missing/very old CLI degrades to the previous (occasionally stale) behavior rather than breaking.

**Alias retention:** CLI-reported entries win (current names, descriptions, effort levels). Options that only the static table knows about — aliases the CLI has stopped advertising but the SDK confirms it still accepts — are appended rather than dropped. This matters because the CLI stops *advertising* an alias well before it stops *accepting* it; dropping one outright would silently reset the stored per-session model selection for anyone who'd picked it.

**Caching:** `claude` is removed from `UNCACHED_PROVIDERS` in the provider-models service. That exemption existed because the old catalog was a free in-memory constant — now that reading it means spawning a process, it's exactly the kind of lookup the existing 3-day disk-persisted cache is for. Concurrent-caller coalescing (`pendingRequests`) also now applies to Claude, so a burst of picker opens right after a cold start only spawns the CLI once.

## Verification

Ran the real `ClaudeProviderModels` class directly (not mocked) against the installed CLI:

- Concurrent calls to `getSupportedModels()` share one in-flight probe.
- Returned catalog has correct current labels (Opus 5, Sonnet 5, etc.) and effort levels straight from the CLI.
- Legacy aliases the CLI no longer lists (`opus`, `sonnet[1m]`, `fable`) are retained from the fallback table.
- `getCurrentActiveModel()` (a per-session hot path) stays at 0ms — it intentionally uses the static default rather than the probed catalog, since spawning a process there isn't worth it for a value that's `default` in every catalog anyway.
- No jsonl file appeared under `~/.claude/projects` for the probe's temp cwd, confirming the pollution the original comment described no longer happens.

On this PR's branch (rebased onto current `main`): `tsc --noEmit` clean, eslint 0 errors, and the provider-models test suite passes (17/17, including an updated test that now exercises Claude going through the same cache path as every other provider instead of asserting it bypasses caching).

## Notes for review

- Timeout is 30s (process spawn + CLI startup); a slow/broken machine degrades to the static table rather than stalling the request.
- I did not touch the static table's *content* beyond what's needed for the merge to make sense — it stays as the fallback and the source of retained legacy aliases.

## Summary by CodeRabbit

* **New Features**
  * Added automatic discovery of installed Claude models through the Claude CLI.
  * Added fallback Claude model options when discovery is unavailable or incomplete.
  * Updated Claude model names and descriptions with current metadata.

* **Bug Fixes**
  * Improved reliability when discovery times out, fails, returns no results, or closes unexpectedly.
  * Ensured current-model selection consistently uses the supported static catalog.

* **Performance**
  * Improved model loading by sharing concurrent discovery requests and consistently using cached results.
  * Added shorter-lived, memory-only caching for fallback model catalogs.
  * Added fresh catalog loading when bypassing the cache.

### DIFF
--- a/server/modules/providers/list/claude/claude-models.provider.ts
+++ b/server/modules/providers/list/claude/claude-models.provider.ts
@@ -220,14 +220,14 @@ const probeClaudeSupportedModels = async (): Promise<ProviderModelsDefinition> =
 
     if (options.length === 0) {
       console.warn('[Claude models] CLI reported no models; using the built-in catalog');
-      return CLAUDE_FALLBACK_MODELS;
+      return { ...CLAUDE_FALLBACK_MODELS, isFallback: true };
     }
 
     return mergeWithFallbackModels(options);
   } catch (error) {
     const message = error instanceof Error ? error.message : String(error);
     console.warn(`[Claude models] Unable to read the CLI catalog (${message}); using the built-in catalog`);
-    return CLAUDE_FALLBACK_MODELS;
+    return { ...CLAUDE_FALLBACK_MODELS, isFallback: true };
   } finally {
     clearTimeout(timeout);
     try {
--- a/server/modules/providers/list/claude/claude-models.provider.ts
+++ b/server/modules/providers/list/claude/claude-models.provider.ts
@@ -1,6 +1,10 @@
 import { readFile } from 'node:fs/promises';
+import os from 'node:os';
+
+import { query } from '@anthropic-ai/claude-agent-sdk';
 
 import { sessionsDb } from '@/modules/database/index.js';
+import { resolveClaudeCodeExecutablePath } from '@/shared/claude-cli-path.js';
 import type { IProviderModels } from '@/shared/interfaces.js';
 import type {
   ProviderCurrentActiveModel,
@@ -14,7 +18,7 @@ export const CLAUDE_FALLBACK_MODELS: ProviderModelsDefinition = {
     {
       value: 'default',
       label: 'Default (recommended)',
-      description: 'Use the Claude Code default model (currently Sonnet 5)',
+      description: 'Use the Claude Code default model (currently Opus 5 with 1M context)',
       effort: {
         default: 'high',
         values: [
@@ -71,7 +75,7 @@ export const CLAUDE_FALLBACK_MODELS: ProviderModelsDefinition = {
     {
       value: 'opus',
       label: 'Opus',
-      description: 'Opus 4.8 · Best for everyday, complex tasks · ~2× usage vs Sonnet',
+      description: 'Opus 5 · Best for everyday, complex tasks · $5/$25 per Mtok',
       effort: {
         default: 'high',
         values: [
@@ -85,8 +89,8 @@ export const CLAUDE_FALLBACK_MODELS: ProviderModelsDefinition = {
     },
     {
       value: 'opus[1m]',
-      label: 'Opus 4.8 (1M context)',
-      description: 'Opus 4.8 with 1M context · Most capable for complex work · $5/$25 per Mtok',
+      label: 'Opus (1M context)',
+      description: 'Opus 5 with 1M context · Best for everyday, complex tasks · $5/$25 per Mtok',
       effort: {
         default: 'high',
         values: [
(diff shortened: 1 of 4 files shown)


mxx1111:
CodeRabbit's inline comment (cache TTL / fallback discrimination) was from the first review pass at 00:39, before the second commit landed at 01:00. That commit (`fix: don't cache a fallback catalog for the same three days as a real…`) addresses it directly:

- `getSupportedModels()` now marks fallback results with `isFallback: true`
- `provider-models.service.ts` uses a 1-minute in-memory TTL for fallback results instead of the 3-day TTL used for real probes
- Fallback results are never written to the persisted disk cache, so a restart can't pin a stale fallback in place either

CodeRabbit re-ran after that commit (01:07) and passed with no new findings. This should be ready for a maintainer review whenever you get a chance — happy to make further changes if anything else comes up. Thanks!

blackmammoth:
@mxx1111, this is intentional.

PR #1133, opened 2026-08-11
fix(codex): preserve successful terminal status

fedecia:
## Summary

- skip the intermediate Codex `turn_complete` normalization
- let the runtime terminal message remain authoritative for `success` and `exitCode`
- add a regression test for the duplicate terminal event

## Problem

Codex emits a `turn_complete` stream event before the runtime emits its final normalized completion. The intermediate event lacks `success` and `exitCode`, but wins terminal-event deduplication, so clients can interpret a successful turn as failed.

## Testing

- `npm test`: 256 passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved live session completion handling to prevent duplicate completion messages.
  * Completion status is now reported only once, using the authoritative final event.

### DIFF
--- a/server/modules/providers/list/codex/codex-sessions.provider.ts
+++ b/server/modules/providers/list/codex/codex-sessions.provider.ts
@@ -850,13 +850,11 @@ export class CodexSessionsProvider implements IProviderSessions {
     }
 
     if (raw.type === 'turn_complete') {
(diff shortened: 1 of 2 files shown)


blackmammoth:
hey @fedecia, can you resolve the merge conflicts?

PR #1140, opened 2026-08-11
fix(chat): close the streamed row at each message boundary

### LINKED ISSUE #1136: Streaming buffer has no message boundary: assistant messages in one turn are merged into a single row
### Confirmed defect: the streaming buffer has no message boundary

In `upstream/main` (v1.37.0) the live streaming buffer is closed only by `stream_end` or by `complete` at the end of the whole turn:

- `src/components/chat/hooks/useChatRealtimeHandlers.ts:179` — every `stream_delta` appends to a single `accumulatedStreamRef`.
- `:184` — the whole accumulator is written back as one row.
- `:206` / `:233` — it is cleared only by `stream_end`, or by `complete`.
- `src/stores/useSessionStore.ts:640` — the row uses the well-known id `__streaming_${sessionId}`, so successive writes replace it rather than starting a new row.

Nothing resets the buffer when an ordinary `text`, `thinking`, `tool_use` or `tool_result` frame arrives. The frontend therefore relies on each provider emitting `stream_end` per assistant message, and that does not hold for the shipped providers:

- `server/modules/providers/list/cursor/cursor-sessions.provider.ts:374-395` — the live `normalizeMessage` maps **every** assistant event to `stream_delta`, and the file never emits `stream_end` at all.
- `server/modules/providers/list/opencode/opencode-runtime.provider.js:224` — the live runtime path emits `stream_delta`, with no `stream_end` in that file.
- `server/modules/providers/list/claude/claude-sessions.provider.ts:316` does emit `stream_end`, which is why Claude is unaffected.

**Consequence that follows from the code alone:** for Cursor and for OpenCode's runtime path, one buffer spans the entire turn, so several distinct assistant messages are merged into a single row instead of appearing as separate rows in order. This part is a code reading of `upstream/main`, not a claim about a session I ran.

### Related symptom I observed, with upstream reproduction unknown

On an ACP-based provider I added in my own fork, the same missing boundary produced a visibly worse result: the real message rows appeared **and** an extra trailing bubble containing every streamed fragment of the turn concatenated together, growing until the turn ended. Because the split fell on a token boundary, the duplicated text broke mid-word — `web sh` in one bubble, `ell` in the next, each with its own timestamp.

That extra-row variant needs a non-delta frame to arrive mid-turn, which is what lays down a real row beside the still-open accumulator. **I have not verified that any provider you ship does that**, so please treat the duplicate-bubble form as unreproduced upstream. I am reporting it only because the mechanism is entirely in the shared frontend files above, so it is reachable from any provider whose live event stream mixes deltas with other kinds.

### Expected

Each assistant message in a turn gets its own row, ordered with the tool rows, without depending on providers to emit `stream_end` per message.

### Fix I run locally

Close the streamed row at every message boundary — any `text`, `thinking`, `tool_use` or `tool_result` frame ends the current buffer — instead of waiting for `stream_end`/`complete`, and key the buffer per session. Closing a row has to go through `finalizeStreaming` rather than merely resetting the accumulator, or the next message's first delta finds the old row still parked at the well-known id and overwrites it.

No second dedupe path is needed for this repo: `pruneRealtimeSupersededByServer` already drops a superseded partial row on refresh. See the comment below for the detail on all three points.

Happy to open a PR if it is useful.

---

*Edited to separate the code-confirmed defect from the symptom I could only observe on a fork-only provider; the original wording asserted the duplicate bubble as upstream behaviour, which the cited evidence does not establish. Edited again to strike a merge-time dedupe this section had listed as part of the fix — that pairing was my fork's requirement, not this repo's, and `upstream/main` already prunes the partial row.*

MannXo:
## Summary

Assistant messages inside one turn were merged into a single chat row, and that row sorted after the tool rows it was interleaved with. The streaming buffer had no notion of a message boundary, so it stayed open for the whole turn unless a provider happened to emit `stream_end` per message.

## Context

Fixes #1136.

`useChatRealtimeHandlers.ts` accumulated every `stream_delta` into one buffer and wrote it to a row keyed by the well-known id `__streaming_${sessionId}`. Nothing closed that buffer when an ordinary frame arrived, only `stream_end` or the terminal `complete`.

OpenCode reproduces it. Its live `normalizeMessage` maps `text` to `stream_delta`, `tool_use` to `tool_use`, and emits `stream_end` only at `step_finish`. A turn that calls a tool therefore streams text, appends the tool row, streams more text, and ends with one row holding both halves concatenated. `updateStreaming` re-stamps `timestamp` on every write, so the open row also outlived and sorted past the tool row.

Claude is unaffected because it emits `stream_end` at `content_block_stop`.

Cursor is listed in the issue as affected. Its live path emits `stream_delta` for every assistant event and nothing else, so no boundary frame ever reaches the frontend and a frontend fix cannot split its messages. That needs provider-side work and is out of scope here.

| Before | After |
| --- | --- |
| image | image |

## Changes

- Add `src/components/chat/hooks/streamBuffers.ts`. Holds the buffer mechanics as plain functions over a `Map` keyed by session id, with store writes and timers injected so the logic is testable without React.
- Close the open row on any `text`, `thinking`, `tool_use` or `tool_result` frame, before that frame's own row is appended.
- Close through `finalizeStreaming` rather than only clearing the accumulator. `updateStreaming` replaces by id, so a boundary that left the row parked at `__streaming_${sessionId}` would have the next message overwrite it, giving one row per turn again holding only the last message.
- Call `finalize` on close even when no buffer is open. A row outlives its buffer when the view moves off a streaming session, and finalizing frees the well-known id. The store call is a no-op when no such row exists.
- Key buffers and their flush timers per session, matching the row they already write to. Previously a single accumulator could flush a background session's text into whichever session the current frame named, and the terminal paths cancelled pending flushes belonging to other sessions.
- `resetStreamingState` now takes an optional session id, so tearing down one view drops only that session's buffer. Called with no argument on unmount, where dropping everything is correct.
- Skip the final write when the row already holds the buffered text, so closing right after a flush does not write twice.

No merge-time dedupe was added. `pruneRealtimeSupersededByServer` already drops the superseded partial row on refresh, and `updateStreaming` stamps each row with the time of its last delta, so a finalized row sorts between the tool rows around it.

## Testing

Node 22, matching `.nvmrc`.

```
npm test        282 passing, 0 failing   (12 new)
npm run typecheck   clean
npm run lint    0 errors, 212 warnings   (212 on main, unchanged)
npm run build   exit 0
```

New tests in `src/components/chat/hooks/streamBuffers.test.ts` cover delta coalescing, one row per message across a boundary, closing a buffer whose flush never fired, per-session isolation of both text and timers, and dropping one session's buffer while another stays live. Three mutations were used to confirm the tests bite. Removing the `finalize` call fails 6, sharing one buffer across sessions fails 8, and not cancelling the pending flush on close fails 1.

The wiring in `useChatRealtimeHandlers.ts` is not unit tested. There is no frontend component test runner in the repo, so the handler's decision to close on boundary frames is covered by the before and after captures above rather than by a test.

The first commit adds `src/**/*.test.ts` to the `npm test` glob. The runner globbed `server/**` only, so the existing `src/stores/sessionMessageReconciliation.test.ts` was never executed. It passes. Happy to drop that commit and leave the new test unrun if you would rather keep the glob as it is.

On how the captures were produced. No provider CLI is installed here and OpenCode needs credentials, so instead of driving a live run I replayed OpenCode's frame sequence for one turn over the chat websocket into the real UI, against the real store and render path. Frames used, in order: user `text`, `stream_delta` chunks, `tool_use` with a result, more `stream_delta` chunks, `stream_end`. I also replayed Claude's sequence, deltas then `stream_end` then the assistant `text` row that echoes the same content, and it still renders as one bubble.

## Notes for reviewers

`text` is treated as a boundary regardless of role. A user `text` frame mid-turn is unusual, and closing on it is harmless because there is no open buffer at that point.

`error`, `interactive_prompt` and `task_notification` are deliberately not boundaries. They append rows too, so the same ordering argument could extend to them, but the issue named the four kinds above and widening the set risks splitting a message on an unrelated notification. Easy to add if you would prefer that.

This PR is frontend only, so `AGENTS.md` points the backend module standards at `server/` and none of it applies.

Fixes #1136

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat streaming reliability by preserving message order and preventing duplicate updates.
  * Added support for concurrent sessions with isolated streaming state.
  * Ensured pending stream content is finalized or safely discarded during session changes and completion.

* **Tests**
  * Expanded automated coverage for streaming boundaries, flushing, cleanup, concurrency, and interleaved tool-call messages.
  * Updated test discovery to include component-level tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "build:server": "tsc -p server/tsconfig.json && tsc-alias -p server/tsconfig.json",
     "preview": "vite preview",
     "typecheck": "tsc --noEmit -p tsconfig.json && tsc --noEmit -p server/tsconfig.json",
-    "test": "tsx --tsconfig server/tsconfig.json --test \"server/**/*.test.ts\" \"server/**/*.test.js\"",
+    "test": "tsx --tsconfig server/tsconfig.json --test \"server/**/*.test.ts\" \"server/**/*.test.js\" \"src/**/*.test.ts\"",
     "lint": "eslint src/ server/",
     "lint:fix": "eslint src/ server/ --fix",
     "start": "npm run build && npm run server",
--- a/src/components/chat/hooks/streamBuffers.test.ts
+++ b/src/components/chat/hooks/streamBuffers.test.ts
@@ -1,0 +1,234 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+
+import type { StreamBuffers, StreamFlushScheduler, StreamRowSink } from './streamBuffers';
+import {
+  appendStreamDelta,
+  closeStreamBuffer,
+  dropStreamBuffers,
+  isStreamBoundaryKind,
+} from './streamBuffers';
+
+type Write = { call: 'update' | 'finalize'; sessionId: string; text?: string };
+
+/** Records store writes in order, so tests can assert on row lifecycle. */
+const createSink = (writes: Write[]): StreamRowSink => ({
+  update: (sessionId, text) => writes.push({ call: 'update', sessionId, text }),
+  finalize: (sessionId) => writes.push({ call: 'finalize', sessionId }),
+});
+
+/** Scheduler whose pending flushes only run when a test says so. */
+const createScheduler = () => {
+  const pending = new Map<number, () => void>();
+  let nextTimer = 1;
+  const scheduler: StreamFlushScheduler = {
+    schedule: (flush) => {
+      const timer = nextTimer++;
+      pending.set(timer, flush);
+      return timer;
+    },
+    cancel: (timer) => {
+      pending.delete(timer);
+    },
+  };
+  return {
+    scheduler,
+    flushAll: () => {
+      const due = [...pending.values()];
+      pending.clear();
+      for (const flush of due) flush();
+    },
+    pendingCount: () => pending.size,
+  };
+};
+
+test('boundary kinds are the frames that carry a row of their own', () => {
+  for (const kind of ['text', 'thinking', 'tool_use', 'tool_result']) {
+    assert.equal(isStreamBoundaryKind(kind), true, kind);
+  }
+  for (const kind of ['stream_delta', 'status', 'complete', 'error', undefined]) {
+    assert.equal(isStreamBoundaryKind(kind), false, String(kind));
+  }
+});
+
+test('deltas are coalesced into one write per flush', () => {
+  const writes: Write[] = [];
+  const buffers: StreamBuffers = new Map();
+  const { scheduler, flushAll, pendingCount } = createScheduler();
+
+  appendStreamDelta(buffers, 's1', 'Hel', createSink(writes), scheduler);
+  appendStreamDelta(buffers, 's1', 'lo ', createSink(writes), scheduler);
+  appendStreamDelta(buffers, 's1', 'there', createSink(writes), scheduler);
+
+  assert.equal(pendingCount(), 1);
+  assert.deepEqual(writes, []);
+
+  flushAll();
+  assert.deepEqual(writes, [{ call: 'update', sessionId: 's1', text: 'Hello there' }]);
+});
+
+test('a boundary finalizes the row so the next message does not replace it', () => {
+  const writes: Write[] = [];
+  const sink = createSink(writes);
+  const buffers: StreamBuffers = new Map();
+  const { scheduler, flushAll } = createScheduler();
+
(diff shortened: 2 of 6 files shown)


blackmammoth:
hey @MannXo, I think this should be fixed now with the latest commits in main. You can check and if issue still persists, you can submit a new PR with merge conflicts resolved.

PR #1158, opened 2026-08-16
fix(chat): resolve the subagent tool under both Task and Agent

mcgm51:
## The problem

Claude Code renamed its subagent tool from `Task` to `Agent`. Six places in this
codebase key off the literal string `'Task'`:

- `getToolConfig` lookup in `toolConfigs.ts`
- subagent-container detection in `useChatMessages.ts`
- the icon category in `ToolRenderer.tsx`
- the container's own switch in `SubagentContainer.tsx`

With no config matching `'Agent'`, the tool falls through to the `Default`
renderer — which titles the block **"Parameters"** and dumps the raw input as
JSON. For the subagent tool that input is *the entire prompt sent to the
subagent*, rendered inline, so spawning an agent looks like an enormous message
the user typed themselves.

The failure mode is what makes this easy to miss: **a missing tool config is not
an error.** Nothing throws, nothing logs. The UI just silently gets worse, and
each of the six sites regressed independently.

## The fix

One alias map, applied at every name-based decision, rather than an
`|| 'Agent'` sprinkled per call site — so the next rename is a one-line change
in one file instead of another hunt through six.

**Both names keep working.** A transcript recorded before the rename still says
`Task`; one recorded after says `Agent`. Old conversations must keep rendering.

## What's included

| File | Change |
|---|---|
| `tools/toolAliases.ts` | New. The alias map, `canonicalToolName`, `isSubagentTool` |
| `tools/toolAliases.test.ts` | New. Regression test |
| `tools/configs/toolConfigs.ts` | Resolve through the alias map |
| `tools/ToolRenderer.tsx` | Icon category via `isSubagentTool` |
| `tools/components/SubagentContainer.tsx` | Alias-aware switch |
| `hooks/useChatMessages.ts` | Container detection via `isSubagentTool` |
| `package.json` | One line: add `src/**` test globs so the new test runs |

+123/−7.

## About the test

It asserts the tool resolves to the **rich** config, not merely that it resolves
to something — because "falls through to Default" is exactly the bug, and a test
that only checked for a non-null result would have passed throughout.

`package.json` currently globs only `server/**` for tests, so the client test
would not run without the added `src/**` globs. Happy to drop that hunk and the
test file if you would rather keep the test scope as it is.

## Cross-provider impact — one open question for you

This repo supports four providers, so I checked what the alias map does to each
rather than only to Claude. **`grep` for tools literally named `Task` or `Agent`
outside the Claude provider:**

| Provider | Emits | Effect of this PR |
|---|---|---|
| Claude | `Agent` (was `Task`) | **Fixed** — the point of the PR |
| Codex | `Task` (`codex-sessions.provider.ts:356,403`) | **None.** `Task` is the canonical name, so it maps to itself and keeps rendering exactly as today |
| Cursor | neither | None |
| OpenCode | **`Agent`** (`opencode-sessions.provider.ts:491`) | **Behaviour changes — please sanity-check this one** |

**The OpenCode case.** It emits `toolName: partType === 'patch' ? 'Patch' :
'Agent'`. Today that `Agent` matches no config and falls through to `Default`,
so OpenCode users see the raw JSON dump. With this PR it resolves to the
subagent container instead.

That is probably an improvement — it is an agent invocation, and the container
is the right shape for one. But I have not verified OpenCode's semantics, and
**`SubagentContainer` was written against Claude's `Task` input shape**
(`prompt`, `subagent_type`, …). If OpenCode's `partData` carries a different
shape, the container could render thin where the JSON dump at least showed
everything. You will know immediately whether that is a win or a regression.

If it is unwelcome, the fix is one line — make `isSubagentTool` provider-aware,
or narrow the alias to the Claude provider only. Happy to make that change.

## Notes

Verified on a clean checkout of `main`, not on my own tree:

- `npm run build` passes.
- The new test passes 6/6 — and **fails 4/6 with the `toolConfigs` change
  reverted**, so it guards the regression rather than its own scaffolding.

Running in a downstream fork.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat rendering for subagent tools using either the current or legacy tool name.
  * Preserved specialized tool displays and meaningful subagent titles for aliased tools.
  * Prevented aliased tools from falling back to generic parameter-only displays.

* **Tests**
  * Added regression coverage for subagent aliases, rendering, titles, and unknown tool handling.
  * Expanded the test command to discover client-side TypeScript test files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "postbuild:server": "node scripts/promote-dist-server.mjs",
     "preview": "vite preview",
     "typecheck": "tsc --noEmit -p tsconfig.json && tsc --noEmit -p server/tsconfig.json",
-    "test": "tsx --tsconfig server/tsconfig.json --test \"server/**/*.test.ts\" \"server/**/*.test.js\"",
+    "test": "tsx --tsconfig server/tsconfig.json --test \"server/**/*.test.ts\" \"server/**/*.test.js\" \"src/**/*.test.ts\" \"src/**/*.test.tsx\"",
     "lint": "eslint src/ server/",
     "lint:fix": "eslint src/ server/ --fix",
     "start": "npm run build && npm run server",
--- a/src/components/chat/hooks/useChatMessages.ts
+++ b/src/components/chat/hooks/useChatMessages.ts
@@ -5,6 +5,7 @@
 
 import type { NormalizedMessage } from '../../../stores/useSessionStore';
 import type { ChatMessage, SubagentChildTool } from '../types/types';
+import { isSubagentTool } from '../tools/toolAliases';
 import { decodeHtmlEntities, unescapeWithMathProtection, formatUsageLimitText } from '../utils/chatFormatting';
 
 function formatToolResultContent(content: unknown): string {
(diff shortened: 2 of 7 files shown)


blackmammoth:
hey @mcgm51, I believe this is addressed now. Check main branch and if issue still persists, please send a new PR with updated merge conflicts.

PR #1162, opened 2026-08-18
fix(i18n): add missing delete-dialog archive keys to all locales

Wasabi81-code:
### What's wrong

`SidebarModals.tsx` renders five translation keys that exist in **none** of the
11 locale catalogs:

| Key | Used at |
|---|---|
| `deleteConfirmation.archiveProject` | project dialog, safe action |
| `deleteConfirmation.archiveSession` | session dialog, safe action |
| `deleteConfirmation.archiveSessionNotice` | session dialog, explanatory text |
| `deleteConfirmation.archivedSessionNotice` | session dialog, explanatory text |
| `deleteConfirmation.deleteSessionPermanently` | session dialog, destructive action |

Each call site passes an inline English default, e.g.

```jsx
{t('deleteConfirmation.archiveProject', 'Archive project')}
```

so nothing ever reports a missing key at runtime — the dialog just silently
renders English in every language.

### Why it matters more than a normal missing string

Both dialogs exist to present **one safe action next to one destructive
action** — and the two dialogs fail differently.

**The project dialog renders half-translated, in all 11 languages.** Its
destructive button uses `deleteConfirmation.deleteAllData`, which *is*
translated everywhere, while the safe button falls back to English:

| Locale | safe button | destructive button |
|---|---|---|
| `de` | Archive project | Alle Daten dauerhaft löschen |
| `ja` | Archive project | すべてのデータを完全に削除 |
| `ru` | Archive project | Удалить все данные навсегда |

(…and the same in the other eight.)

**The session dialog renders entirely in English**, in every locale, because
both of its button keys are missing.

The half-translated pair is the worse of the two: it undercuts exactly the
distinction the dialog is there to make, and it does so at the expense of the
*safe* option, which ends up looking foreign and technical next to a fluent
warning. The destructive path removes the project row, its session rows and the
`*.jsonl` transcript files from disk; the safe path only sets `isArchived` and is
reversible via `restoreArchivedProject`. That difference should read equally
clearly in every language.

### Cause

These are the post-refactor names of strings that *are* still translated under
their old keys — the rename came in with #715. Every catalog still carries `removeFromSidebar`
("Remove from sidebar only"), `allConversationsDeleted` and `cannotUndo`, none of
which is referenced by any component any more. The keys were renamed in the
components; the catalogs were never updated.

### What this PR does

Adds the five keys to all 11 locales (`en, de, es, fr, it, ja, ko, ru, tr,
zh-CN, zh-TW`), following the pattern of #896. Locale JSON only — no component
or logic changes.

The three now-unreferenced keys are **left untouched** to keep this change
purely additive. Happy to remove them in a follow-up if you'd like.

### Out of scope

A repo-wide sweep would be much larger: roughly 110 `t(key, 'English default')`
call sites currently have no catalog entry, about half of them under
`src/components/sidebar` (the whole `archived.*` namespace is absent too). This
PR deliberately fixes only the destructive-confirmation dialogs, where the
missing translations change what the user understands they are about to do.
Glad to follow up on the rest if that's wanted.

### How to reproduce

1. Start the app and switch the language to any non-English locale
   (Settings → language), e.g. German or Japanese.
2. In the sidebar, open a project's context menu and choose to remove it.
3. The confirmation dialog shows a translated headline, a translated
   *"Delete all data permanently"* button — and an untranslated English
   *"Archive project"* button above it.
4. Same in the session delete dialog: *"Archive session"*, *"Delete permanently"*
   and the explanatory sentence are all English regardless of locale.

With this patch every element of both dialogs renders in the selected language.

### Prior art checked

No open issue or PR adds these keys; the only other hits for `archiveProject`
and `deleteConfirmation` are closed PRs, #715 among them.

### Checking it without running the app

Resolving each `t()` call the way i18next does — catalog value if the key
exists, otherwise the inline default — against the real component and the real
catalogs reproduces the tables above for `main`, and shows every cell served
from the catalogs on this branch, for both dialogs and all 11 locales.

### Verification

All five `en` values are byte-identical to the inline fallbacks already in the
components, so English rendering is provably unchanged — `en/sidebar.json` is the
only file whose content needs review, the other ten are additive.

`npm run lint`, `npm run typecheck`, `npm run build` and the server test suite
(262/262) pass on this branch. All 11 JSON files re-serialize byte-identically
apart from the added keys (2-space indent, trailing newline preserved).

## Summary by CodeRabbit

* **New Features**
  * Added localized options to archive projects and sessions.
  * Added notices explaining that archived sessions are hidden while their history is retained.
  * Added messaging for sessions that have already been archived.
  * Added a permanent session deletion action label.

* **Localization**
  * Added translations across English, German, Spanish, French, Italian, Japanese, Korean, Russian, Turkish, Simplified Chinese, and Traditional Chinese.


blackmammoth:
hey @Wasabi81-code, can you resolve merge conflicts?

Wasabi81-code:
Rebased onto current `main` — the branch is conflict-free now.

The divergence wasn't in the translations themselves: upstream moved
`src/i18n/locales/` to `src/modules/i18n/locales/`, and this branch was
still writing to the old path. The rebase re-applies the five
`deleteConfirmation` archive keys at the new location, unchanged; nothing
else is touched.

All 11 locales carry all 5 keys, and `SidebarModals.tsx` on `main` already
references them (`archiveProject`, `archiveSession`, `archiveSessionNotice`,
`archivedSessionNotice`, `deleteSessionPermanently`) with English fallbacks,
so this only replaces those fallbacks with real translations.

blackmammoth (approved):

PR #1164, opened 2026-08-18
feat(settings): close settings modal with escape and backdrop click

MiguelBel:
## What

The settings modal can now be dismissed with the <kbd>Esc</kbd> key or by clicking the backdrop. Until now its X button was the only way out, which is inconsistent with the rest of the app — `ConfirmActionModal`, `NewBranchModal`, `NewWorktreeModal` and `FolderBrowserModal` all close on Escape already.

## Why the extra plumbing

Two details make this more than a one-line `keydown` listener:

**1. Escape has to peel one layer at a time.** Several things stack on top of settings: the provider login modal, the MCP server form modal (which portals to `document.body`), and `ActionMenu` dropdowns. `ActionMenu` already closes itself on Escape via a `document` listener, so a naive handler on settings meant a single Escape dismissed the dropdown *and* the settings modal underneath it — and the MCP form modal would have been torn down with a half-filled form.

Those three now mark their overlay with `data-escape-layer`, and settings ignores Escape while any such layer is in the DOM. One Escape closes the dropdown; the next closes settings. A DOM-subtree check wouldn't have worked here, since the MCP form modal renders through a portal.

**2. A backdrop click must start *and* end on the backdrop.** Selecting text inside the panel and releasing the mouse over the dimmed area fires a `click` whose target is the backdrop. Closing on that would throw away whatever the user was editing, so the handler tracks where the press started.

## Manual QA

With the settings modal open:

| Action | Expected |
| --- | --- |
| Press <kbd>Esc</kbd> | Modal closes |
| Click the dimmed backdrop | Modal closes |
| Click anywhere inside the panel | Stays open |
| Select text inside the panel, release over the backdrop | Stays open |
| Open the "Add MCP Server" dropdown, press <kbd>Esc</kbd> | Only the dropdown closes |
| Press <kbd>Esc</kbd> again | Modal closes |
| Open the MCP server form, press <kbd>Esc</kbd> | Both stay open (form modal keeps its own behaviour) |

`npm run typecheck` and `npm test` (262 passing) are green.

## Out of scope

`ProviderLoginModal` and `McpServerFormModal` still don't close on their own Escape — this PR only makes sure settings doesn't steal that keypress from them. Giving each modal its own Escape handling seems worth a separate PR. The settings panel also still lacks `role="dialog"` / `aria-modal`, which the git-panel modals have.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Escape-key handling for settings, modals, and action menus to support layered overlays correctly.
  * Settings now closes from backdrop clicks only when the click both starts and ends on the backdrop, preventing accidental closure when interacting with contained content.


blackmammoth:
hey @MiguelBel, i like this, it's very good in terms of UX. Can u update the merge conflicts before it gets reviewed?

MiguelBel:
@blackmammoth Hi, the PR is ready without conflicts.

blackmammoth (approved):

PR #1166, opened 2026-08-18
fix(sidebar): let Running now collapse its project groups

materemias:
## Problem

In the sidebar's Running view, every project group is forced open. Clicking its chevron cannot collapse it, but still updates the separate Projects-view accordion state.

## Change

- Keep Running groups expanded by default, while allowing each group to be collapsed and reopened independently.
- Track Running collapses separately so those clicks do not change the Projects-view accordion. Projects retains its existing single-open behavior.
- Replace `expandedProjects` and `forceExpanded` row props with `isProjectExpanded(projectId)`, shared by the chevron and session-list rendering.
- Rename callback parameter labels from `projectName` to `projectId` to match the values already passed. This is a type-documentation correction, not a change to project starring or renaming.

The collapse state is held in memory for the mounted sidebar; this PR adds no persisted preference.

## Verification

Rebased onto upstream `main` at `5e73a49b89` as requested. The PR remains one feature commit, `3695bb33`, changing six files.

Checks passed on that rebased head:

- `npm run typecheck`
- `npx vitest run src/modules/sidebar/tests/sidebarRowProps.test.tsx` — 7 tests passed, including per-project expansion rendering across a rerender.
- `npm run build:client` — passed with CSS-minification and chunk-size warnings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sidebar project expansion behavior: running projects can be expanded or collapsed independently, while other project views retain their existing behavior.
  * Deleted sessions are now removed from Recent Conversations immediately, with totals updated without requiring a refresh.
  * Renamed sessions now display their updated titles in Recent Conversations immediately.

* **Improvements**
  * Updated project and session actions to use stable project identifiers for more consistent sidebar interactions.
  * Added support for displaying compaction status and summaries in chat messages.

### DIFF
--- a/src/modules/sidebar/Sidebar.tsx
+++ b/src/modules/sidebar/Sidebar.tsx
@@ -80,7 +80,7 @@ function Sidebar({
 
   const {
     isSidebarCollapsed,
-    expandedProjects,
+    isProjectExpanded,
     activeRename,
     showNewProject,
     initialSessionsLoaded,
@@ -191,7 +191,7 @@ function Sidebar({
     selectedSession,
     isLoading,
     loadingProgress,
-    expandedProjects,
+    isProjectExpanded,
     activeRename,
     initialSessionsLoaded,
     currentTime,
@@ -202,7 +202,6 @@ function Sidebar({
     loadingMoreProjects,
     activeSessions,
     attentionSessionIds,
-    forceExpanded: searchMode === 'running',
     isProjectStarred,
     onRenameDraftChange: updateRenameDraft,
     onToggleProject: toggleProject,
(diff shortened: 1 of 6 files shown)


blackmammoth:
hey @mate, can you rebase this branch on to main and it'll be reviewed.

materemias:
@blackmammoth Rebased onto latest main (5e73a49b89), as requested. The new head is 3695bb33, still one feature commit. Typecheck, all 7 sidebar row tests, and the frontend build pass. I also updated the description to match the rebased diff. Ready for your review.

materemias:
Thanks for flagging this. The finding is valid: the nested state updater can decrement the Recent Conversations total more than once, and deleting a session outside the loaded page leaves the total stale. This code was introduced in 7015ffc8 (#1157) and is already present in this PR’s base; #1166 does not modify the deletion path. I’m leaving it unchanged here to keep this PR focused on Running-group collapse. The deletion-count fix should be handled separately.

blackmammoth (approved):

PR #1167, opened 2026-08-18
fix(chat): stop scrolled-away messages rendering as tall empty boxes

materemias:
`content-visibility: auto` sizes an off-screen element by
`contain-intrinsic-size` rather than by its content, and the reserved sizes were
guesses: a collapsed tool row is about 32px and was reserving 240px, so
messages scrolled out of view became tall empty boxes with their text floating
inside them.

The `auto` keyword makes it worse rather than better - the browser then
remembers the last rendered height, so a row expanded once (a diff, a long tool
output) keeps that height forever while skipped. Messages are already
paginated, so the DOM stays small and skipping off-screen work buys little.

Keep the paint/layout isolation, drop the sizing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat message rendering consistency and reduced blank or incomplete content in transcripts.
  * Improved printed transcript reliability so chat messages render correctly across pages.
  * Preserved efficient lazy loading and layout behavior while improving message display stability.

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -588,20 +588,13 @@
     clip-path: inset(-8px -8px 0 -8px);
   }
 
-  .chat-message {
+  /*
+   * LazyMessageRow owns off-screen sizing through measured placeholders.
+   * Containing its stable wrapper without size containment or content visibility
+   * keeps those placeholder heights and mounted content heights authoritative.
+   */
+  .chat-message-row {
     contain: layout style paint;
-    content-visibility: auto;
-    contain-intrinsic-size: auto 180px;
-  }
-
-  .chat-message.assistant {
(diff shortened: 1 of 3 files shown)


blackmammoth:
Closed as this didn't have a clear reproduction step of the issue.

materemias:
Fair close — the description had no reproduction step, which `CONTRIBUTING.md` requires. Reopened as #1280 with one: deterministic steps, before/after `scrollTop`/`scrollHeight` numbers, and a DevTools snippet that prints the placeholder heights so you can see the `96`/`240` intrinsic-size fallbacks that match no rendered row. Same commit, `e85f4c63`, no code change.

PR #1168, opened 2026-08-18
fix(chat): stop escape-unescaping from shredding LaTeX and currency

materemias:
The escape pass rewrote `\t`, `\n` and friends across the whole message, so a
`\text{…}`, `\theta` or `\rho` inside an equation became a tab, a newline or a
carriage return. Its math guard only protected `$$…$$` and single-`$` spans, and
the single-`$` rule did the opposite damage: `$200k … $100k` in prose was
swallowed as one equation, hiding the text between two amounts.

Protect code spans and real math regions instead — fenced and inline code,
`$$…$$`, `\[…\]` and `\(…\)` — and rewrite the LaTeX delimiters to dollars,
since remark-math only understands those. Single `$` is no longer a math
delimiter, so `singleDollarTextMath` is disabled in the Markdown renderer to
match. Placeholders are NUL-prefixed and widened on collision, so a message
containing the placeholder text cannot have its own words replaced.

Covered by chatFormatting.test.ts: LaTeX survival, currency prose, code-block
escapes, and the placeholder-collision case.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LaTeX rendering in chat messages, including bracket and parenthesis formats.
  * Prevented currency values and LaTeX-like text in code blocks from being misinterpreted as math.
  * Kept escaped LaTeX delimiters as literal text.
  * Ensured completed and streaming responses render mathematical notation consistently.

* **Tests**
  * Added coverage for LaTeX normalization, currency text, code blocks, escaped delimiters, and streaming content.

### DIFF
--- a/src/modules/chat/tests/chatFormatting.test.ts
+++ b/src/modules/chat/tests/chatFormatting.test.ts
@@ -1,8 +1,13 @@
+import React from 'react';
+
+import { render } from '@testing-library/react';
 import assert from 'node:assert/strict';
 
 import { test } from 'vitest';
 
 import { stripProposedPlanEnvelope } from '@/modules/chat/utils/chatFormatting';
+import { Markdown } from '@/modules/chat/transcript/Markdown';
+import StreamingMarkdown from '@/modules/chat/transcript/StreamingMarkdown';
 
 test('stripProposedPlanEnvelope removes a complete outer plan envelope', () => {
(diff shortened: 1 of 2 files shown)


blackmammoth:
I think #1244 is trying to address this in a general way, so I'll be closing this one.

PR #1174, opened 2026-08-19
fix(auth): stop a cached API response from restoring an expired token

materemias:
## Problem

An installed PWA is logged out on almost every cold start, with "Your session expired", while a long-lived browser tab on the same server never is.

The cause is HTTP caching of `/api` responses:

- Express sends an `ETag` on JSON responses but no `Cache-Control`, so `/api/auth/user` is cacheable and ETag-revalidated.
- `authenticateToken()` only attaches `X-Refreshed-Token` when the token is past half-life, so a revalidated `304` usually carries no such header.
- The browser then merges the `304` onto the **cached** response, and JS reads the `X-Refreshed-Token` stored with it, which can be days old.
- `storeAuthToken()` validated only that the value was JWT-shaped, so it persisted that stale token over the live one.
- The next `getStoredAuthToken()` read it as expired, wiped the session and dispatched `auth-session-expired`.

Measured against a running server:

```
request 1 : 200 | etag: W/"6b-..." | cache-control: null | X-Refreshed-Token: present
request 2 : 304 |                                        | X-Refreshed-Token: ABSENT
```

A cold start replays those requests against a warm cache. A tab that stays open avoids that cold-start check, but token changes can also re-run authentication checks in an already-open tab.

Instrumenting the shipped bundle on the affected device confirmed the sequence within a single launch: boot with a token valid for 7 more days, then a write of a token that had expired 4 days earlier, then the wipe.

[TadMSTR independently reproduced the cached-header condition](https://github.com/siteboon/claudecodeui/pull/1174#issuecomment-5343245708) and reported that alternating old and new tokens also triggered a WebSocket reconnect storm. Their comparison with the token-ordering guard stopped the alternation. They opened #1177 to address the token-dependent auth-check loop that amplifies it.

## Fix

- Serve `/api` with `Cache-Control: no-store`, so per-session headers are not stored in HTTP caches.
- Force `cache: 'no-store'` in `authenticatedFetch`, even if a caller requests caching. This also bypasses caches already holding a poisoned entry.
- `storeAuthToken()` requires readable, finite `iat` and `exp` claims. It rejects an older `iat`, and rejects a different token with the same `iat` as the stored token. The identical stored token remains acceptable.
- Expiry is not a storage rejection criterion: a newer token is accepted even if already expired. `getStoredAuthToken()` retains responsibility for local expiry handling.

## Tests

`src/shared/tests/authToken.test.ts` adds coverage for older tokens, conflicting equal-`iat` tokens, identical tokens, a newer-but-expired token, and unreadable claims. `src/shared/tests/authenticatedFetch.test.ts` covers forced cache bypass and a slower older refresh response arriving after a newer one.

The current diff does not change the client test glob or `import.meta.env` handling. Those changes and the old `src/utils/api.test.js` path belonged to the original version of this PR.

The original submission reported clean typecheck, lint with no new findings, 270 server tests and 56 client tests. Those results predate the current posted diff and are not current-head verification.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -25,6 +25,7 @@ import {
     authenticateWebSocket,
     authRoutes,
     validateApiKey,
+    preventApiResponseCaching,
 } from './modules/auth/index.js';
 import { taskmasterRoutes } from './modules/taskmaster/index.js';
 import { commandsRoutes } from './modules/commands/index.js';
@@ -143,6 +144,8 @@ app.get('/health', (req, res) => {
         version: RUNNING_VERSION
     });
 });
+
+app.use('/api', preventApiResponseCaching);
 
 // Optional API key validation (if configured)
 app.use('/api', validateApiKey);
--- a/server/modules/auth/auth.middleware.ts
+++ b/server/modules/auth/auth.middleware.ts
@@ -8,6 +8,12 @@ import { userDb, appConfigDb } from '../database/index.js';
 // Use env var if set, otherwise auto-generate a unique secret per installation
 const JWT_SECRET = process.env.JWT_SECRET || appConfigDb.getOrCreateJwtSecret();
 
+/** Used by the server entrypoint so API responses carrying auth state never enter HTTP caches. */
+export const preventApiResponseCaching = (_req, res, next) => {
+  res.setHeader('Cache-Control', 'no-store');
+  next();
+};
+
 // Optional API key middleware
 const validateApiKey = (req, res, next) => {
   // Skip API key validation if not configured
--- a/server/modules/auth/index.ts
+++ b/server/modules/auth/index.ts
@@ -7,3 +7,5 @@ export { authenticateToken } from './auth.middleware.js';
 export { authenticateWebSocket } from './auth.middleware.js';
 // validateApiKey: used by the server entrypoint for optional API-wide key validation.
 export { validateApiKey } from './auth.middleware.js';
+// preventApiResponseCaching: used by the server entrypoint before API routes can emit auth state.
+export { preventApiResponseCaching } from './auth.middleware.js';
--- a/src/shared/api.ts
+++ b/src/shared/api.ts
@@ -32,6 +32,8 @@ export const authenticatedFetch = (
 
   return fetch(url, {
     ...options,
+    // Auth responses carry session headers, so callers cannot opt back into caching.
+    cache: 'no-store',
     headers: {
       ...defaultHeaders,
       ...options.headers,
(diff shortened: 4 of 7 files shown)


TadMSTR:
Independent corroboration, plus a second symptom this fixes that the description doesn't mention.

I chased a websocket reconnect storm — thousands of full `connect → authenticate → connected → disconnected` cycles on `/ws`, tens of minutes at a time, self-resolving. Three occurrences captured server-side; the largest was **53,571 cycles at ~20/sec sustained for ~40 minutes**. It traced back to the same cached-`X-Refreshed-Token` condition you measured here.

**The response shape reproduces.** Against a running server, `/api/auth/user` returns an `ETag` and **no `Cache-Control`** — cacheable and revalidated, exactly as you describe.

**One detail worth adding:** `generateToken` stamps `iat` with one-second resolution, so reissues within the same second are byte-identical. I sent 25 rapid `/api/auth/user` requests with a past-half-life token: `X-Refreshed-Token` came back on **25/25, but with only one distinct value**. That matters because it means a token that only ever moves *forward* is self-limiting — `setToken` bails on equality. Your bug needs the token to move *backwards*, which is precisely what the cached header does. It also explains why this is intermittent rather than constant.

**Why it also storms.** `checkAuthStatus` takes `token` as a `useCallback` dependency and the mount effect depends on that callback, so each token change re-runs the full session verification — three more authenticated requests, which can change the token again. Once a cached header makes the value alternate, that becomes an unbounded loop. Each pass calls `setUser()` with a freshly parsed object and toggles `isLoading`, and `WebSocketContext`'s connect effect depends on `[isAuthLoading, token, user]`, so the socket is torn down and reopened every pass.

Measured with the real providers mounted against a real server, 8-second window:

| | verification requests | websocket connections |
|---|---|---|
| token only moving forward | 6 | 2 |
| token alternating | 46,997 | 15,044 |
| token alternating, **with this PR's `storeAuthToken` guard** | **6** | **2** |

So this PR fixes the reconnect storm too, not just the PWA cold-start logout — the `iat` check rejects the older cached token and the alternation stops. Worth noting in the description, since anyone searching for the reconnect symptom won't currently find their way here.

I've opened #1177 for the amplifier — making `checkAuthStatus` independent of `token` so that no future source of token churn can turn into a request flood. It's complementary, not competing: yours removes the trigger, that one removes the multiplier, and either alone resolves the incident I saw.

Small heads-up on a conflict: #1177 optional-chains `import.meta.env` in `src/shared/utils.ts`, the same one-line change this PR makes, because otherwise any test importing client code throws. Trivial to resolve whichever lands second.

materemias:
this is still relevant today, anything blocking it @blackmammoth ?

PR #1175, opened 2026-08-19
fix(auth): keep the session when auth cannot be verified

materemias:
## Problem

`checkAuthStatus()` runs on mount and again when its callback dependencies change, including the token. It previously ended the session on **any** non-2xx response from `/api/auth/user`:

```js
const userResponse = await api.auth.user();
if (!userResponse.ok) { clearSession(); return; }
```

`!ok` is true for a `502`, a `503` and a raw `304`, not just a real rejection. An installed PWA executes the check on every cold start, so reopening the app during a server restart can delete a perfectly valid token. A warm tab can also re-run the check when its token changes.

Measured through a reverse proxy in front of the server, the restart window returns `502` about 500 ms in, which is exactly the interval a PWA launch can land in.

A thrown request was mishandled in the other direction: the catch left `user` null with loading finished, and `ProtectedRoute` renders `LoginForm` for any `!user`, so an offline resume *looked* signed out even though the token was intact. Neither failure path scheduled its own retry, so restoring the network alone did not recover the cold-start check.

By contrast `refreshSession()` already tolerates failure, with the comment "A transient network failure must not sign the user out". This brings the initial check in line with it.

## Fix

- `classifyAuthProbe()` distinguishes `authenticated` / `rejected` / `unavailable`. A successful HTTP response is classified as authenticated; a non-success response is rejected only when it carries `X-Auth-Error`.
- `X-Auth-Error` is the rejection discriminator, not the status code. A bare `401`/`403` without it does not establish that this application's token check rejected the session, so the probe treats it as unavailable rather than deleting the token.
- Non-success responses without `X-Auth-Error`, an unreadable or missing user payload, and thrown requests preserve the held token. On a cold start with no loaded user, `authUnavailable` selects a "still signed in, reconnecting" screen instead of the login form. Local token expiry and explicit authentication rejection still end a session.
- While a token is held, auth is unavailable, and no user has loaded, silent retries run every 5 s, on `online`, and when the document becomes visible. The screen also has a manual retry button. Retry triggers share an in-flight promise, so another trigger waits for that retry to settle. An already-loaded user remains in the app or onboarding flow and does not arm this unavailable-auth retry loop.
- Probe IDs discard superseded results, and the rejection branch compares the probed token with the stored token before clearing. `authenticatedFetch` also checks token ownership and whether a later-started request has already succeeded before expiring a session. It derives the bearer token from the merged `Authorization` header rather than assuming every request uses the stored credential.
- `resolveAuthView()` makes screen precedence a single pure function rather than nested JSX conditionals.

## Tests

`src/modules/auth/tests/authState.test.tsx` covers classification, view precedence, token-matched rejection, preserving tokens/preferences/drafts on network failure, clearing them on rejection, superseded probes, retry triggers, and sharing a slow in-flight retry. `src/shared/tests/authenticatedFetch.test.ts` adds cases for a late rejection after a newer successful request and a caller's `Authorization` override carrying a different token.

The original submission reported browser verification against the production bundle with `/api/*` synthesised via request interception on a throwaway origin: a `502` cold start and an outright network failure kept the token and showed the offline screen; recovery happened unprompted 5 s later; a `401` authentication rejection still cleared the session. These are historical results, not verification of the current posted head.

The original submission also reported clean typecheck, lint with no new findings, 270 server tests and 59 client tests. Those counts predate the current posted diff.

### DIFF
--- a/src/modules/auth/AuthUnavailableScreen.tsx
+++ b/src/modules/auth/AuthUnavailableScreen.tsx
@@ -1,0 +1,25 @@
+import AuthScreenLayout from '@/modules/auth/AuthScreenLayout';
+
+type AuthUnavailableScreenProps = {
+  onRetry: () => void;
+};
+
+export default function AuthUnavailableScreen({ onRetry }: AuthUnavailableScreenProps) {
+  return (
+    <AuthScreenLayout
+      title="CloudCLI"
+      description="Cannot reach the server. You are still signed in, and CloudCLI will reconnect on its own."
+      footerText="Your stored session has been kept."
+    >
+      <div role="status" aria-live="polite">
+        <button
+          type="button"
+          onClick={onRetry}
+          className="flex w-full items-center justify-center rounded-xl bg-primary px-4 py-2.5 font-medium text-primary-foreground shadow-lg shadow-primary/25 transition-all duration-200 hover:shadow-primary/30 hover:brightness-110 focus:outline-none focus:ring-2 focus:ring-primary/40 focus:ring-offset-2 focus:ring-offset-card active:scale-[0.99]"
+        >
+          Try again
+        </button>
+      </div>
+    </AuthScreenLayout>
+  );
+}
--- a/src/modules/auth/ProtectedRoute.tsx
+++ b/src/modules/auth/ProtectedRoute.tsx
@@ -4,6 +4,7 @@ import { IS_PLATFORM } from '@/shared/utils';
 import { useAuth } from '@/modules/auth/context/AuthContext';
 import { Onboarding } from '@/modules/onboarding';
 import AuthLoadingScreen from '@/modules/auth/AuthLoadingScreen';
+import AuthUnavailableScreen from '@/modules/auth/AuthUnavailableScreen';
 import LoginForm from '@/modules/auth/LoginForm';
 import SetupForm from '@/modules/auth/SetupForm';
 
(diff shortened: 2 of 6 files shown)


materemias:
All three review findings are addressed in `973dfab`, `63f01df` and `94920ba`.

**Overlapping probes (CodeRabbit).** Confirmed, and it was worse than the analysis showed: `AuthContext` was not even the last word. `authenticatedFetch` calls `expireAuthSession()` on any `X-Auth-Error`, which deletes the token and dispatches `auth-session-expired` before the caller can apply any freshness check, so guarding only `checkAuthStatus` would have left the race open.

Both layers are now guarded:

- `checkAuthStatus` takes a monotonic id that `setSession` and `clearSession` also bump. Only the newest run may touch session state.
- A rejection in `authenticatedFetch` has to still apply: the token the request presented must be the one in storage, and no request that started later may have already come back authorised. A request that carried no token may only end a still-empty session.

Verified in a browser against the production bundle with `/api/*` synthesised through request interception, replaying the exact ordering: a `502` cold start arms the retry, the probe at 5 s is held for 9 s and answers `401` with `X-Auth-Error`, the probe at 10 s answers `200`. With the fix the app is still signed in at 18 s with the token intact; with the same run against the branch tip before these commits, the stale `401` at 14.5 s deletes the token and renders the login form.

**Catch with no stored token (Copilot).** Right — nothing was being kept there, and the retry effect stays disarmed without a token, so the "your session is kept" message was on the login form for someone who was never signed in. That path now sets the plain network error.

**Non-exhaustive switch (Copilot).** Variants are spelled out and the default asserts `never`.

`npm run typecheck` clean, `npm run test:client` 59 pass, `npm run build` clean.

materemias:
Fixed in `6a6f114`.

**Rejection clearing a refreshed session.** Valid, and the reasoning is exactly right: `handleTokenRefreshed` stores a token without moving the probe id, so the id alone cannot separate "this rejection is about the token in storage" from "a refresh replaced that token while the probe was in flight". The token is the discriminator there, so `rejectionEndsSession(sentToken, storedToken)` now gates the branch, which is the same rule `authenticatedFetch` applies to the same race one layer out. It is a pure function with unit coverage: identical tokens end the session, a replaced token does not, and neither does a probe that presented nothing.

**On serializing probes instead.** I went with freshness rather than an in-flight lock deliberately. A lock makes one stuck request block every later one, and that is not hypothetical: while replaying this scenario I held a probe open and the app sat on "Loading authentication state…" for the whole run, because nothing else could be applied. Discarding stale answers keeps the newest verdict authoritative and leaves recovery working even when a request never returns. Accumulation is bounded by the 5 s interval and two requests per probe, and every answer but the newest is dropped before it can touch state.

`npm run test:client` 62 pass, `npm run typecheck` clean, `npx eslint` 0 errors on the changed files, `npm run build` clean.

PR #1176, opened 2026-08-19
fix(database): never rotate persisted secrets on a failed read

materemias:
## Problem

`appConfigDb.get()` swallowed every error into `null`:

```ts
try { /* SELECT value FROM app_config WHERE key = ? */ }
catch { return null; }  // "so early-startup reads (e.g. JWT secret) do not crash"
```

`getOrCreateJwtSecret()` therefore could not tell *this key has no row* from *this read did not complete*. A transient read failure could cause it to mint a new secret and upsert it over the good one if the subsequent write succeeded. Every JWT signed with the old secret then fails signature verification, invalidating sessions that used it. The upsert preserves `created_at`, so that timestamp does not reveal the rotation.

Two things widen the window:

- `getConnection()` assigned the module singleton **before** running `exec(APP_CONFIG_TABLE_SCHEMA_SQL)`. Once that DDL failed, the cached half-initialised handle was still handed to the writer, so the read could fail while the write succeeded, which is exactly the ordering that completes a rotation.
- The connection did not explicitly configure `busy_timeout` or request WAL mode for concurrent access by the server, CLI and dev watcher.

`browser-use`'s `getOrCreateMcpToken()` had the identical shape and could silently rotate the token that authenticates the registered MCP server entry.

Reproduced against a copy of a real database, with a write lock held during startup: the stored secret changed while `created_at` stayed frozen, and nothing was logged.

## Fix

- `get()` propagates read errors. Callers that genuinely tolerate absence, such as `browser-use`'s `readSettings`, already have their own `try`/`catch`.
- New `getOrCreateSecret(key, generate)` returns an existing non-empty secret. When the row is absent, it writes with `INSERT OR IGNORE` and then re-reads, so a racing writer keeps its value and the caller gets the persisted value rather than its local candidate. A failed read, a still-absent row, or an empty stored secret throws. An empty generated candidate is rejected before insertion. Both the JWT secret and the MCP token use this helper.
- `getConnection()` publishes the singleton only after the pragmas and schema succeed, and closes the handle on failure so the next call retries cleanly.
- Set `busy_timeout = 5000` before requesting `journal_mode = WAL`. Inspect the returned journal mode; if it is not `wal`, warn and continue in the returned mode rather than refusing startup. A thrown pragma or schema error still closes the connection and propagates.

## Authentication rejection logging

The token-authentication rejection branches now log `no-token`, `user-not-found`, `session-expired`, or `invalid-token`. Expiry logs include the token's `iat` and expiration time to help investigate token age. Those timestamps alone do not prove whether a client missed a refresh or was simply inactive.

The log includes the request method, `req.baseUrl + req.path`, a bounded user agent and rejection details. It deliberately excludes the query string: SSE endpoints accept the JWT as a `?token=` query parameter. Request-derived fields pass through a helper that replaces control characters and double quotes and caps their length, limiting log injection.

## Tests

`server/modules/database/tests/app-config.test.ts` covers a secret surviving a reconnect, avoiding regeneration of an existing value, returning a racing writer's value, propagating failed reads, refusing an empty stored secret, and rejecting an empty generated candidate without persisting it. Database initialization is inside the test helper's `try` block so its cleanup also runs when initialization fails.

`server/modules/auth/tests/auth.middleware.test.ts` covers a forged user agent producing a single sanitized rejection line while retaining the 401 response and `X-Auth-Error` header.

The original submission reported clean typecheck, lint with 0 errors and 271 passing server tests, plus a replay of the rotation reproduction that left the secret intact. Later discussion reported additional test cases and higher counts. These are historical results, not verification of the current posted head.

### DIFF
--- a/server/modules/auth/auth.middleware.ts
+++ b/server/modules/auth/auth.middleware.ts
@@ -22,6 +22,15 @@ const validateApiKey = (req, res, next) => {
   next();
 };
 
+// A log line is one line: a newline, a control character or a quote coming
+// from the request would let a caller forge entries or break the `ua="..."`
+// field. Everything interpolated below goes through this first.
+const forLogLine = (value: unknown, maxLength: number): string =>
+  String(value ?? '')
+    .replace(/[\u0000-\u001F\u007F]/g, ' ')
+    .replace(/"/g, "'")
+    .slice(0, maxLength);
+
 /**
  * Records why a request was rejected with 401.
  *
@@ -34,15 +43,16 @@ const logAuthRejection = (req, reason: string, detail = ''): void => {
   // `baseUrl + path` deliberately, never `originalUrl`: SSE endpoints pass the
   // JWT as a `?token=` query parameter, which would put bearer credentials in
   // the log. `path` alone is router-relative, so it loses the endpoint.
-  const target = `${req.method} ${req.baseUrl || ''}${req.path}`;
+  const target = forLogLine(`${req.method} ${req.baseUrl || ''}${req.path}`, 200);
 
   // Every client reaches this server through the same loopback proxy, so the
   // remote address cannot tell one device from another. The user agent can,
   // which is what makes a rejection attributable to a specific browser.
-  const userAgent = String(req.headers['user-agent'] ?? 'unknown').slice(0, 120);
+  const userAgent = forLogLine(req.headers['user-agent'] ?? 'unknown', 120);
+  const safeDetail = forLogLine(detail, 200);
 
   console.warn(
-    `[Auth] 401 ${reason} ${target}${detail ? ` ${detail}` : ''} ua="${userAgent}"`
+    `[Auth] 401 ${reason} ${target}${safeDetail ? ` ${safeDetail}` : ''} ua="${userAgent}"`
   );
 };
 
--- a/server/modules/auth/auth.middleware.ts
+++ b/server/modules/auth/auth.middleware.ts
@@ -22,6 +22,30 @@ const validateApiKey = (req, res, next) => {
   next();
 };
 
+/**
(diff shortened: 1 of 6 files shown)


materemias:
Both review findings are addressed in `a1d18a0` and `5754a12`.

**Log injection (Copilot).** Correct, and the user agent is the field that matters: it is attacker-supplied, so a newline appends an invented entry and a quote closes the `ua` field early. Everything interpolated into the line now goes through one helper that replaces control characters and quotes and caps the length, so the text still shows up but cannot reframe the line. `server/modules/auth/tests/auth.middleware.test.ts` drives `authenticateToken` with a user agent carrying both and asserts one line with exactly one `ua` field.

**Empty stored value (Copilot).** Also correct. Presence is now `null` versus string. An empty value throws under its own name rather than being reported as a failed insert, because `INSERT OR IGNORE` cannot replace that row and signing with an empty secret would reject every token already issued. Covered by a new case in `app-config.test.ts`.

`npm test` 273 pass, `npm run typecheck` clean, `npm run lint` 0 errors.

materemias:
Fixed in `b02a135`.

**Empty generated candidate.** The finding is right about the consequence, which is what makes it worth guarding even though neither real caller can produce it (both generate hex from `crypto.randomBytes`). `INSERT OR IGNORE` makes any row that lands permanent, so an empty candidate would not merely fail once - it would leave a key that can never be created again, and the throw afterwards would repeat on every boot with no way out. The candidate is now checked before the statement runs, so nothing empty reaches the table.

The new test pins both halves: the call throws, the key is still absent afterwards, and a later call with a real generator still creates the secret.

`npm test` 274 pass, `npm run typecheck` clean, `npx eslint` 0 errors on the changed files.

PR #1177, opened 2026-08-19
fix(auth): verify the session once per mount, not once per token refresh

TadMSTR:
## Problem

A long-lived tab intermittently enters a websocket reconnect storm: full
`connect → authenticate → connected → disconnected` cycles on `/ws`, thousands of
them, for tens of minutes, ending on its own. Captured server-side across three
separate occurrences, the largest being **53,571 cycles at ~20/sec sustained for
~40 minutes**.

The websocket is the symptom. Underneath it the client is re-verifying its
session in a loop and flooding its own server with authenticated requests.

### The loop

`checkAuthStatus` takes `token` as a `useCallback` dependency:

```js
}, [checkOnboardingStatus, clearSession, token]);
```

and the mount effect depends on that callback:

```js
useEffect(() => { ...; void checkAuthStatus(); }, [checkAuthStatus, checkOnboardingStatus]);
```

So **every token change re-runs the full session verification** — `/api/auth/status`,
`/api/auth/user`, `/api/user/onboarding-status`. And `authenticateToken()` attaches
`X-Refreshed-Token` to *every* authenticated response once the caller's token is past
half-life (`auth.middleware.ts`), which `authenticatedFetch` pipes straight into
`storeAuthToken()` → `AUTH_TOKEN_REFRESHED_EVENT` → `setToken()`. The cycle closes: a
token change causes requests that cause a token change.

### Why it is normally invisible

`generateToken` stamps `iat` with one-second resolution, so repeated reissues **within
the same second are byte-identical** and `setToken` bails on equality. Measured against
a running server — 25 rapid `/api/auth/user` requests with a past-half-life token
returned `X-Refreshed-Token` on 25/25, but only **one distinct value**.

So while the token only ever moves *forward*, the loop converges after a pass or two.
It only becomes unbounded once something makes the token **alternate between two
values**, at which point no equality check ever fires and it runs at request speed
until the alternation stops.

That alternating condition is real and already documented in this repo: #1174 measured
`/api/auth/user` being served with an `ETag` and no `Cache-Control`, so a revalidated
`304` merges onto the cached response and JS reads an `X-Refreshed-Token` stored with
it. I confirmed the same response shape here (`ETag` present, `Cache-Control` absent).

### Why the reconnect storm is what you see

Each pass calls `setUser(userPayload.user)` with a **freshly parsed object** — new
identity even when the user is unchanged — and toggles `isLoading`. `WebSocketContext`'s
connect effect depends on `[isAuthLoading, token, user]`, and its cleanup deliberately
nulls `onclose` before `close()`, so the next effect run reconnects immediately. **The
3000ms reconnect timer is never involved**, which is why the cycles land inside a single
log timestamp and why adding backoff there does not help.

## Fix

Read the token from storage in the guard instead of from state:

```diff
-      if (!token) {
+      if (!readStoredToken()) {
```
```diff
-  }, [checkOnboardingStatus, clearSession, token]);
+  }, [checkOnboardingStatus, clearSession]);
```

`authenticatedFetch` sends whatever storage holds, so the guard now asks exactly the
question the request will answer — and the callback keeps a stable identity, so the
mount effect means what its name says. A refreshed token still updates state, the
scheduled-refresh effect and the websocket still see it; it just no longer restarts
session verification.

Measured with the real provider mounted against a real server, 8-second window:

| | verification requests | websocket connections |
|---|---|---|
| before, token only moving forward | 6 | 2 |
| before, token alternating | **46,997** | **15,044** |
| after, token alternating | **3** | **1** |

## Tests

`src/components/auth/context/AuthContext.test.tsx` mounts the real `AuthProvider`
against a stub server that keeps handing back refreshed tokens, and asserts the session
is verified exactly once. Against the unpatched provider it reports **11** (the stub
stops offering refreshes after ten, so the failure is a clean assertion rather than a
hung run).

This needs a renderer that executes effects — the existing component test uses
`renderToStaticMarkup`, which does not — so `react-test-renderer` is added as a dev
dependency. It is React-team maintained and matches the pinned React 18. **Happy to drop
the test or rework it if you would rather not take the dev dependency**; the one-line
source fix stands on its own.

`src/shared/utils.ts` optional-chains `import.meta.env` so client modules can be
imported outside Vite — without it any test reaching this code throws on import.
**#1174 makes the same one-line change**, so expect a trivial conflict with whichever
lands second.

`npm test` (270 pass), `npm run test:client` (47 pass), `npm run typecheck` and
`npm run lint` (0 errors, warning count unchanged from `main`) all clean.

## Relationship to #1174

Complementary, not competing. #1174 stops the token from moving backwards, which removes
the trigger I observed — I verified that applying its `storeAuthToken` guard alone takes
the alternating case from 15,044 websocket connections to 2. This PR removes the
amplifier, so no future source of token churn can turn into a request flood. Either one
fixes the incident; both together close it properly.

For completeness, one thing I tried that **does not work**: making the websocket effect
depend on a stable projection of `user` (`user?.username`) instead of the object. It
changes nothing, because `token` and `isAuthLoading` still churn on every pass — 16,086
connections. Mentioning it so nobody spends time there.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved authentication session handling to prevent repeated verification requests after token refreshes.
  - Reduced unnecessary loading and user-state changes during authentication.
  - Improved compatibility when environment configuration is unavailable.

- **Tests**
  - Added regression coverage for token refresh behavior and authentication request frequency.


blackmammoth:
hey @TadMSTR, I think this is not reproducible on latest main. Can you check and resubmit the PR if the issue still persists?

TadMSTR:
Thanks for looking @blackmammoth — resubmitted as #1310.

You're right that this doesn't apply to latest `main`, and the reason is a file move rather than a fix. #1206 moved `src/components/**` to `src/modules/**`, so the file this PR patches (`src/components/auth/context/AuthContext.tsx`) no longer exists at that path; it's now `src/modules/auth/context/AuthContext.tsx`.

The shape it fixes did survive the move. On `main` at `5e73a49b`:

- `server/modules/auth/auth.middleware.ts:78` — sets `X-Refreshed-Token` on any authenticated response past half-life
- `src/shared/api.ts:40-42` — reads the header, calls `storeAuthToken()`
- `src/shared/authToken.ts:97` — dispatches `AUTH_TOKEN_REFRESHED_EVENT`
- `src/modules/auth/context/AuthContext.tsx:192` — the listener calls `setToken(nextToken)`
- `src/modules/auth/context/AuthContext.tsx:247` — `checkAuthStatus` still lists `token` in its dependency array
- `src/modules/auth/context/AuthContext.tsx:260` — the mount effect still depends on `checkAuthStatus`

#1310 has the same fix against the current layout, plus a vitest regression test that reports **1** request per mount with the fix and **11** with `token` restored to the dependency array. It also drops the `react-test-renderer` devDependencies and the `src/shared/utils.ts` hunk (#1174 landed the latter), so it touches no dependencies at all.

Happy to close this one in favour of #1310.

PR #1178, opened 2026-08-19
perf(sessions): resolve the session title without holding the transcript

TadMSTR:
## What this changes

`extractSessionTitle` reads the whole `.jsonl` transcript with `readFile(filePath, 'utf8')`
and splits it, on every watcher add/change event, for any session without a settled title —
in practice every currently-active session. On a multi-megabyte transcript that is a
multi-megabyte string plus a full `lines` array allocated per filesystem event.

This replaces the forward full-file pass with a backwards chunked scan.

## Why backwards

#1258 established the priority `custom-title` > `ai-title` > `last-prompt`, taking the
**last** occurrence of each type. Scanning backwards, the first hit of a type *is* its last
occurrence — so the same answer falls out without reading forward to EOF, and a
`custom-title`, which nothing earlier in the file can outrank, stops the scan where it is
found.

`readLinesBackwards` (new, in `server/shared/utils.ts`) is an async generator that walks a
file from EOF in bounded chunks and yields whole lines. The subtle part is that a chunk
boundary almost always falls mid-line and may split a multi-byte character, so the bytes
before a chunk's first newline are carried **undecoded** into the next chunk, and a range is
decoded only once it is known to be a whole line. Peak memory is one chunk plus the longest
line rather than the file.

## What this does and does not buy

Being straight about this, because the first revision of this PR oversold it:

- Bounded peak memory — no multi-megabyte string plus `lines` array per fs event.
- An early exit for `/rename`d sessions: the scan stops at the `custom-title`.
- **Not** a smaller read for every session. A session that was never renamed still scans
  back to its `ai-title`, which is written after the first exchange and so sits near the
  *start* of a long transcript. There it reads about as much as before — just without
  holding it all at once.

That earlier revision read only the last 64 KiB. That is not compatible with #1258: under
the priority chain, finding a marker in the tail does not authorise returning it, because a
higher-priority marker may sit earlier. Only a `custom-title` hit is safe to return early,
and `ai-title` is normally nowhere near EOF — so an exact tail version falls back to the
full read for essentially every session that was never renamed. The tail window is gone.

## Tests

`server/modules/providers/tests/claude-session-title-scan.test.ts` replaces the old
tail-window test file:

- **reverse reader** — reverse order across many chunk boundaries; a 4-byte emoji
  deliberately split by a chunk boundary, where the test first asserts the boundary really
  does split it, so it cannot silently stop testing what it claims; an exhaustive sweep over
  every chunk size from 1 upward against the forward `split('\n')`; sub-chunk file; no
  trailing newline; empty file; missing file; early `break` closes the handle.
- **extraction** — an `ai-title` behind 4 MB of transcript is still returned; `custom-title`
  outranks later markers; newest of each type wins; another session's markers are ignored;
  no marker gives `undefined`.
- the `custom-title` early exit is asserted **by measuring bytes delivered to userspace**
  (`/proc/self/io` `rchar`), not by the returned title — the title is the same either way,
  so a value-only assertion would pass whether or not the early exit actually works.

`server/modules/providers/tests/claude-sessions.test.ts` passes unedited.

## Alternative

If the reverse reader looks like more machinery than this warrants, the simpler exact option
is a forward `readline` stream: it still removes the big-string allocation and is much less
code, but it has no early exit. Happy to switch if you would prefer that.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved automatic session-title detection across large transcripts.
  - Session titles now prioritize the most recent relevant title information and correctly fall back to the latest prompt when needed.
  - Improved handling of multiline text, special characters, empty transcripts, and incomplete files.

- **Performance**
  - Reduced file reading when the appropriate session title is found, improving responsiveness for lengthy transcripts.

### DIFF
--- a/server/modules/providers/list/claude/claude-session-synchronizer.provider.ts
+++ b/server/modules/providers/list/claude/claude-session-synchronizer.provider.ts
@@ -8,9 +8,21 @@ import {
   extractFirstValidJsonlData,
   findFilesRecursivelyCreatedAfter,
   normalizeSessionName,
+  readFileTail,
   readFileTimestamps,
 } from '@/shared/utils.js';
 import type { IProviderSessionSynchronizer } from '@/shared/interfaces.js';
+
+/**
+ * How much of a transcript's tail to scan for a title marker.
+ *
+ * Markers are appended as the session runs, so the newest sits within the last
+ * few hundred bytes: on a 6.1 MB transcript measured here the final marker began
+ * 342 bytes from EOF. 64 KiB is ~190x that, and still covers files whose last
+ * records are unusually large, while a miss only costs the full read that this
+ * function already performed unconditionally.
+ */
+const SESSION_TITLE_TAIL_BYTES = 64 * 1024;
 
 type ParsedSession = {
   sessionId: string;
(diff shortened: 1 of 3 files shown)


TadMSTR:
Rebased onto current `main` (`5e73a49b`). Offset-only churn, no conflicts — same three files, same diffstat.

The target is unchanged on current `main`: `claude-session-synchronizer.provider.ts:166` still does the full-file `await readFile(filePath, 'utf8')` on every add/change event. #1206's transcript-performance work did not touch this scan, so the PR still applies as written.

The 10 tests in `server/modules/providers/tests/claude-session-title-tail.test.ts` pass on the new base.

blackmammoth:
hey @TadMSTR, there is a merge conflict because #1258 was merged. Can u check if this is still relevant and if so, resolve the merge conflicts?

TadMSTR:
@blackmammoth thanks for the ping — yes, still relevant. The conflict turned out to be semantic rather than textual, so I have rewritten the change rather than merged it.

**Still relevant:** `extractSessionTitle` on current `main` still does an unconditional `readFile(filePath, 'utf8')` of the whole transcript per watcher event. #1258 also removed the early exit the old reverse scan had, so the forward pass now walks to EOF every time.

**Why it was rewritten rather than merged:** #1258's priority (`custom-title` > `ai-title` > `last-prompt`, last occurrence of each) is incompatible with the 64 KiB tail window this PR used to have. Finding a marker in the tail does not authorise returning it, since a higher-priority one may sit earlier in the file — and `ai-title` is written after the first exchange, so on a long session it is nowhere near EOF. Resolving the conflict in this PR's favour would have reintroduced exactly the bug #1258 fixed; I checked before rewriting, and it fails three of your title tests:

- `synchronizeFile uses ai-title from JSONL when no DB custom_name exists`
- `synchronizeFile uses custom-title from JSONL when no DB custom_name and no ai-title`
- `synchronizeFile does NOT treat "Untitled Claude Session" in DB as a real custom_name`

So the tail window is gone. The scan now runs backwards in bounded chunks, where the first hit of a type is its last occurrence — the same answer as the forward pass — and a `custom-title` stops it immediately, since nothing earlier can outrank it. `claude-sessions.test.ts` passes unedited.

**What it actually buys**, to be straight about it: bounded peak memory (no multi-MB string plus `lines` array per fs event) and an early exit for `/rename`d sessions. It is *not* a smaller read for every session — one that was never renamed still scans back to its `ai-title` near the start. The original "does not read the whole transcript" framing was too strong and I have dropped it from the description.

If the reverse reader looks like more machinery than this warrants, the simpler exact alternative is a forward `readline` stream: kills the big allocation, much less code, no early exit. Say the word and I will swap it.

PR #1179, opened 2026-08-19
fix: queue chat.send when WebSocket is not open and reconnect faster

sandy9707:
Fixes mobile WebSocket race condition where chat.send was silently dropped when the socket was not OPEN.

- sendMessage now queues messages in memory when the socket is not OPEN instead of silently dropping them.
- Flush queued messages on reconnect in onopen.
- Reduce onclose reconnect delay from 3000ms to 600ms.
- Reconnect immediately on visibilitychange / online / focus.
- Show a small toast when the connection drops and when it recovers.
- Remove the stale ws.readyState guard in useQueuedMessageAutoSend because sendMessage now handles queueing.

Type: bugfix / mobile reliability

## Summary by CodeRabbit

* **New Features**
  * Outgoing messages are queued while disconnected and automatically sent after reconnection.
  * Queues can now grow beyond the previous 20-message limit.
  * Queued messages are dispatched without waiting for a connection-state check.
  * Connections retry sooner and respond to visibility, online, and focus changes.
  * Added notifications when the connection is lost and restored.

* **Bug Fixes**
  * Failed message sends now preserve the failed message and following messages for retry.
  * Pending messages are cleared when the authenticated account changes.
  * Prevented stale authentication responses caused by browser caching.

### DIFF
--- a/server/index.ts
+++ b/server/index.ts
@@ -142,6 +142,12 @@ app.get('/health', (req, res) => {
     });
 });
 
+// Hotfix for CloudCLI auth stale-token cache issue (#1174)
+app.use('/api', (_req, res, next) => {
+    res.setHeader('Cache-Control', 'no-store');
+    next();
+});
+
 // Optional API key validation (if configured)
 app.use('/api', validateApiKey);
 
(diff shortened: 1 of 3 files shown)


blackmammoth:
hey @sandy9707, queued messages are now stored in the database so they should work with this case as well. If issue persists, you can resubmit the PR.

PR #1207, opened 2026-08-25
fix(claude): ignore <synthetic> model placeholder when resolving session model

coocon:
## Problem

Claude Code stamps locally-synthesized rows in session JSONL files (API-error placeholders and similar) with `model: "<synthetic>"`. This is an internal marker, not a real model id.

`readClaudeSessionModelFromJsonl` (in `server/modules/providers/list/claude/claude-models.provider.ts`) walks the session JSONL backwards and takes the first event carrying a model field as the session's active model. When the last row of a session happens to be such a placeholder (e.g. the previous turn ended with an API error), `<synthetic>` is surfaced as the session's model.

From there it propagates end to end:

1. `resolveSessionModel` returns it (`source: 'provider'`) via `GET /:provider/sessions/:sessionId/active-model`;
2. the composer adopts it as the session's current model;
3. the next send passes it back as `clientOptions.model`, which is recorded via `setSessionModel` and handed to the Agent SDK;
4. the SDK issues a real API request with `model: "<synthetic>"`, which the API rejects with **404 model_not_found**.

For users behind an API gateway this is worse than a single failed request — gateways commonly rate-limit an account after an upstream `model_not_found`, so one poisoned session can block real traffic for a while.

## Fix

Treat angle-bracketed values as placeholders in `extractClaudeEventModel` and skip them, so the scan falls through to an earlier event with a real model id, or ultimately to the provider default. This filters the poison at its single entry point; everything downstream is just faithful propagation.

## Testing

- `npm run typecheck` passes
- `server/modules/providers/tests/provider-models.service.test.ts` passes (16/16)
- Reproduced against a real session JSONL containing `"model":"<synthetic>"` rows: before the fix the active-model endpoint reported `<synthetic>`; after the fix it falls back correctly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Claude model detection by excluding synthetic placeholder values such as `<synthetic>`.
  * Ensured model information is reported consistently across message and event data.
  * Preserved valid model detection when placeholder values appear alongside real model information.

### DIFF
--- a/server/modules/providers/list/claude/claude-models.provider.ts
+++ b/server/modules/providers/list/claude/claude-models.provider.ts
@@ -207,19 +207,20 @@ const extractClaudeModelFromTextContent = (content: string): string | null => {
   if (localCommandStdout !== null) {
     const cleanedStdout = stripAnsi(localCommandStdout).replace(/\s+/g, ' ').trim();
     const changedModel = /(?:set|changed|switched)\s+model\s+to\s+(.+?)\.?$/i.exec(cleanedStdout);
-    if (changedModel?.[1]?.trim()) {
-      return changedModel[1].trim();
+    const stdoutModel = changedModel?.[1]?.trim();
+    // A placeholder stdout hit must not shadow a real <model> tag further down.
+    if (stdoutModel && !isPlaceholderModel(stdoutModel)) {
+      return stdoutModel;
     }
   }
 
   const modelTag = extractTaggedContent(content, 'model')?.trim();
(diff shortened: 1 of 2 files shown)


blackmammoth (approved):
Hey @coocon , thanks for the PR. It's been merged. Also, if you’d like to join project discussion or community chat, we have a Discord here: [[link](https://discord.gg/buxwujPNRE)](https://discord.gg/buxwujPNRE). Totally optional!

Issue #1008, opened 2026-07-12
Codex auth status reports "not authenticated" when Codex CLI (>=0.144) stores credentials in the OS keychain


pfedotovsky:
## Summary

Codex CLI 0.144+ stores login credentials in the OS keychain by default (macOS keychain item `Codex Auth`) and no longer writes `~/.codex/auth.json`. The Codex auth check in `server/modules/providers/list/codex/codex-auth.provider.ts` (`checkCredentials`) reads **only** `auth.json`, so `GET /api/providers/codex/auth/status` returns `authenticated: false` even right after a successful `codex login`, and the UI tells the user Codex is not signed in.

Notably, **actual Codex chat sessions work fine** in this state: `@openai/codex-sdk` spawns a codex binary, and the keychain item's ACL matches OpenAI's code signature, so the binary reads its own credentials without any prompt. Only the status endpoint is wrong.

## Reproduce

1. macOS host, Codex CLI >= 0.144 (verified with 0.144.1), CloudCLI 1.36.1.
2. `codex login` → completes, `codex login status` prints "Logged in using ChatGPT".
3. `ls ~/.codex/auth.json` → does not exist (credentials are in the keychain).
4. `GET /api/providers/codex/auth/status` → `{"authenticated": false, "error": "Codex not configured"}`.

## Why not read the keychain directly

Reading the keychain item from Node (or shelling out to `security find-generic-password`) triggers a macOS GUI permission dialog, because the item's ACL only covers OpenAI-signed codex binaries — a headless server can't get past that.

## Suggested fix

Fall back to the Codex CLI itself when `auth.json` is missing or has no tokens: `codex login status` exits `0` when logged in (any credential store) and non-zero otherwise, and the codex binary reads its own keychain item silently. Something like:

```ts
private checkCliLoginStatus(): CodexCredentialsStatus | null {
  try {
    const result = spawn.sync('codex', ['login', 'status'], { encoding: 'utf8', timeout: 10000 });
    if (result.status === 0) {
      const output = `${result.stdout ?? ''}${result.stderr ?? ''}`;
      const method = /api key/i.test(output) ? 'api_key' : 'keychain';
      return { authenticated: true, email: 'Authenticated', method };
    }
  } catch {
    // CLI missing or unresponsive; treat as no keychain credentials.
  }
  return null;
}
```

[...]

PR #1216, opened 2026-08-26
fix(providers): report Claude auth status from the CLI instead of files

### LINKED ISSUE #1209: Claude account reports "login has expired" when the access token is refreshable
**Describe the bug**
Settings → Agents → Claude → Account reports the account as disconnected, with
"Claude login has expired. Run claude /login again.", for a login that is still valid.

**To Reproduce**
1. Log in with `claude /login` (OAuth credentials land in `~/.claude/.credentials.json`)
2. Leave the CLI unused until `claudeAiOauth.expiresAt` passes (the access token lives ~12 hours)
3. Open Settings → Agents → Claude → Account

**Expected behavior**
Connected. The access token is stale but `claudeAiOauth.refreshToken` is present, and the
CLI refreshes it silently on its next run — nothing about the login is lost. `claude auth
status --json` reports `{"loggedIn": true, ...}` in exactly that state.

**Error message**
```
Claude login has expired. Run claude /login again.
```

**Additional context**
`checkCredentials()` in `server/modules/providers/list/claude/claude-auth.provider.ts`
compares `expiresAt` against `Date.now()` and never looks at `refreshToken`. Anyone who
hasn't used the agent since the previous day sees a disconnected account until something
makes the CLI run.

- OS: Linux
- Browser: Chromium
- Version: CloudCLI 1.37.2
- Claude Code: 2.1.236

sebauvray:
Closes #1209.

## The problem

`ClaudeProviderAuth.checkCredentials()` infers whether the user is logged in by reading `~/.claude/.credentials.json`. That file is a cache the CLI is free to bypass, so `GET /api/providers/claude/auth/status` reports a disconnected account for sessions that are perfectly valid.

Two ways it goes wrong:

- **macOS Keychain.** Claude Code stores its OAuth token in the Keychain and leaves `~/.claude/.credentials.json` behind as a stale copy. On my machine that file is dated a month ago and a successful `claude /login` never rewrites it — so the "Login" button in Settings → Agents → Claude cannot fix the state it reports. This is the Claude twin of #1008.
- **Refreshable token.** Even when the file is authoritative, `accessToken` expires ~12 h before the session does, because `refreshToken` renews it silently. That is #1209 as filed.

In both cases chat sessions keep working — the provider spawns the `claude` binary, which reads its own credentials. Only the status endpoint is wrong, which makes the UI actively misleading.

## The fix

Ask the CLI instead of guessing:

```
$ claude auth status --json
{ "loggedIn": true, "authMethod": "claude.ai", "email": "...", "subscriptionType": "max" }
```

`checkCliStatus()` runs that and, on `loggedIn: true`, reports the account as connected along with the real email.

**Placement.** The probe sits *after* the explicit environment and `settings.json` checks, so a hand-configured `ANTHROPIC_API_KEY` / `CLAUDE_CODE_OAUTH_TOKEN` still wins, and *before* the credentials-file read, which is the branch that is wrong. No existing precedence changes.

**Fails open.** A missing, older or slow CLI returns `null` and the previous file-based detection runs untouched. `spawn.sync` is already used by `checkInstalled()` in the same class, and the call is bounded by a 10 s timeout.

## Tests

The probe is stubbed in `claude-auth.test.ts`, which also restores the "never shells out" property the test file documents — before that, the existing assertions silently depended on whether the machine running the suite happened to have a logged-in CLI, and each one spawned a real process (~640 ms → <1 ms).

Two cases added: a logged-in CLI wins over an expired credentials file, and explicit env credentials still win over the probe.

Verified locally: `tsc --noEmit -p server/tsconfig.json`, `eslint` on the touched files, `npm run build:server`, and `npm test` (`claude-auth.test.ts` 7/7).

One note in case CI is red: `server/modules/agent/tests/agent.routes.test.ts` is flaky on my machine (Node 26, `.nvmrc` asks for 22) — it fails with `Unable to deserialize cloned data` on roughly one run in three, with and without this change. Happy to look into it separately if it is not already known.

## Summary by CodeRabbit

- **New Features**
  - Authentication now recognizes active sign-ins reported by the Claude CLI.
  - Active CLI authentication can remain valid even when stored credentials are expired.
  - Explicitly configured environment credentials continue to take priority.

- **Bug Fixes**
  - Improved authentication checks when credential files are unavailable, expired, or unsupported.

### DIFF
--- a/server/modules/providers/list/claude/claude-auth.provider.ts
+++ b/server/modules/providers/list/claude/claude-auth.provider.ts
@@ -78,6 +78,46 @@ export class ClaudeProviderAuth implements IProviderAuth {
   }
 
   /**
+   * Asks the Claude Code CLI whether it is logged in.
+   *
+   * This is the only authoritative source. The credentials file below is a
+   * cache the CLI is free to bypass: on macOS the OAuth token lives in the
+   * Keychain and `~/.claude/.credentials.json` is left behind as a stale
+   * copy, and even when the file is used its access token expires long
+   * before the session does, because `refreshToken` renews it silently.
+   *
+   * Returns null when the CLI cannot answer (missing, too old, timing out),
+   * so the caller falls back to the previous file-based detection.
+   */
+  private checkCliStatus(): ClaudeCredentialsStatus | null {
+    const cliPath = resolveClaudeCodeExecutablePath(process.env.CLAUDE_CLI_PATH);
+
+    try {
+      const result = spawn.sync(cliPath, ['auth', 'status', '--json'], {
+        encoding: 'utf8',
+        timeout: 10000,
+      });
+
+      if (result.status !== 0 || typeof result.stdout !== 'string') {
+        return null;
+      }
+
+      const status = readObjectRecord(JSON.parse(result.stdout));
+      if (!status || status.loggedIn !== true) {
+        return null;
+      }
+
+      return {
+        authenticated: true,
+        email: readOptionalString(status.email) ?? 'Authenticated',
+        method: 'cli_status',
+      };
+    } catch {
+      return null;
+    }
+  }
+
+  /**
    * Checks Claude credentials in the same priority order used by Claude Code.
    */
   private async checkCredentials(): Promise<ClaudeCredentialsStatus> {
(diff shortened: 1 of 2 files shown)


blackmammoth:
hey @sebauvray, this should be fixed with the latest commits in main. Check them and if the issue still persists, you can submit the PR.

PR #1220, opened 2026-08-27
fix(sessions): keep an archived session archived across a rescan

### LINKED ISSUE #1219: [BUG] Archiving a session does not stick: the full rescan un-archives every session created since the last project-list fetch
**Describe the bug**

Archiving a session does not stick. The session reappears in the sidebar the next time the project list is loaded — but only if its transcript file was *created* after the previous project-list fetch. Archiving an older session works, which is why this is easy to miss.

Measured here: a session archived at 20:14:02 and never written to afterwards was back at `isArchived = 0` at 20:17:38 — the same second `scan_state.last_scanned_at` advanced. Nothing had touched the transcript in those three and a half minutes; a bare `GET /api/projects` was enough.

**To Reproduce**

1. Load the sidebar (this advances the scan cursor to now).
2. Start a new session and send one message.
3. Archive that session from its menu — it disappears as expected.
4. Reload the page.

The session is back in the list, and its `isArchived` column is `0` again.

**Expected behavior**

An archived session stays archived until something is actually written to it.

**Screenshots**

Not applicable — the state is in the database.

**Error message**

None. It fails silently: the archive succeeds and is undone by the next list fetch.

 - OS: Ubuntu 26.04 on WSL2 (server), macOS (browser)
 - Browser: Chrome 151
 - Version: CloudCLI 1.37.2, installed from npm, Node 26

**Smartphone (please complete the following information):**
 - Not device specific; the flag is reset server side, so every client sees it.

**Additional context**

`sessionsDb.createSession` resets the flag on every re-index, unconditionally, in both paths (`server/modules/database/repositories/sessions.db.ts`, around lines 114 and 144):

```js
isArchived = 0,
```

Re-indexing is not limited to sessions that changed:

- `getProjectsWithSessions()` and `getArchivedProjectsWithSessions()` call `sessionSynchronizerService.synchronizeSessions()` on every request, so the full rescan runs on every sidebar load, not only at server start.
- The rescan selects files with `findFilesRecursivelyCreatedAfter(...)`, which compares **`birthtime`** against `scan_state.last_scanned_at`. It therefore re-indexes every transcript created since the last fetch, whether or not its content changed — and each of those loses its archive flag.

The reset itself is clearly intended: a session that is being written to again should come back out of the archive. What is missing is the check that anything was actually written:

```sql
-- UPDATE path (needs the timestamp bound a second time)
isArchived = CASE WHEN julianday(?) > julianday(updated_at) THEN 0 ELSE isArchived END,

-- ON CONFLICT path
isArchived = CASE WHEN julianday(excluded.updated_at) > julianday(sessions.updated_at)
                  THEN 0 ELSE sessions.isArchived END,
```

`julianday()` parses both formats that occur in the column (`2026-08-27 20:14:02` from the column default and `2026-08-27T20:17:38.896Z` from `readFileTimestamps`), and a NULL timestamp fails the comparison, which keeps the session archived rather than discarding the user's action.

I am running exactly this as a local patch against `dist-server/`: untouched sessions stay archived, and sessions that are written to again still come back. Happy to send it as a PR if the approach looks right.

buttjer:
Fixes #1219.

## The bug

`createSession` resets `isArchived` on every re-index, without checking whether anything was actually written:

```js
isArchived = 0,
```

That is fine for the case it was written for — a session being written to again should come back out of the archive — but re-indexing is not limited to sessions that changed. `getProjectsWithSessions()` runs `synchronizeSessions()` on every request, so the full rescan happens on each sidebar load, and the rescan picks files by **`birthtime`** against `scan_state.last_scanned_at`. Every session created since the previous fetch is therefore re-indexed unchanged — and loses its archive flag.

The practical effect: archiving a session you just finished never sticks, while archiving an older one works. Measured before the fix: a session archived at 20:14:02 and never touched again was back at `isArchived = 0` at 20:17:38, the same second the scan cursor advanced.

## The fix

Reset only when the caller actually reports newer activity:

```sql
isArchived = CASE WHEN ? IS NULL OR julianday(?) > julianday(updated_at) THEN 0 ELSE isArchived END,
```

- An explicit timestamp that is newer than the stored one → reactivate, as before.
- No timestamp at all → reactivate. The statement already writes `updated_at = COALESCE(?, CURRENT_TIMESTAMP)` for that call, so the two now agree. This arm is load-bearing: coalescing inside the comparison instead is not equivalent, because `CURRENT_TIMESTAMP` resolves to whole seconds and "now" is not *newer* than a row written in the same second — which is precisely what `createSession reactivates archived rows when the session becomes active again` does, and it goes red without it.
- Same timestamp (an unchanged file, re-indexed by the (rescan) → leave the flag alone.

`julianday()` parses both forms that occur in the column: `2026-08-27 20:14:02` from the column default and `2026-08-27T20:17:38.896Z` from `readFileTimestamps`.

The `ON CONFLICT` path gets the same condition against `excluded.updated_at`, which is already coalesced there.

## Tests

`server/modules/database/tests/sessions.db.integration.test.ts` gains a case for the rescan: create with timestamps, archive, re-index with the *same* timestamps, assert it is still archived — then re-index with a newer one and assert it comes back.

Full server suite: 271 passing, 0 failing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Archived sessions now remain archived when recreated with older session data.
  * Sessions are reactivated when recreated without timestamps or with current or newer data.
  * Improved archive-state handling for sessions created without explicit provider timestamps.
  * Prevented outdated session information from unintentionally changing archive status.

* **Tests**
  * Added coverage for reactivating archived sessions and preserving their archived status when older data is received.
  * Added validation for timestamped and app-created session scenarios.


buttjer:
@coderabbitai review

buttjer:
@coderabbitai full review

blackmammoth (approved):

PR #1225, opened 2026-08-30
fix(chat): merge back-to-back queued messages instead of overwriting

nateQQ:
## What

While a turn is running, queuing a second message replaced the first one instead of appending to it - both in React state and in the `queued_message_<sessionId>` localStorage record.

## Repro

1. Send a message that starts a long-running turn.
2. While it's still running, type "A" and press Enter - it gets queued (shown in the queued-message card).
3. Before the turn finishes, type "B" and press Enter.
4. The queued card now shows only "B". "A" is gone and is never sent.

## Fix

`useChatComposerState.ts`'s `handleSubmit` now checks for an existing queued draft for the session before building the new one, and merges into it instead of replacing it:
- text is joined with a blank line
- attachments and uploaded-attachment descriptors are concatenated
- send options use the latest snapshot (most recent model/permission settings win)

So queuing "A" then "B" mid-turn now delivers a single message containing both once the turn ends, instead of silently dropping "A".

## Testing

- `npm run typecheck` - passes
- `npx eslint src/components/chat/hooks/useChatComposerState.ts` - clean
- `npm run build` - passes
- Traced both the same-session queue path and the cross-session dispatch path (`useQueuedMessageAutoSend.ts`, which only reads/sends after a turn completes and needed no changes) to confirm the merged draft flows through unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message queuing during active conversations.
  * Multiple queued messages are now combined instead of replacing earlier content.
  * Attachments and uploaded files are preserved across queued submissions.
  * Queued content remains accurate when messages or uploads finish in a different order.
  * Prevented newer submissions from overwriting previously queued text or attachments.

### DIFF
--- a/src/components/chat/hooks/useChatComposerState.ts
+++ b/src/components/chat/hooks/useChatComposerState.ts
@@ -724,16 +724,21 @@ export function useChatComposerState({
         }
 
         // A message may already be queued for this turn (the user queued "A"
-        // and is now queuing "B" before the run finishes). Append onto it
-        // instead of replacing it so the earlier message is never dropped.
-        const existingDraft =
-          queuedSessionKey && queuedDraftSessionRef.current === queuedSessionKey ? queuedDraft : null;
+        // and is now queuing "B" before the run finishes). Re-read the
+        // persisted draft here rather than trusting the `queuedDraft` closure:
+        // two submissions can each be past their own upload await with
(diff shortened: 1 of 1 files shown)


blackmammoth:
hey @nateQQ, can you resolve the merge conflicts?

PR #1227, opened 2026-08-30
Let a launcher name a desktop window and open Local CloudCLI at startup

edgar965:
Two small, opt-in additions for driving the desktop app from a launcher script. Both are off unless an environment variable is set, so nothing changes for anyone who does not use them.

## Why

Several desktop windows can already run side by side — the single-instance lock is keyed on `--user-data-dir`, so one profile per window is enough. But they are hard to live with:

- every window carries the same title, so they are indistinguishable in the taskbar and in Alt+Tab, and
- every window starts on the launcher and needs a click to reach the local app.

## `CLOUDCLI_INSTANCE_NAME`

When set, the value is prefixed to the window title: `myproject - CloudCLI - Local CloudCLI`. It goes in front because that is the end Windows keeps when it truncates a taskbar label.

One detail that took a second attempt: a loaded document's `<title>` replaces whatever `setTitle` put there, so the launcher page dropped the prefix again as soon as it finished loading. `page-title-updated` now re-applies it while keeping the document title.

## `CLOUDCLI_DESKTOP_OPEN_LOCAL` / `CLOUDCLI_DESKTOP_START_PATH`

With `CLOUDCLI_DESKTOP_OPEN_LOCAL=1`, the window opens Local CloudCLI directly instead of the launcher. `CLOUDCLI_DESKTOP_START_PATH` optionally picks the page inside the web UI, resolved against the local server url.

`openLocalAtStartup()` mirrors the existing `openLocalInDesktop()` — same pending target, same startup logs — and only adds the path. Any failure goes through the usual `showError`, so a bad value cannot stop the app from starting.

## Testing

Verified on Windows 11 with several windows open at once, each with its own `--user-data-dir` and name, all sharing one local server. With neither variable set, titles and startup are byte-for-byte as before.

## Summary by CodeRabbit

* **New Features**
  * Added optional instance-name prefixes to desktop window titles, making side-by-side instances easier to distinguish.
  * Added automatic opening of the local environment at startup.
  * Added support for specifying an initial startup path.

* **Bug Fixes**
  * Preserved instance-name prefixes when loaded pages update their document titles.
  * Improved startup path validation, safe fallback behavior, and error reporting for unusable or external paths.


edgar965:
@coderabbitai review

PR #1229, opened 2026-08-30
Collapsible branches in the model picker

edgar965:
## The problem

The model picker on the new-session screen lists every model of every installed provider in
one flat run. An OpenCode installation alone contributes ninety, so reaching the provider
below it means scrolling past all of them, every time.

## What this changes

Each provider branch collapses by clicking its heading, with the model count next to it.
The collapsed set is remembered in `localStorage` (`model-groups-collapsed`), so a picker
comes back the way it was left.

Two details that are easy to get wrong:

- **A search opens every branch** for as long as it runs. A hit inside a collapsed branch
  would otherwise be invisible, which turns the search into a liar.
- **A collapsed branch still renders one row** — `N hidden — click to show`. cmdk removes a
  group whose items are all filtered out, heading included, and a heading that is no longer
  in the DOM cannot be clicked open again. That row doubles as the click target.

The rendering moved into `ModelGroupList.tsx` rather than growing the 440-line empty state
further. `ModelGroup` keeps the branch and the provider it selects as separate fields, so a
branch does not have to be exactly one provider.

## Notes

- No behaviour change when nothing is collapsed: same groups, same order, same items.
- `localStorage` access is wrapped — a browser that refuses it just does not remember the
  state.

## Summary by CodeRabbit

- **New Features**
  - Added a grouped model picker organized by provider, with collapsible sections and remembered expansion state.
  - Added model search, selected-model indicators, custom-model badges, and loading placeholders.
  - Improved provider and model selection with centralized picker controls and easier navigation to the model library.
  - Search is cleared when the picker is closed for a fresh experience the next time it opens.


edgar965:
Fixed in `2033369` — the finding is right. The updater now only computes the next set, and
an effect on `collapsedKeys` does the `localStorage` write, so only committed state is
persisted (and a doubled updater call under StrictMode no longer writes twice).

edgar965:
@coderabbitai review

blackmammoth (approved):
@edgar965, thanks for the PR. It's been merged. Also, if you’d like to join project discussion or community chat, we have a Discord here: [[link](https://discord.gg/buxwujPNRE)](https://discord.gg/buxwujPNRE). Totally optional!

PR #1235, opened 2026-08-31
Put the composer's tools above the input, not below it

edgar965:
The attach, voice, model and permission controls sat under the text area. On a tall message the row was pushed off the bottom of the composer, and while typing it moved further away with every line — so the controls kept sliding out from under the pointer.

Moving the row above the input fixes the position: the tools stay where they were when you started typing, and the text grows downward instead of pushing them around.

## Summary by CodeRabbit

* **Style**
  * Updated the chat composer layout so the input area is contained within the footer.
  * Improved spacing and wrapping for composer controls, providing a more consistent interface across screen sizes.


edgar965:
@coderabbitai review

blackmammoth:
@edgar965 tested and it works fine. 
image
So, closing this.

PR #1237, opened 2026-09-01
Desktop: stop the content view from outgrowing its window

edgar965:
### The symptom

In the desktop app, the page is cut off at the right edge — mid-word, mid-sentence, with no scrollbar to suggest anything is missing. It does not always happen; it depends on how the window was maximised, resized or moved.

### The cause

`ViewHost.attach` gives the content view two things that both decide its size:

```js
view.setBounds(this.getContentViewBounds());
view.setAutoResize({ width: true, height: true });
```

`setAutoResize` grows the view along with the window on its own, and `DesktopWindowManager` also calls `resizeActiveView()` — which sets the bounds to the full window width — on every `resize`. The two additions stack, and the view ends up wider than the window it lives in. The page inside is fine; it is simply rendered into a view that reaches past the right edge.

### The change

- `setAutoResize` is gone. `setBounds` alone covers every case it covered, and it is the one that knows about `TITLEBAR_HEIGHT`.
- The bounds are now re-applied on `maximize`, `unmaximize`, `enter-full-screen` and `leave-full-screen` as well. `resize` alone misses a maximise that reports its final size late.
- `move` re-applies them too, for a drag onto a display with a different scaling factor: the window keeps its size in device pixels but changes it in CSS pixels, and no `resize` is emitted.

Two files, sixteen lines. `setAutoResize` has been there since the desktop app landed in 97c9b67.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved content layout when the desktop window is resized, maximized, restored, moved, or switched into or out of full-screen mode.
  - Fixed an issue where page content could extend beyond the right edge of the window, potentially hiding content and preventing access to the horizontal scrollbar.
  - Settings window positioning now stays synchronized across window size and display-state changes.


edgar965:
@coderabbitai review

blackmammoth:
@edgar965, closed this as the issue you suggested doesn't exist when I tried to reproduce it.

PR #1240, opened 2026-09-01
Only sign out over a token the request actually sent

edgar965:
### The bug

`authenticateToken` answers a request with no `Authorization` header with the same `X-Auth-Error` header it uses to reject a bad token:

```js
if (!token) {
  res.setHeader('X-Auth-Error', 'invalid-token');
  return res.status(401).json({ ... });
}
```

`authenticatedFetch` acts on that header by clearing `auth-token` and firing `AUTH_SESSION_EXPIRED_EVENT`, which puts "Your session expired. Please log in again." on screen — **even when the request never carried a token to begin with.**

### Why it matters

Any call that goes out before the token is in place ends the session:

- a tab restored from a cold start, fetching while the auth context is still loading
- a desktop window that has just been handed a login (each Electron profile has its own `localStorage`)
- a background poll racing the first render

The user lands on the sign-in screen holding a perfectly valid token, signs in, and the next such race throws it away again. It reads as "the login is broken", which is how I found it — five times in a row on a local desktop install.

### The fix

A response can only mean *your token was refused* if a token went out:

```js
if (token && response.headers.get('X-Auth-Error')) {
  expireAuthSession();
}
```

One line, no behaviour change for a request that really was rejected.

### Tests

`src/shared/tests/authenticatedFetch.test.ts` already covered "X-Auth-Error with a token signs out". This adds the other direction — without a token, nothing is cleared. Reverting the guard turns the new test red.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented authentication errors from incorrectly ending an active session when a request did not use the stored sign-in token.
  * Preserved stored sign-in information for requests authenticated through cookies or custom credentials.
  * Improved session-state handling to prevent unexpected sign-outs while maintaining accurate behavior when the stored token is genuinely rejected.


edgar965:
Hi — a note to whoever maintains this, since these two are not fork-specific.

I run CloudCLI as a self-hosted desktop install and keep a fork. Both bugs below are in upstream `main` and affect anyone on the desktop app, so I would rather contribute them back than keep patching downstream.

**This PR (#1240) — the app signs itself out over requests that never carried a token.**
`authenticateToken` answers a request with no `Authorization` header with the same `X-Auth-Error` header it uses to reject a bad token, and `authenticatedFetch` reacts to that header by clearing `auth-token` and raising "Your session expired" — even though nothing was rejected, because nothing was sent. Anything that fetches before the token is in place trips it: a tab restored from a cold start, a desktop window that has just been handed a login, a background poll racing the first render. The user lands back on the sign-in screen holding a valid token, signs in, and the next race throws it away again. Here it happened five times in a row before I found it, and it reads as "the login is broken". The fix is one condition, and the existing test for the rejecting case stays green.

**#1241 — the desktop app is told an address the server never bound to.**
`writeLocalServerMarker` writes `DISPLAY_HOST` into `local-server.json`, which maps every loopback address to `localhost`, and `electron/localServer.js` opens exactly that url. `localhost` and `127.0.0.1` are separate browser origins, so a window on the other one cannot see the first one's `localStorage` — the auth token included, which is the same sign-in screen again, on a server that is signed in on the other origin. In the desktop app there is no second place the token could come from, since every window has its own Electron profile. On Windows there is a measurable cost on top: `getaddrinfo('localhost')` returns `::1` first and the server listens on IPv4 only, so every connection pays for a failed IPv6 attempt — median 2923 ms over `localhost` against 840 ms over `127.0.0.1`, same endpoint, five calls each. Only the machine-readable marker changes; `DISPLAY_HOST` keeps the friendlier spelling wherever a human reads it.

Both are small and self-contained, and I am happy to adjust either — naming, placement, or a different approach entirely if you see a better one.

There are a few more PRs open from the same fork (#1226–#1237): Windows path handling, the model picker, some desktop-window work. No pressure on those — if reviewing that many at once is inconvenient, say which are worth your time and I will close the rest.

Thanks for the project, it is what I work in every day.

edgar965:
@coderabbitai review

PR #1241, opened 2026-09-01
Tell the desktop app the address the server is actually on

edgar965:
### The bug

`writeLocalServerMarker` writes `DISPLAY_HOST` into `~/.cloudcli/local-server.json`:

```js
const DISPLAY_HOST = getConnectableHost(HOST);   // any loopback → "localhost"
...
url: `http://${DISPLAY_HOST}:${SERVER_PORT}`,
```

The desktop app opens exactly that url — `electron/localServer.js` returns `marker.url` — so it connects by a name the server never bound to.

### Why it matters

**The login disappears.** `localhost` and `127.0.0.1` are separate browser origins. A window that lands on the other one cannot see the `localStorage` of the first, and the auth token lives there. The app shows "Your session expired" while the very same server, on the very same port, is signed in on the other origin. It is worse in the desktop app because every window runs with its own Electron profile, so there is no second place the token could come from.

**Windows pays for a failed IPv6 attempt on every connection.** `getaddrinfo('localhost')` returns `::1` first there, and the server listens on IPv4 only. Measured on this machine, same endpoint, five calls each: median 2923 ms over `localhost` against 840 ms over `127.0.0.1`.

### The fix

The marker names what is actually being served:

```js
const MARKER_HOST = isWildcardHost(HOST) || isLoopbackHost(HOST) ? '127.0.0.1' : HOST;
```

`DISPLAY_HOST` is untouched everywhere it is only read by a human — the startup banner, the log lines — where `localhost` is the friendlier spelling. Only the machine-readable marker changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved local server address reporting so generated connection markers point to the address the server is actually serving.
  * Corrected handling for wildcard and loopback host configurations.
  * Improved IPv6 support, including proper preservation and formatting of IPv6 loopback and other IPv6 addresses.
  * Ensured connection addresses remain accurate across IPv4 and IPv6 server configurations.


edgar965:
@coderabbitai review

PR #1244, opened 2026-09-02
feat(markdown): support TeX-style math delimiters

### LINKED ISSUE #1242: [Feature] Support TeX-style math delimiters in Markdown rendering
**Is your feature request related to a problem? Please describe.**

CloudCLI already renders mathematical expressions written with `$...$` and `$$...$$`. However, LLM responses also commonly use the standard TeX-style delimiters `\(...\)` for inline math and `\[...\]` for display math.

These delimiters are currently displayed as plain text instead of being rendered by KaTeX. This is especially noticeable in mathematical or technical conversations, where users cannot control which delimiter style a model chooses.

**Describe the solution you'd like**

Support the following additional math delimiters in chat messages and the Markdown preview:

- `\(...\)` for inline math
- `\[...\]` for display math, including multiline expressions

The existing `$...$` and `$$...$$` behavior should remain unchanged. Math delimiters inside inline code or fenced code blocks should continue to be treated as literal text.

Streaming responses should also handle an unfinished `\[...\]` block without briefly rendering it as regular Markdown before the closing delimiter arrives.

**Additional context**

Examples that should render as math:

Inline:

`\(\alpha + \beta = \gamma\)`

Display:

    \[
    E = mc^2
    \]

KaTeX is already used for rendering, so this request concerns recognizing the additional delimiters in the Markdown parsing stage rather than replacing the math renderer.

ByteEnchanter:
## Summary

- Add support for `\(...\)` inline math delimiters.
- Add support for same-line and multiline `\[...\]` display math delimiters.
- Apply the same math parsing behavior to chat messages and the Markdown editor preview.
- Keep incomplete display math blocks intact while chat content is streaming.
- Preserve the existing `$$...$$` behavior and keep single-dollar text unchanged.

## Motivation

LLM responses commonly use TeX-style `\(...\)` and `\[...\]` delimiters. Although CloudCLI already uses KaTeX, these delimiters are not currently converted into Markdown math nodes, so users see the raw delimiters instead of rendered formulas.

This change recognizes those delimiters before the existing KaTeX rendering stage without changing the renderer itself.

## Implementation notes

- Adds a local micromark extension for TeX-style inline and display delimiters.
- Reuses the existing `remark-math` MDAST integration and `rehype-katex` renderer.
- Leaves delimiters inside inline code and fenced code blocks untouched.
- Leaves escaped or unclosed delimiters as literal text.
- Updates the streaming Markdown splitter so blank lines inside an unfinished `\[...\]` block are not treated as safe split boundaries.
- Uses micromark utility packages already present in the dependency graph.
- Adds the required MIT attribution for the adapted tokenizer implementation to the existing `NOTICE` file.

## Screenshots

**Before**
image

**After**
image

image

## Testing

- `npm run test:client`
- `npm run typecheck`
- `npm run build:client`

Added coverage for:

- Inline `\(...\)` rendering
- Same-line and multiline `\[...\]` rendering
- TeX line-break options such as `\\[1em]`
- Code spans and fenced code blocks
- Escaped and unclosed delimiters
- Existing dollar-delimited behavior
- Streaming and static rendering equivalence

Closes #1242 

## Summary by CodeRabbit

- **New Features**
  - Added support for TeX-style inline math delimiters: `\( ... \)`.
  - Added support for TeX-style display math delimiters: `\[ ... \]`.
  - TeX math now renders consistently in chat transcripts and code-editor Markdown previews.
  - Display math supports multiline content, blockquotes, lists, and TeX line-break options.
  - Code spans and fenced code blocks continue to display math-like text as code.

- **Bug Fixes**
  - Improved streaming Markdown handling so TeX display blocks close correctly without disrupting subsequent content.


NOTICE
```diff
@@ -11,3 +11,29 @@ Contributions by Siteboon AI B.V. prior to commit 004135ef were originally
 published under GPL-3.0 and are hereby relicensed to AGPL-3.0-or-later.
 Contributions by other authors prior to that commit remain under GPL-3.0
 and are incorporated into this work as permitted by GPL-3.0 Section 13.
+
+Portions of src/shared/markdownMath.ts are adapted from
```
blackmammoth: @ByteEnchanter, remove this notice since it's included in the npm package downloads for it.

PR #1245, opened 2026-09-02
fix(claude): background agents were reported completed at launch

maslyankov:
## The bug

A background agent (`Agent` with `run_in_background: true`) is shown as
`completed` the instant it launches, and disappears from the UI while it is
still working.

## Why

An async agent is only *known* to have finished once its `<task-notification>`
arrives. The liveness check in `claude-sessions.provider.ts` additionally
required the agent's own transcript to end mid-tool-call:

```ts
const isAwaitingAsyncAgent = message.toolUseResult?.isAsync === true
  && !notification
  && (!subagent || subagent.endedMidToolCall);
```

That last clause is wrong for a background agent. One routinely ends its own
turn **cleanly** while its work continues — it replies *"I've started the
command in the background, waiting for it to complete…"* and stops. So
`endedMidToolCall` is `false` while no answer has arrived, and the agent is
reported `completed` immediately.

## The fix

Drop that clause. For an async agent, the absence of the notification *is* the
signal that it is still outstanding.

Note the predicate is gated on `isAsync`, so this cannot affect synchronous
agents — one short-circuits on the first conjunct and could never report
`running` here, with or without a dangling tool call. The change is confined to
the async branch.

## One existing expectation flips, deliberately

`Claude history reads a missing notification off the agent's own transcript`
asserted `completed` for exactly this shape, so it is updated (and renamed).

I'd rather be straight about it than bury it: **that scenario is genuinely
ambiguous.** A notification compacted out of a long session and a still-running
background agent look identical on disk — both end their turn cleanly with no
notification. Nothing in the transcript separates them.

The tie is now resolved in favour of the live agent, because the two error
modes are not symmetric:

|  | old behaviour | new behaviour |
|---|---|---|
| Background agent still working | **wrong, always** — vanishes at launch | correct |
| Notification compacted away | correct | shows `running` permanently |

The first is a wrong answer for *every* background agent, every time. The
second affects historical sessions whose notification has already been
compacted away — and to be clear, it **does not heal**: the notification is
gone from the file for good, and status is recomputed from the file on each
load, so that session keeps showing the agent as running. I originally wrote
that the next load corrects it; that was wrong.

Separating the two needs evidence the transcript does not carry — a staleness
cutoff on the agent's last activity would do it heuristically, a liveness
signal from the run registry exactly. I kept this PR to the one clause that is
wrong; glad to follow up with either if you have a preference.

## Two consequences worth flagging

- **`endedMidToolCall` is now computed but never read** — that clause was its
  only consumer. Kept, with a comment, because it is the only in-file trace of
  an interrupted run and is what a staleness rule would need. Happy to remove
  it instead.
- **`Claude history keeps an agent running when its transcript stops mid tool
  call` now passes for a different reason than its name** — its fixture sets
  `isAsync: true`, so it now passes on `!notification` alone. Left alone rather
  than rewritten, since it isn't what this PR is about.

## Verification

Observed on a live run: same agent, `isAsync` true, `hasNotification` false,
`endedMidToolCall` false — `isAwaitingAsyncAgent` went `false` → `true`, and
the agent now reports `running` while its work is outstanding.

`npm test` green (394 passing).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved status reporting for asynchronous agents when task completion notifications are unavailable.
  * Agents with an unknown outcome are now shown as still running instead of completed, reducing the risk of prematurely reporting unfinished work.

### DIFF
--- a/server/modules/providers/list/claude/claude-sessions.provider.ts
+++ b/server/modules/providers/list/claude/claude-sessions.provider.ts
@@ -487,12 +487,20 @@ async function getSessionMessages(
       const toolUseId = readAgentToolUseId(message);
       const notification = toolUseId ? notificationsByToolUseId.get(toolUseId) : undefined;
       // An async agent's launch row never tells you it finished — only the
-      // later notification does. When that notification is missing (a live run,
-      // or one compacted out of the transcript), the agent's own transcript is
-      // the evidence: a timeline that does not stop mid-tool-call is done.
+      // later notification does, so a missing notification is itself the signal
+      // that the agent is still outstanding.
+      //
+      // The agent's own transcript cannot stand in for that. A background agent
+      // routinely ends its turn cleanly while its work continues — it replies
+      // "I've started the command in the background, waiting for it to
+      // complete…" and stops. `endedMidToolCall` is then false even though no
+      // answer has arrived, which marked every background agent `completed` the
+      // moment it launched.
+      //
+      // `endedMidToolCall` remains the right evidence for a *synchronous* agent,
+      // where a dangling tool call is the only trace of an interrupted run.
       const isAwaitingAsyncAgent = message.toolUseResult?.isAsync === true
-        && !notification
-        && (!subagent || subagent.endedMidToolCall);
+        && !notification;
 
       if (subagent) {
         if (subagent.activity.length > 0) {
--- a/server/modules/providers/list/claude/claude-sessions.provider.ts
+++ b/server/modules/providers/list/claude/claude-sessions.provider.ts
@@ -497,8 +497,12 @@ async function getSessionMessages(
       // answer has arrived, which marked every background agent `completed` the
       // moment it launched.
       //
-      // `endedMidToolCall` remains the right evidence for a *synchronous* agent,
-      // where a dangling tool call is the only trace of an interrupted run.
+      // Note this whole predicate is gated on `isAsync`, so dropping the clause
+      // changes nothing for a synchronous agent — one has never been able to
+      // report `running` here, with or without a dangling tool call. That also
+      // leaves `endedMidToolCall` computed but unread; it is kept because it is
+      // the only in-file evidence of an interrupted run, and is what a staleness
+      // rule would need.
       const isAwaitingAsyncAgent = message.toolUseResult?.isAsync === true
         && !notification;
 
--- a/server/modules/providers/tests/claude-sessions.test.ts
+++ b/server/modules/providers/tests/claude-sessions.test.ts
@@ -219,7 +219,7 @@ async function dropTaskNotification(parentPath: string): Promise<void> {
   );
 }
 
-test('Claude history reads a missing notification off the agent\'s own transcript', { concurrency: false }, async () => {
+test('Claude history leaves an async agent running while its outcome is unknown', { concurrency: false }, async () => {
   const tempRoot = await mkdtemp(path.join(os.tmpdir(), 'claude-finished-agent-'));
 
   try {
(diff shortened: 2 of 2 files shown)


maslyankov:
Thanks — one of these is right and I've fixed it; the other I don't think holds, and checking it turned up two things worth flagging. Pushed as `c57be57` (comments only, no behaviour change).

### Stale `running` does not self-correct — correct, fixed

My test comment said the stale spinner is "corrected by the next load". That's wrong. Once the notification has been compacted out it is gone from the file permanently, and status is recomputed from the file on every load, so that session shows the agent as running forever. The comment now says so, and names what would actually separate the two cases (a staleness cutoff, or a liveness signal from the run registry).

I've corrected the same claim in the PR description.

### Synchronous agents are not affected — I don't think this one holds

The suggestion is to keep synchronous agents with `endedMidToolCall === true` running, on the basis that this change makes them disappear. But the whole predicate is gated on `isAsync`:

```ts
const isAwaitingAsyncAgent = message.toolUseResult?.isAsync === true
  && !notification
  && (!subagent || subagent.endedMidToolCall);   // ← the clause being removed
```

`isAsync === true` is the first conjunct, so a synchronous agent short-circuits to `false` and resolves to `completed` — before this change and after it. It could never report `running` here, dangling tool call or not. Removing the third conjunct cannot change that.

Worth saying that a synchronous agent interrupted mid-tool-call showing `completed` may well be wrong. But it is pre-existing, not introduced here, and fixing it means changing behaviour beyond the launch bug — happy to do it in a separate PR if you want it.

### Two real consequences, both from digging into this

1. **`endedMidToolCall` is now computed but never read.** Line 495 was its only consumer. I've kept it and said why in a comment — it is the only in-file trace of an interrupted run, and exactly what a staleness rule would need. Say the word and I'll remove it instead, or wire it into the synchronous branch.

2. **`Claude history keeps an agent running when its transcript stops mid tool call` now passes for a different reason than its name.** Its fixture sets `isAsync: true`, so after this change it passes on `!notification` alone; the dangling tool call no longer contributes. It still guards useful behaviour but is now largely redundant with the test above it. I've left it alone rather than quietly rewrite a test that isn't what this PR is about.

I also misdescribed the first point in my own comment ("`endedMidToolCall` remains the right evidence for a synchronous agent") — it never was, and that sentence is probably what made this look like a regression. Fixed.

`npm test` green (394 passing).

maslyankov:
Two corrections and some measurements, following up on the inline suggestion.

**First, my own error:** I gave the commit as `c57be57` above — that was lint-staged's backup stash, not the commit. The corrections are in **`824d5d9`**.

### On restoring a synchronous liveness check

I went and measured this rather than argue it. Across the Claude transcripts on this machine — **48 synchronous agent results** (`toolUseResult.agentId` present, `isAsync` not `true`), every transcript located:

| | count |
|---|---|
| synchronous agent results | 48 |
| agent transcript found | 48 |
| `endedMidToolCall === true` | **0** |

So the proposed `isAwaitingSyncAgent` branch would not change the status of a single one. It is not restoring behaviour this PR removed — a synchronous agent could never report `running` here, since `isAsync === true` is the first conjunct and short-circuits — and on real data it would be inert.

Three further details from working through the patch:

1. **`subagent?.endedMidToolCall ?? true` — the fallback is unreachable.** The status assignment sits inside `if (subagent) { … }`, so `subagent` is always truthy at that point and the `?? true` never applies. Worth knowing, because if it *were* reachable it would report `running` for every synchronous agent whose transcript could not be located.

2. **Synchronous agents already carry an authoritative answer.** All 48 have `status: "completed"` on `toolUseResult` itself. If the synchronous case ever does need handling, that field is better evidence than inferring liveness from transcript shape, and it cannot contradict itself the way an inference can.

3. **There is no synchronous fixture to add a regression test to.** The only fixture, `writeClaudeSubagentSession`, sets `isAsync: true`, so every test in the file exercises the async path. A synchronous test would mean inventing a fixture for a shape I have 48 real samples of and zero of them trigger.

I've left the code as-is: this PR removes one clause that is wrong for background agents, and the synchronous path is untouched by it. If you'd like the synchronous branch covered defensively anyway, I'm happy to do it in a follow-up — I'd rather base it on `toolUseResult.status` than on `endedMidToolCall`, for the reason above.

### On the stale-session rationale

That one was right and is fixed in `824d5d9`, along with the same claim in the PR description. The comment now says the stale `running` persists for compacted historical sessions rather than being corrected on the next load.

PR #1192, opened 2026-08-22
feat(i18n): connect remaining hardcoded UI strings to translations

saillill:
## What this does

This PR connects the remaining hardcoded interface strings to the i18n translation system. These strings were written directly inside components,so they stayed English no matter which language the user picked. With this change they follow the language setting: sentences with counts use the plural rules of each language, and commit dates in the history view now render in the active locale instead of a fixed en-US format.

## Which modules changed

The source control panel had the most changes because it had no translations at all. This covers the four view tabs, the header action buttons (fetch,pull, push, publish branch, revert latest commit, refresh), the branch dropdown with its search box and new-branch entry, and the titles and button labels of eight confirmation dialogs. In the changes view it includes the staged and unstaged section counters, stage all, unstage all, and the first commit guide shown when a repository has no commits yet. The commit composer placeholder, selected file counts and commit button are wired up too. On each file row there are expand diff, discard changes, delete untracked file, the scroll and wrap toggle, plus the selection bar with its select all and deselect all controls, the file status legend, and every label inside the commit history view such as author, date, line statistics and the changed file list. Four modals (new branch, merge worktree, new worktree, remove worktree) had their form labels, descriptions and loading states handled as well, along with the worktree list badges, change counters, action buttons and the explainer at the bottom.

The skills page received the upload dropzone, choose files and choose folder buttons, the ready-to-install list, install location hints, the add skill dialog, six scope badges, skill counters, empty states and a set of validation messages like folder size limits and a missing SKILL.md file.

The browser monitor panel and its settings tab were two files without any translations; everything from the panel title, runtime status badges and relative session times to the settings toggle labels is now covered.

Other areas include the onboarding flow (git configuration form, agent connection step, progress bar, validation errors), the account setup form with its fields, placeholders and auth error messages, the MCP server views(add menu, scope options and descriptions, delete confirmation), the command palette page tabs and command groups, the version upgrade modal progress output, the folder browser dialog, the TaskMaster banner, and about twenty scattered strings across the chat area such as export chat and copy command.

## Implementation and compatibility

The implementation follows the existing useTranslation pattern already used in the codebase, with no new dependencies and no build config changes. Sentences that used to concatenate template strings became t() calls with named parameters; plural-sensitive keys use the _one/_other forms that i18next resolves from the count parameter, and languages without plural distinctions simply repeat the same text in both forms.

[...]

PR #1248, opened 2026-09-02
feat(themes): user-selectable colour themes, imported from VS Code

### LINKED ISSUE #1105: feat: user-selectable color themes (theme registry)
## Problem

Right now the color palette is hardcoded in `src/index.css` under a single `:root` block (light) and `.dark` block (dark). If you want a different color scheme, the only option today is to fork and maintain a full CSS diff, then re-merge it by hand on every upstream release.

I've been running a Gruvbox-based fork for a few months exactly this way: a full override of every CSS variable in `index.css`, rebased on every update. Not sustainable for a personal fork, and not something the project should have to carry either.

## Evidence there's demand

- #1079 (open) asks for the theme to follow OS appearance - easier to reason about once themes are externalized from `index.css` in the first place.
- #907 and #787 (closed) were bugs about the code editor / dark mode not syncing - symptoms of theme state living in more than one place.

## Proposed design

Goal: don't break anything currently running, don't change the default look.

**a. Extract color variables out of `index.css`**
Move the existing `:root` / `.dark` block into a `default` theme file, e.g. `src/themes/default.css`, scoped under `[data-theme="default"]` and `[data-theme="default"].dark`. Add `src/themes/gruvbox.css` with the same variable set. No component changes needed - everything already consumes `hsl(var(--background))` etc.

**b. Extend `ThemeContext.jsx`**
Currently it only exposes `{ isDarkMode, toggleDarkMode }`. Add:

```js
{ isDarkMode, toggleDarkMode, theme, setTheme, availableThemes }
```

`setTheme` writes `document.documentElement.dataset.theme` and persists to `localStorage` (`color-theme`). `isDarkMode` stays as-is - light/dark remains its own axis, independent from the color palette.

**c. Theme registry**
`src/themes/index.ts` exporting `[{ id, name, previewColors }]`. Adding a theme becomes: one CSS file + one entry in this array.

**d. Settings UI**
A dropdown in Settings with a small color swatch preview per option.

**e. Sync `meta[name="theme-color"]`**
`ThemeContext.jsx` currently hardcodes the iOS status bar color per mode:

```js
themeColorMeta.setAttribute('content', '#141414'); // dark
themeColorMeta.setAttribute('content', '#f6f4ef'); // light
```

Reading the active `--background` value instead is a small fix that falls naturally out of this work.

**f. Scope of the first PR**
Just `default` + one more theme, to keep the diff reviewable. Additional themes would be follow-up PRs.

## Screenshot

_Attaching below - my daily driver running the Gruvbox palette purely via CSS variable overrides, no component code touched. That's roughly the shape of what a second theme would look like._

Image

## Question for maintainers

If this direction looks reasonable, I'm happy to build it. One thing I'd like your take on before starting: should theming live entirely in core (as above), or would you rather core just expose the registry/hook and themes ship as plugins? Core-first seems like the safer incremental step either way, since plugins currently only have a `tab` slot (`server/utils/plugin-loader.js`) with no way to touch global styles - but wanted to check your preference before writing code.

carsso:
Implements the theme registry requested in #1105, and takes the "import a VS Code theme" idea further than a second hardcoded palette.

## What this does

The colour tokens leave `src/index.css` for `src/shared/themes/`, and a registry decides which set is in force. Rather than shipping a second built-in palette, the app can read the themes a user already has: a `*-color-theme.json`, a whole `.vsix` extension, or a Marketplace / Open VSX URL.

Point by point against the design in the issue:

- **(a) Extract the colour variables.** The `:root` / `.dark` colour blocks move verbatim into `src/shared/themes/default.css`. They stay on the bare selectors rather than `[data-theme="default"]`, so an unstyled first paint and a user whose stored theme no longer exists both land on the defaults; themes override with `:root[data-theme=…]`, which outranks them whatever the bundle order. **No visual change to the default look.**
- **(b) Extend `ThemeContext`.** Now exposes `colorTheme`, `setColorTheme`, `availableThemes`, `addImportedTheme`, `removeImportedTheme` and `canToggleDarkMode`. Named `colorTheme` rather than `theme` because the existing `theme` preference already means light/dark. Persisted through the existing server-side preference store, so a palette follows the user across devices like their other settings.
- **(c) Registry.** `src/shared/themes/index.ts`. Adding a built-in theme is one CSS file, one side-effect import and one entry.
- **(d) Settings UI.** A picker with swatches per option, plus the import controls.
- **(e) `meta[name="theme-color"]`.** Read off the resolved `--background` instead of the two hardcoded hex values, so every palette — including one nobody could have hardcoded — gets a matching status bar.
- **(f) Scope.** One reviewable change; syntax highlighting is deliberately left out (see below).

### Light/dark is still its own axis — with one twist

The default theme ships both variants and leaves the toggle to the user. An imported palette states which of the two it is, and the `dark` class follows that statement: ~770 `dark:`-prefixed utility classes read it, so a dark palette rendered without it would put light-mode text on dark surfaces. While such a theme is active the dark-mode toggle is disabled and says so, and the command-palette entry is hidden. Switching back to the default restores the user's own light/dark choice, which was never overwritten.

### Importing

- **`.vsix`** — unzipped in the browser with `jszip` (already a client dependency, loaded through a dynamic `import()`). Reads `extension/package.json`, imports **every** theme the extension contributes, and resolves `include` chains, which only works from inside the archive. Variants are labelled from the manifest, the way VS Code's own picker labels them.
- **`*-color-theme.json`** — a hand-written JSONC parser, because published themes are full of `//` notes and trailing commas that `JSON.parse` rejects. Translucent surfaces (`#ffffff0a`) are flattened against the background, since an `hsl()` triplet has nowhere to put the alpha.
- **URLs** — the Marketplace sends no CORS headers and both registries answer an extension page with HTML, so the download goes through `POST /api/themes/download`. The server returns the raw archive and the browser does the parsing, so the file picker and the URL share one implementation of the colour mapping.

The 19 semantic roles are filled from a chain of fallbacks per role, with anything missing derived from the two keys no theme can omit. Two mapping decisions came out of testing against real themes:

- The accent comes from `activityBarBadge.background` / `textLink.foreground` before `button.background` — One Dark Pro's button is a flat grey and Solarized Light's is olive, which drained the whole UI.
- Text on the accent is chosen by WCAG contrast ratio among the theme's own extremes plus black and white — Solarized Light's mid-grey foreground is unreadable on its own yellow.

### Security

`/api/themes/download` fetches a user-supplied URL from inside the server's network, so it is constrained: https only, a host allowlist (`marketplace.visualstudio.com`, `open-vsx.org`, `openvsx.eclipsecontent.org`, `*.gallerycdn.vsassets.io`), redirects followed by hand with the allowlist re-checked on **every** hop, a 30 s timeout and a 60 MB ceiling enforced on both the header and the body. A test covers the SSRF shape: a registry URL bouncing to `169.254.169.254` is refused.

### The theme picker

Theme picker

One `.vsix` contributes several themes and all of them are registered. The dark-mode row reads "This theme sets its own light or dark appearance" and is disabled, because the active palette states which it is.

### Importing from a Marketplace or Open VSX URL

Import from a URL

### The same workspace under four palettes

| Default (unchanged) | One Dark Pro |
| --- | --- |
| Default | One Dark Pro |

| Nebula Aura Violet | Nebula Aura Light |
| --- | --- |
| Nebula Aura Violet | Nebula Aura Light |

## Testing

- 400 client tests and 401 server tests pass; `npm run typecheck`, `npm run lint` and `npm run build` are clean.
- New coverage: the theme context (fixed-appearance palettes driving the `dark` class, a deleted palette falling back without erasing the stored choice), the JSON importer (JSONC, alpha flattening, appearance classification, the two mapping decisions above), the `.vsix` importer (multi-theme extensions, `include` chains, manifest labels, partial failures) and the download service (URL resolution, redirect allowlist, SSRF refusal, size ceiling).
- Verified by hand against real extensions: One Dark Pro (5 themes) from the Marketplace and Nebula Aura (4 themes) from Open VSX, both by URL and as a downloaded `.vsix`. The screenshots above are that session.

## Known limitations

- **Syntax highlighting is not themed.** Chat code blocks read `--cc-syntax-*`, derived at runtime from two fixed Prism themes. Mapping an imported theme's `tokenColors` onto them is the obvious follow-up, and it touches the chat module, so it is kept out of this change.
- **~139 `text-white` and ~300 `dark:bg-gray-*` occurrences are hardcoded in components**, which caps how far any palette can go. It shows on a light palette with a pale accent: the send button's arrow stays white instead of following `--primary-foreground`. Worth a separate pass over those classes.

## Question for the maintainers

The issue asked whether theming should live in core or ship as plugins. This is the core-first version, since plugins currently have no way to touch global styles. If you would rather core only exposed the registry and the hook, the split point is `src/shared/themes/index.ts` and the import modules next to it — happy to rework it that way.

Closes #1105

## Summary by CodeRabbit

* **New Features**
  * Added customizable color themes, including built-in palettes and VS Code theme imports from files or extension URLs.
  * Added theme selection, removal, import progress, and error messaging in Appearance settings.
  * Added authenticated downloads from supported theme marketplaces.
  * Applied saved themes before the app’s first render to reduce visual flashing.
  * Added localized theme settings text across supported languages.

* **Bug Fixes**
  * Theme-controlled appearances now disable dark-mode toggles when appropriate.

* **Tests**
  * Expanded coverage for theme importing, downloads, validation, and preferences.


carsso:
Thanks for the review — four of the five findings were real and are fixed in cfd209a. Details below, including the one I'd rather not take.

### Fixed

**1. Uncontrolled buffering on download** (`themes.service.ts`) — correct, and worse than the size check alone suggested: `clearTimeout` fired in a `finally` as soon as the headers arrived, so the body was read with no deadline at all. The response is now pulled through `response.body.getReader()` with a running byte count that aborts past the ceiling, and the timer stays armed until the stream closes. The `Content-Length` check is kept as a cheap early exit, but the streamed count is what enforces the limit. New test: an endless `ReadableStream` with no `Content-Length`, against a ceiling injected through the service dependencies so the case runs in milliseconds.

**2. Uncompressed extraction budget** (`vsixThemeImport.ts`) — correct. `data.byteLength` bounds only the compressed archive. Each entry this reads is now refused past 8 MB: before decompression when jszip's `_data.uncompressedSize` is available, and again on the resulting string, since that declared size is the archive's own claim about itself. Only the manifest and the contributed theme files are ever read, so nothing else in the archive is materialised. New test: 9 MB of a repeated character, which compresses to a few kilobytes.

**3. Trailing-comma removal inside strings** (`vscodeThemeImport.ts`) — correct, and reproduced before fixing: `{"name":"Night,}"}` imported as `Night}`. The blanket regex is replaced by a string-aware pass, so only commas that actually precede a closing brace or bracket are dropped. New test with that exact input.

**4. Malformed persisted theme records** (`themes/index.ts`) — correct. `importedThemes` round-trips through the server as opaque JSON, so a `null` in the array was read for its `id` during the first render, before Settings could be opened to remove it. Entries are now validated on read — id, name, appearance, preview colours, and the token record when present — and anything else is dropped. New test with a mix of `null`, a string, a valid theme and a record with a numeric id.

### Not taking: conflict-safe `importedThemes` writes

The observation is accurate — `addImportedTheme` writes the whole array, and two devices importing inside the 400 ms debounce window will lose one of the two imports.

I'd rather not fix it here, because this is the semantics of *every* preference in this app: `codeEditorSettings`, `uiPreferences`, `claudePermissions` and the rest are all blobs the server replaces wholesale for a given key. Adding a revisioned read-modify-write with conflict retry for the theme list alone would make it the only preference with its own concurrency protocol, and would be a fair amount of machinery for a window of a few hundred milliseconds on a list a user edits by hand.

If the maintainers want last-write-wins addressed, it seems worth doing once for the whole preference store rather than for this key — happy to open that separately if there's interest.

### Verification

403 client tests, 402 server tests (1 skipped), `npm run typecheck`, `npm run lint` and `npm run build` all clean.

ovhgalan:
really need this

carsso:
Rebased onto `main` (7015ffc) — one conflict, on the "Toggle theme" command item, where #1192's `t('commandPalette.toggleTheme')` met this branch's `canToggleDarkMode` guard; resolved by keeping both.

PR #1249, opened 2026-09-02
fix(chat): keep an option label containing ", " as one answer

materemias:
Question/answer cards split a multi-select answer on ", " unconditionally,
so a single option whose own label contains ", " ("Yes, always") rendered as
two chips, neither matching an option, both marked (custom).

Match one exact option first and split only when no option matches, which
leaves genuine multi-select answers unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved answer display for options whose labels contain commas.
  * Corrected multi-select answer handling when no exact option match exists.

* **Tests**
  * Added regression coverage for comma-containing option labels and unmatched multi-select answers.

### DIFF
--- a/src/modules/chat/tests/QuestionAnswerContent.test.tsx
+++ b/src/modules/chat/tests/QuestionAnswerContent.test.tsx
@@ -77,3 +77,31 @@ test('still renders a well-formed question + answer', () => {
   );
   assert.ok(html.includes('Pick one?'));
 });
+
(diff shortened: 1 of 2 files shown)


blackmammoth (approved):
Hey @materemias, thanks for the PR. It's been merged. Also, if you’d like to join project discussion or community chat, we have a Discord here: [[link](https://discord.gg/buxwujPNRE)](https://discord.gg/buxwujPNRE). Totally optional!

PR #1259, opened 2026-09-05
fix: explain Codex writer conflicts in web chat and allow terminal handoff

MayMrZ:
Web chat can fail with `thread ... already has an active writer` after a Codex reply has finished. An interactive Codex process still owns the thread while waiting for input. CloudCLI's Terminal Disconnect only detaches the WebSocket and retains that PTY for 30 minutes, so the completed reply/disconnected UI does not mean the writer was released.

This change adds an actionable writer-conflict explanation in the web chat transcript, with the original error available in collapsed details. For Codex terminals, the Disconnect action becomes explicitly labeled **End terminal**, with a warning that running work will be interrupted. It requests PTY termination and waits for the process-exit acknowledgement before disconnecting. Ordinary network disconnects keep their existing reconnect behavior.

On POSIX, launch Codex with `exec` so the terminal process is the thread writer. Remove the `resume || codex` fallback (and its Windows counterpart), which otherwise silently starts a new conversation when resume fails. Termination is restricted to the socket currently owning that PTY; a stale socket cannot stop its replacement. No lock files are deleted and no external Codex clients are killed automatically.

Reproduction:

1. Resume a Codex session in CloudCLI Terminal and let a reply finish.
2. Switch to Chat (or disconnect the terminal) and send into the same session.
3. Codex rejects the second writer. Previously this appeared as a raw CLI error with no explanation that a finished interactive session still owns a lock.
4. With this change, Chat explains the conflict. Reconnect to Terminal, choose End terminal, wait for disconnection, and resend in Chat. External clients must be exited in their own interface.

Validation:

- Backend shell tests, including a real Linux PTY holding an OS file lock: disconnect preserves the lock; explicit termination releases it.
- Frontend tests cover the web-chat conflict display, ordinary errors, disconnect versus termination, and waiting for exit acknowledgement.
- Full frontend suite, build, typecheck, and lint (existing lint warnings remain).

English and Simplified Chinese copy is included; other locales use English fallbacks. This is an explicit handoff, not automatic termination on chat send or a general cross-client lock manager. Windows process-tree behavior is not covered by the Linux integration test.

## Summary by CodeRabbit

* **New Features**
  * Added an option to explicitly end Codex terminal sessions so they can be continued in chat.
  * Added clear guidance when a Codex session remains active, including recovery steps and expandable technical details.
  * Added translated messaging in English and Simplified Chinese.

* **Bug Fixes**
  * Prevented reconnects while a terminal is shutting down.
  * Ensured terminal shutdown completes before the connection closes.
  * Preserved terminals during ordinary disconnects while restricting termination to the active connection.

### DIFF
--- a/server/modules/websocket/services/shell-websocket.service.ts
+++ b/server/modules/websocket/services/shell-websocket.service.ts
@@ -29,6 +29,7 @@ type PtySessionEntry = {
   timeoutId: NodeJS.Timeout | null;
   projectPath: string;
   sessionId: string | null;
+  terminating?: boolean;
 };
 
 const ptySessionsMap = new Map<string, PtySessionEntry>();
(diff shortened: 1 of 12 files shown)


blackmammoth:
@MayMrZ, please address the comments and make sure the rest has no issues.

Issue #840, opened 2026-06-06
OpenCode Go Model dropdown/list issue


SiberTutor:
**Is your feature request related to a problem? Please describe.**
The model selection dropdown under the OpenCode provider currently lists standard mainstream models like ChatGPT, Codex, Gemini, and Claude. However, these are models that should belong to their own separate providers. The actual OpenCode provider models are completely missing from the list.

**Describe the solution you'd like**
I would like the UI to properly map and list the actual OpenCode lineup. When OpenCode is selected, it should display models like Minimax, DeepSeek, Qwen, etc., instead of showing the generic mainstream models from other providers.

**Additional context**
This issue specifically occurs when using the OpenCode Go environment/backend. Almost none of the true OpenCode models are being loaded or displayed in the UI model selection dropdown, leaving users unable to select the intended open-source or custom infrastructure models.

larizzo:
This is still broken on main: the curated catalog ships no `opencode-go/*` entries, so a Go subscriber counts as authenticated while the connected-provider filter falls back to the full catalog - a picker full of models the CLI rejects and none of the Go models. A fix is up in #1265: it adds the 27 `opencode-go` models `opencode models --verbose` reports, with reasoning-variant metadata wired as effort values so the effort picker works through `--variant` as well.

PR #1265, opened 2026-09-05
fix(opencode): list OpenCode Go models with reasoning variants

larizzo:
## How to reproduce

1. Connect OpenCode Go (e.g. `opencode auth login`, or `/connect` → `OpenCode Go` in the TUI). `~/.local/share/opencode/auth.json` gains an `opencode-go` entry, and CloudCLI reports the opencode provider as authenticated.
2. Open the OpenCode model picker: every Go model is missing. On a Go-only install the connected-provider filter finds no `opencode-go` options and falls back to the full catalog, so the picker lists Zen/Anthropic/OpenAI models the CLI refuses outright (`Model opencode/claude-sonnet-4-6 is not valid`).
3. Equivalently, `GET /api/providers/opencode/models` returns `authenticated: true` with zero `opencode-go/*` OPTIONS.

## What this fixes

The OpenCode provider authenticated any credentials-file entry, so a subscriber who connected OpenCode Go (`opencode-go` in `auth.json`) counted as logged in — but the curated catalog shipped zero `opencode-go/*` entries. The connected-provider filter then found no Go options, fell back to the full catalog, and left Go users staring at a list of models their CLI rejects while every model they pay for was missing. #840 reported exactly this and was closed without a fix; the bug persists on main.

## What changed

- Add the 27 `opencode-go/<model-id>` entries `opencode models --verbose` reports (per the [OpenCode Go docs](https://opencode.ai/docs/go/)).
- Wire each model's reasoning variants from the same CLI output as effort metadata, so the existing effort picker resolves them through `--variant` (the runtime already validates and passes it).
- Labels follow the catalog's existing style; the CLI's marketing suffixes ("(2x usage)", "(New)") are dropped.
- `DEFAULT` stays `opencode/gpt-5.6-terra`; the filter already relocates the default to `opencode-go/grok-4.6` on Go-only installs.
- The docs also list `minimax-m2.5`, but the CLI does not report it, so it is deliberately left out; the catalog mirrors CLI output.

## What about screenshots

No UI markup changed — the picker renders whatever `OPTIONS` the models API returns; the fixtures and the API response are the verification surface.

## Tests

- The curated-catalog provider-set assertion gains `opencode-go`, with spot checks: 27 Go entries, effort values exactly as the CLI reports, and the models without variants carrying no effort field.
- New fixture: an auth store with only `opencode-go` resolves the whole catalog to Go models, with the default on the first one and `getCurrentActiveModel` agreeing.
- `npm test` (397 server tests pass), `npm run typecheck`, `npm run lint:server`, and `npm run build` all pass.

## Notes

- No file overlap with #1230 (per-model sampling settings); reviewed for conflicts.
- Users who manually added Go models through the model library may see duplicates once these entries are predefined; deduping custom vs. predefined rows is deliberately out of scope for this PR.

## Summary by CodeRabbit

- **New Features**
  - Added OpenCode Go subscription models to the predefined model catalog.
  - Added configurable effort levels for supported OpenCode Go models.
  - OpenCode Go authentication now resolves the available Go model catalog, with `grok-4.6` as the default.

- **Tests**
  - Added coverage validating OpenCode Go model availability, descriptions, effort settings, and default selection.

### DIFF
--- a/server/modules/providers/list/opencode/opencode-models.provider.ts
+++ b/server/modules/providers/list/opencode/opencode-models.provider.ts
@@ -21,9 +21,9 @@ import {
  * Curated OpenCode catalog shipped as immutable CloudCLI defaults.
  *
  * OpenCode routes by `<providerID>/<modelID>`, so this list mirrors the
- * providers `opencode models --verbose` reports: the OpenCode Zen gateway plus
- * the Anthropic and OpenAI providers OpenCode can address directly with the
- * user's own credentials.
+ * providers `opencode models --verbose` reports: the OpenCode Zen gateway, the
+ * OpenCode Go subscription gateway, and the Anthropic and OpenAI providers
+ * OpenCode can address directly with the user's own credentials.
  */
 export const OPENCODE_PREDEFINED_MODELS: ProviderModelsDefinition = {
   OPTIONS: [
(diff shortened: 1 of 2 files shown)


blackmammoth (approved):
Hey @larizzo, thanks for the PR. It's been merged. Also, if you’d like to join project discussion or community chat, we have a Discord here: [[link](https://discord.gg/buxwujPNRE)](https://discord.gg/buxwujPNRE). Totally optional!

PR #1275, opened 2026-09-07
fix(codex): list models from model_catalog_json when present

### LINKED ISSUE #901: When using Codex, only GPT models can be switched via the model switching feature. However, it should ideally support switching any models listed in the model_catalog_json field within the configuration file.
When using Codex, only GPT models can be switched via the model switching feature. However, it should ideally support switching any models listed in the model_catalog_json field within the configuration file.

libra0037:
## Problem

Closes #901

Codex reads custom model definitions from the `model_catalog_json` file named in `~/.codex/config.toml`, and its TUI `/model` picker lists them. CloudCLI's Codex model picker only ever showed the curated predefined catalog, so models defined in the catalog (for example `deepseek/deepseek-v4-flash` or `ainet/gpt-6-astra`) could not be selected in the web UI even though Codex itself accepted them.

## Change

- `CodexProviderModels.getSupportedModels()` now reads `~/.codex/config.toml`. When `model_catalog_json` is set and the file parses cleanly, it returns the catalog's entries as the provider model list — including `display_name`, `description`, `default_reasoning_level` and `supported_reasoning_levels` — and skips entries the catalog hides (`visibility = "hide"`), mirroring the Codex TUI.
- Any failure (missing config, unparseable TOML/JSON, missing catalog file, empty model list) falls back to the curated predefined catalog so the picker never comes up empty.
- When the config sets `model` to a value the catalog does not declare, that model is appended and used as the catalog default; otherwise the curated default is kept when present.
- Models added by users through the UI model library are unaffected: they are merged by `providerModelsService` on top of whichever list the adapter returns.
- The curated catalog now also includes `gpt-6-astra` to match the latest Codex `models.json` entries, using the label style of the existing list.

## Validation

- New unit tests for the adapter cover fallback behavior, metadata mapping, hidden/duplicate/slug-less entries, and default-model derivation (`server/modules/providers/tests/codex-models.test.ts`).
- `provider-models.service.test.ts` and `provider.routes.test.ts` still pass.
- Server typecheck (`tsc -p server/tsconfig.json`) and `oxlint` are clean.
- Smoke-tested against a real `~/.codex/config.toml` + catalog: the picker list matches the catalog and the configured `model` becomes the default.

## Summary by CodeRabbit

- **New Features**
  - Added support for the `gpt-6-astra` Codex model.
  - Codex model availability now reflects the configured model catalog when provided.
  - Reasoning options and default selections are derived from catalog information.
  - Configured models are included even when missing from the catalog.

- **Bug Fixes**
  - Improved resilience by falling back to the curated model list when configuration or catalog data is unavailable or invalid.
  - Hidden, duplicate, and incomplete catalog entries are excluded.


blackmammoth:
intentional. If you wanna add models, you can do it by clicking "Add models"

PR #1282, opened 2026-09-07
fix(chat): keep transcript scroll and history writes owned by the visible session

materemias:
Two transcript scroll writers act without checking who owns the viewport: the auto-follow, which recomputes "is the user reading?" from raw distance on every scroll event, and in-flight history requests, which apply their page into whatever transcript is on screen when they land.

## 1. A short reading gesture is cancelled by its own scroll event

`handleScroll` recomputes `isUserScrolledUp` from `isNearBottom()`, whose window is 50px. A deliberate upward gesture smaller than that leaves the flag `false`, so the next arriving row scrolls the view back to the bottom while the user is reading.

### Reproduction

1. Open a session with enough messages to scroll (60 short rows here).
2. Let it settle at the bottom.
3. Wheel **up 30px** — small, deliberate, still inside the 50px window.
4. Let one new message arrive.

The view snaps back to the bottom mid-read.

Chromium, 1280×900, one message appended between the two reads:

| | after the 30px gesture | after the message arrives |
|---|---|---|
| `main` | `scrollTop` 1252, 30px from bottom | `scrollTop` 1345, **0px from bottom** |
| this PR | `scrollTop` 1191, 30px from bottom | `scrollTop` **1191**, 93px from bottom |

On `main` the reader's position is discarded. With the PR `scrollTop` does not move at all; the distance grows only because the new row extended the content below.

Following re-arms when the viewport returns to the true bottom, so the "scroll to bottom" affordance still hands control back — verified in the same session: click it, `scrollTop` 1191 → 1284 (0px from bottom), and the next appended message is followed again.

## 2. An in-flight history page writes into the transcript that replaced it

A page requested for session A is applied when it resolves, even if the user has since switched to B, or left and re-entered A. The store had no way to ask whether the request still belongs to the visible transcript.

### Reproduction

1. Open session A and scroll to the top to start an older-history page.
2. Before it resolves, switch to session B, then back to A.

Holding that one request (`?limit=20&offset=20`) at the network layer makes it deterministic. On `main` the re-entered transcript stays parked at the top carrying the previous transcript's scroll state (`scrollTop` 0), and when the retired page lands the view jumps to `scrollTop` 1303. With this PR the re-entered transcript opens at the bottom (`scrollTop` 1107) and releasing the retired page contributes none of its rows: wrapper count stays 20, mounted rows stay 20, and the first rendered row stays `ALPHA-45` instead of becoming one of the older rows that page carried.

A no-release control pins what is and is not attributable to it. Holding the page and never releasing it leaves geometry byte-identical across the same interval (`scrollHeight` 1783, `scrollTop` 1107, both unchanged). Releasing it moves both by 53px (1836 / 1160) with the rendered row set unchanged, and the network log shows a further `?limit=20&offset=0` tail refresh firing after the release. That 53px is the follow-up refresh re-rendering A's tail, not the retired page's content entering the transcript.

## The change

- `useChatSessionState` derives a transcript generation from project + session identity and exposes `ownsTranscript(generation, identity, sessionId)`. Every async continuation — history pages, the latest-slot refresh, token usage, search jumps, reconnect refresh — checks it before touching state.
- `useSessionStore` request options take a `canRequest` predicate, so a retired request stops before its next hop instead of being filtered afterwards.
- An explicit upward gesture is recorded separately from the distance heuristic, and cleared when the viewport is back at the bottom.
- `ChatMessagesPane` distinguishes the two things the pane was conflating: `onScrollCheck` (any scroll, bookkeeping) and `onReadGesture` (deliberate upward wheel or touch drag).

## Tests

`src/modules/chat/tests/transcriptScrollOwnership.test.tsx`

- `does not yank the view back down after the user starts reading upward` — drives the gesture **and** the scroll event the gesture emits; fails on `main` with `got [5000]`, the follow write.
- `still sticks to the bottom when the user has not scrolled away`
- `discards an older A page after navigating A to B to A`
- `does not scroll the next session when the previous session refresh finishes`
- `does not follow the user into the next session`

`src/modules/chat/tests/sessionStoreTruncate.test.tsx` covers the store-side `canRequest` contract, including a page whose generation was retired mid-flight.

`npm run typecheck` clean; `npm run test:client` 377 tests in 52 files pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved chat scrolling so reading older messages no longer unexpectedly returns the view to the bottom.
  - Scrolling and touch gestures now more reliably preserve the reader’s position.
  - Prevented outdated message history or reconnect results from appearing after switching sessions or transcripts.
  - Prevented delayed responses from overwriting newer chat state.
  - Improved handling of interrupted or cancelled history loading requests.

### DIFF
--- a/src/modules/chat/ChatInterface.tsx
+++ b/src/modules/chat/ChatInterface.tsx
@@ -141,6 +141,7 @@ function ChatInterface({
     tokenBudget,
     setTokenBudget,
     visibleMessageCount,
+    handleUserScrollGesture,
     visibleMessages,
     loadEarlierMessages,
     loadAllMessages,
@@ -151,7 +152,6 @@ function ChatInterface({
     showLoadAllOverlay,
     createDiff,
     scrollContainerRef,
-    scrollToBottom,
     scrollToBottomAndReset,
     handleScroll,
     requestLatestMessages,
@@ -249,7 +249,6 @@ function ChatInterface({
     onSessionEstablished: handleSessionEstablished,
     onFileOpen,
     onShowSettings,
-    scrollToBottom,
     addMessage,
     setIsUserScrolledUp,
     setPendingPermissionRequests,
@@ -414,19 +413,15 @@ function ChatInterface({
     );
   }
 
-
   return (
     <PermissionContext.Provider value={permissionContextValue}>
       <div className="flex h-full min-h-0 flex-col">
         <ChatMessagesPane
           scrollContainerRef={scrollContainerRef}
-          // Not redundant with the `scroll` listener. A first page is 20 rows,
-          // tool results fold into their calls, and the "load earlier" link is
-          // hidden while more pages exist — so a short transcript is often not
-          // scrollable at all and never emits `scroll`. Wheel and touch are
-          // then the only way to reach the top pager or the "load all" overlay.
-          onWheel={handleScroll}
-          onTouchMove={handleScroll}
+          // A short transcript may not emit `scroll`, so wheel and touch
+          // gestures also run the top-pager check.
+          onScrollCheck={handleScroll}
+          onReadGesture={handleUserScrollGesture}
           isLoadingSessionMessages={isLoadingSessionMessages}
           isProcessing={isProcessing}
           hasActivityIndicator={hasActivityIndicator}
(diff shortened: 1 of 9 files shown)


blackmammoth:
@materemias please use screenshots and don't generate everything with AI.

PR #1283, opened 2026-09-07
fix(ws): reconnect broken chat channels

materemias:
## Purpose

Reconnect the shared `/ws` chat channel when it stops working, then catch the selected chat up on missed messages. Provider implementations, defaults, authentication modes, and the `/shell` and plugin WebSocket transports are unchanged.

Rebased onto upstream `main` at `5e73a49b`. That upstream revision still reproduces the half-open-channel failure described below.

## The failure

A dropped network path can leave the browser reporting `WebSocket.OPEN` without a close event. Upstream reconnects after close, so an apparently connected chat can stop receiving messages. Returning from suspend is one way to encounter this, but recovery must not depend on a visibility or page-show event.

The original version of this PR only probed on those page events. This revision adds periodic liveness checks and fixes the ordering of chat catch-up after reconnect.

## Changes

- Probe with `chat.ping { nonce }` 30 seconds after opening or receiving a matching pong. The gateway answers only that socket with `{ kind: "pong", nonce }`, without invoking a provider. Pong frames stay inside the transport layer.
- If no matching pong arrives within 3 seconds, retire the socket and retry after the existing 3-second reconnect delay. A handshake stuck in `CONNECTING` is retired after 30 seconds. Constructor, socket, and send failures also retire or retry the connection as appropriate.
- Use one auth-scoped timer for handshake, next ping, pong deadline, or retry. Socket identity guards and cleanup prevent retired sockets, old credentials, or pending timers from affecting a replacement. Repeated resume events cannot postpone an outstanding pong deadline.
- Keep `visibilitychange` to visible and `pageshow` as earlier probe triggers. Periodic checks do not require either event or a visibility transition.
- On socket replacement, publish the socket to chat only after it opens, send `chat.subscribe`, and wait for the matching selected-session acknowledgement before requesting the persisted tail. Arm the listener before sending; discard pending acknowledgements when the socket/session changes or the hook unmounts. Remove the obsolete duplicate reconnect subscription callback.
- Preserve hidden-chat history gating: hidden chats still subscribe, but persisted-tail HTTP waits until activation and acknowledgement. Initial connections do not gain an extra reconnect refresh. Visibility-only changes do not resubscribe.

The acknowledgement confirms subscription handling, not replay completion or provider persistence. The server sends replay after the acknowledgement, so replay and the HTTP refresh can overlap. Browser suspension or background timer throttling can delay detection; there is no wall-clock guarantee while browser timers are paused. Failed application messages are not automatically resent.

### Native Chromium reproduction

Used an isolated browser fixture mounting the actual `WebSocketProvider`, with an authentication stub, the actual chat gateway handler, native WebSockets, and a TCP proxy. The proxy dropped both directions for existing connections without closing them, while allowing new connections through. The page stayed visible and no `visibilitychange` or `pageshow` event fired during the observation.

| Source | Replacement connections | Message emitted after blackholing |
| --- | --- | --- |
| Upstream `5e73a49b` | 0 after 51 seconds; still reported connected | Not received |
| This revision | Exactly 1 by the 40-second observation | Received by the browser consumer |

The responsive replacement remained connected through subsequent heartbeat checks. A separate token-rotation check opened one replacement and received a fresh subscription acknowledgement. The final blackhole run also received a new subscription acknowledgement before the emitted message. This was a transport fixture, not a provider-backed conversation or a physical laptop-suspend test.

### Regression and preservation checks

- The no-page-event half-open regression fails against the actual upstream transport source and passes with this revision.
- The delayed-acknowledgement regression fails against the original PR's rebased session hook because it starts HTTP before acknowledgement; it passes with this revision.
- Transport tests exercise healthy pongs, wrong nonces, repeated page events, close during a probe, stale callbacks, ordinary disconnects, failed constructors/sends, stalled handshakes, token replacement, logout, unmount, expired OSS tokens, and tokenless platform mode.
- Chat tests use the real session store to check delayed/unrelated/stale acknowledgements, displayed history, visibility changes, hidden reconnects, activation before acknowledgement, session switching, and unmount.
- A real-handler/store regression verifies that synthetic reconnect notifications never enter transcript history; it failed before restoring the explicit early return and now passes.
- A real-socket gateway test checks that pings receive the matching pong without invoking provider runtimes or broadcasting to another client.

Commands run on the rebased revision:

- `npm run test:client`: 422 passed.
- `npm test`: 416 passed, 1 skipped because the Codex rollout fixture is unavailable.
- `npm run typecheck`: passed.
- `npm run build`: passed, with CSS and bundle-size warnings.
- `npm run lint`: exited successfully with 132 warnings and no errors. Follow-up lint on all touched implementation/test files reports no warnings in the new tests.

## Summary by CodeRabbit

- **New Features**
  - Added WebSocket heartbeat monitoring to detect stalled connections and automatically recover them.
  - Added `chat.ping` messages and matching `pong` responses for connection checks.
  - Chat sessions now wait for successful re-subscription before refreshing message history after reconnecting.
  - New chat turns can optionally interrupt an active provider response and complete it before starting the next turn.

- **Documentation**
  - Updated WebSocket protocol and recovery documentation with heartbeat and reconnection behavior.

### DIFF
--- a/server/modules/websocket/README.md
+++ b/server/modules/websocket/README.md
@@ -33,7 +33,7 @@ Benefits:
 |---|---|
 | `services/websocket-server.service.ts` | Creates `WebSocketServer`, binds `verifyClient`, routes connection by pathname |
 | `services/websocket-auth.service.ts` | Authenticates upgrade requests and attaches `request.user` |
-| `services/chat-websocket.service.ts` | Handles the `/ws` chat protocol (`chat.send` / `chat.abort` / `chat.subscribe` / `chat.permission-response`) |
+| `services/chat-websocket.service.ts` | Handles the `/ws` chat protocol (`chat.send` / `chat.abort` / `chat.subscribe` / `chat.permission-response` / `chat.ping`) |
 | `services/chat-run-registry.service.ts` | Tracks live provider runs per app session id: seq numbering, event replay buffer, provider-id mapping, completion state |
 | `services/chat-session-writer.service.ts` | Gateway writer handed to provider runtimes: remaps provider session ids to app ids, swallows `session_created`, assigns `seq` |
 | `services/shell-websocket.service.ts` | Handles `/shell` PTY lifecycle, reconnect buffering, auth URL detection |
@@ -133,6 +133,7 @@ flowchart TD
   D -->|chat.abort| F[providerRuntimeService.abort + synthetic complete]
   D -->|chat.subscribe| G[chat_subscribed ack + attach socket + replay events seq > lastSeq]
   D -->|chat.permission-response| H[providerRuntimeService.resolveToolApproval]
+  D -->|chat.ping| J[send kind:pong with matching nonce]
   D -->|other| I[send kind:protocol_error]
 ```
 
--- a/server/modules/websocket/README.md
+++ b/server/modules/websocket/README.md
@@ -108,7 +108,7 @@ When a chat socket connects:
 
 1. Add socket to `connectedClients`.
 2. Parse each incoming message with `parseIncomingJsonObject`.
-3. Dispatch by `data.type` (four message types, none provider-specific).
+3. Dispatch by `data.type`; chat commands are provider-neutral.
 4. On close, remove socket from `connectedClients`.
 
 ### Session identity model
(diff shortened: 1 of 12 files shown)


materemias:
Superseded by #1301, a clean single-commit replacement on current upstream main. It carries the reviewed reconnect fixes and adds bounded recovery for a missing subscription acknowledgement. Please continue review there; this discussion is retained for reference.

PR #1301, opened 2026-09-09
fix(ws): recover broken chat connections and subscriptions

materemias:
## Purpose

Supersedes #1283 with one commit based on upstream `main` at `5e73a49b`. This carries the reviewed reconnect fix and adds recovery when a replacement socket never acknowledges its chat subscription.

Restore the shared `/ws` chat connection when it stops working, then restore the selected chat's subscription and persisted history. Provider startup, abort behavior, defaults, authentication modes, `/shell`, and plugin WebSockets are unchanged.

## Behavior

- Send a nonce-bearing `chat.ping` 30 seconds after connection or a matching pong. If no matching pong arrives within 3 seconds, retire the socket and retry after the existing 3-second delay. A handshake stuck for 30 seconds also retries. The gateway rejects missing or non-string nonces with `protocol_error` / `INVALID_NONCE`, without emitting a pong or closing the socket.
- Keep visible-page and `pageshow` probes as faster triggers, but do not depend on those events. Use one transport timer and reject stale socket callbacks; token changes, logout, and unmount clean up the old connection.
- Publish only open sockets to chat. After replacement, subscribe first and wait for the matching `chat_subscribed` acknowledgement before refreshing history. Synthetic reconnect notifications never enter transcript storage.
- Bound the reconnect acknowledgement wait to 10 seconds. If it expires, close that socket and let the existing transport reconnect and resubscribe, even when heartbeats remain healthy. A timer-only HTTP fallback would not restore live subscription, so history still waits for a successful acknowledgement. Persistent server errors keep retrying rather than being treated as success.
- Cancel the acknowledgement deadline on acknowledgement, socket/session change, or unmount. Preserve initial history loading and hidden-chat HTTP gating; visibility-only changes do not resubscribe.

The server sends replay after the acknowledgement, so replay and the history request may overlap. Browser timer suspension/throttling can delay recovery. Failed application messages are not automatically resent. Retry backoff and provider-run interruption changes are outside this PR.

## Verification

**Native Chromium fixtures, using actual client hooks and gateway code:**

- A TCP-blackholed socket on upstream `5e73a49b` stayed apparently connected with no replacement or new message after 51 seconds. The carried transport fix made exactly one replacement by the 40-second observation and received the message, without visibility/page-show events.
- For the new failure path, forced the gateway's pending-approval lookup to throw during reconnect subscription while heartbeats continued working. The previous PR left history blocked. This revision closed the unacknowledged socket, received a fresh subscription acknowledgement, refreshed persisted history, and received a live message through the run registry.
- The new visible/hidden missing-ack regressions fail against the original PR hook and pass with this revision. Tests also cover acknowledgement/deadline cleanup, auth/token rotation, stale callbacks, healthy sockets, initial loading, and hidden-session activation. The session-switch test also fires the old deadline after the new session renders but before passive-effect cleanup; it fails without the session-identity guard and passes with it.

These are isolated fixtures with stubbed authentication/provider execution, not a physical laptop-suspend test or a provider-backed conversation.

The real-socket gateway regression rejects a malformed nonce without a pong, then confirms valid pings still work on the same socket without provider calls or broadcasts. It fails before nonce validation and passes afterward.

**Commands on this branch:**

- `npm run test:client`: 424 passed.
- `npm test`: 416 passed, one Codex fixture-dependent skip.
- `npm run typecheck`: passed.
- `npm run build`: passed, with CSS/bundle-size warnings.
- `npm run lint`: successful exit, 132 warnings and no errors.

## Summary by CodeRabbit

- **New Features**
  - Added WebSocket health checks to detect unavailable connections and recover automatically.
  - Added support for resuming chat sessions after reconnecting, including subscription confirmation and message catch-up.
  - Added reconnection handling when returning to an active browser tab or window.
  - Added support for `chat.ping` requests with corresponding `pong` responses.

- **Bug Fixes**
  - Improved handling of connection timeouts, failed sends, interrupted handshakes, and stale connections.
  - Prevented reconnect notifications from appearing as chat messages.
  - Improved authentication and logout handling during WebSocket recovery.

### DIFF
--- a/server/modules/websocket/README.md
+++ b/server/modules/websocket/README.md
@@ -33,7 +33,7 @@ Benefits:
 |---|---|
 | `services/websocket-server.service.ts` | Creates `WebSocketServer`, binds `verifyClient`, routes connection by pathname |
 | `services/websocket-auth.service.ts` | Authenticates upgrade requests and attaches `request.user` |
-| `services/chat-websocket.service.ts` | Handles the `/ws` chat protocol (`chat.send` / `chat.abort` / `chat.subscribe` / `chat.permission-response`) |
+| `services/chat-websocket.service.ts` | Handles the `/ws` chat protocol (`chat.send` / `chat.abort` / `chat.subscribe` / `chat.permission-response` / `chat.ping`) |
 | `services/chat-run-registry.service.ts` | Tracks live provider runs per app session id: seq numbering, event replay buffer, provider-id mapping, completion state |
 | `services/chat-session-writer.service.ts` | Gateway writer handed to provider runtimes: remaps provider session ids to app ids, swallows `session_created`, assigns `seq` |
 | `services/shell-websocket.service.ts` | Handles `/shell` PTY lifecycle, reconnect buffering, auth URL detection |
@@ -108,7 +108,7 @@ When a chat socket connects:
 
 1. Add socket to `connectedClients`.
 2. Parse each incoming message with `parseIncomingJsonObject`.
-3. Dispatch by `data.type` (four message types, none provider-specific).
+3. Dispatch by `data.type`; chat commands are provider-neutral.
 4. On close, remove socket from `connectedClients`.
 
 ### Session identity model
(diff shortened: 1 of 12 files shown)


blackmammoth:
@materemias please break up the PR into small pieces with a reproduction step of the error for each.

PR #1303, opened 2026-09-09
fix(providers): strip ANSI escapes from OpenCode and Cursor stderr

jjscarafia:
## The bug

When an OpenCode run writes to stderr, `opencode-runtime.provider.js` forwards the chunk to the chat verbatim as a `kind: 'error'` message. OpenCode styles that output for a terminal, so the ANSI escapes are rendered as literal text in the UI:

```
!  Error
^[[93m^[[1m! ^[[0m^[[0mpermission requested: external_directory (/etc/*); auto-rejecting
```

`cursor-runtime.provider.js` surfaces its stderr the same way, so it has the same problem.

### Reproduce

1. Pick OpenCode as the provider and send a prompt that makes the CLI print a warning on stderr — e.g. anything that trips a permission prompt, such as asking it to read a path outside the project.
2. The error bubble in the chat shows the raw escape sequences instead of styled or plain text.

## The fix

- Strip the escapes in both runtimes before the text is sent to the client. A chunk that turns out to have been styling only now cleans down to nothing and is dropped, instead of producing an empty error bubble.
- Four private ANSI strippers had already grown in `server/`, each with a different regex and none of them exported:
  - `websocket/services/shell-websocket.service.ts` (`stripAnsiSequences`)
  - `providers/list/claude/claude-sessions.provider.ts` (`stripAnsiFormatting`)
  - `providers/services/session-conversations-search.service.ts` (`stripAnsiFormatting`)
  - `providers/list/claude/claude-models.provider.ts` (`stripAnsi`)

  One implementation now lives in `server/shared/utils.ts` next to `ANSI_TERMINAL_STYLES`, and every caller uses it. The shared pattern is the union of what those four covered, so no caller loses coverage: OSC (BEL or ST terminated), CSI in both the `ESC [` and 8-bit introducer forms, and the remaining ECMA-48 escape sequences such as `ESC ( B`.

No behavior change for the Claude session/search readers — they get a strictly wider pattern on the same inputs.

## Verification

- `npm test` — 422 tests, 0 failures (the 1 skip is the pre-existing Codex-fixture skip).
- New unit tests in `server/shared/tests/strip-ansi-sequences.test.ts`, including the exact string from the report.
- `npm run typecheck`, `npm run lint` (no new warnings), `npm run build` all clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of terminal formatting codes across Claude, Cursor, OpenCode, and shell output.
  * Chat errors now display clean, readable text without ANSI escape sequences.
  * Empty or formatting-only error output is no longer shown.
  * Claude session and search results consistently remove terminal styling from command output.

* **Tests**
  * Added coverage for terminal color, cursor-control, hyperlink, and styling sequences.

## Notes for review

Three things worth flagging rather than leaving for you to find:

- **One observable behaviour change in the shell WebSocket.** The previous
  alternation order there left `8;;http://example.com` behind when stripping an
  OSC 8 hyperlink (`ESC ] 8 ; ; <url> ST`). That is now removed properly. It is a
  fix, but it means a URL announced *only* through an OSC 8 hyperlink is no longer
  picked up by the terminal's URL detector, since the text is gone along with the
  sequence.
- **The two runtime providers stay as `.js`.** The backend module standards ask for
  a TypeScript migration of files you touch, but they also warn against migrations
  outside the scope of a change, and both files are large. Happy to split that into
  its own PR if you prefer it the other way.
- **Known limitation, unchanged by this PR.** stderr arrives in chunks, so a
  sequence split across two `data` events (`ESC [9` then `3m`) still leaks its tail
  into the UI. Fixing that needs buffering at the transport level, which felt out
  of scope here.

The OSC alternative is bounded on `ESC` as well as `BEL`, so an unterminated OSC
cannot scan to the end of a large chunk (a 40 KB chunk of unterminated OSC
introducers went from ~750 ms to 0 ms) and no longer swallows whatever CSI follows
it. `ST` still matches, since it begins with `ESC`.

### DIFF
--- a/server/modules/providers/list/claude/claude-models.provider.ts
+++ b/server/modules/providers/list/claude/claude-models.provider.ts
@@ -7,7 +7,7 @@ import type {
   ProviderModelOption,
   ProviderModelsDefinition,
 } from '@/shared/types.js';
-import { buildDefaultProviderCurrentActiveModel } from '@/shared/utils.js';
+import { buildDefaultProviderCurrentActiveModel, stripAnsiSequences } from '@/shared/utils.js';
 
 /**
  * Ultracode is not one of the SDK's reasoning-effort levels. Selecting it runs the turn at
@@ -179,13 +179,6 @@ type ClaudeInitEvent = {
   };
 };
 
-const ANSI_PATTERN = new RegExp(
-  '[\\u001B\\u009B][[\\]()#;?]*(?:'
-  + '(?:[0-9]{1,4}(?:;[0-9]{0,4})*)?[0-9A-ORZcf-nqry=><]'
-  + '|(?:[\\dA-PR-TZcf-ntqry=><~]))',
-  'g',
-);
-
 /**
  * Claude Code stamps locally-synthesized rows (API-error placeholders and the
  * like) with `model: "<synthetic>"`. Angle-bracketed values are placeholders,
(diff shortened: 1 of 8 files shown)


blackmammoth (approved):
Hey @jjscarafia, thanks for the awesome PR!! It's been merged. Also, if you’d like to join project discussion or community chat, we have a Discord here: [link](https://discord.gg/buxwujPNRE). Totally optional!

PR #1304, opened 2026-09-10
fix(auth): render the message of a structured error envelope, not the object

### LINKED ISSUE #1300: Wrong password blanks the login screen: the structured error envelope is rendered as a React child
**Describe the bug**

Entering a wrong password on the login screen blanks the app. React throws while rendering the error message, the `<form>` is torn down, and nothing is left on the page — no error text, no way back except a reload.

The two ends of the message do not agree on its shape:

- **Server.** The auth routes hand every failure to `next(error)` (`server/modules/auth/auth.routes.ts`), and the global error middleware serialises an `AppError` as a structured envelope (`server/index.ts:254`):
  ```json
  {"success":false,"error":{"code":"AUTH_INVALID_CREDENTIALS","message":"Invalid username or password"}}
  ```
- **Client.** `resolveApiErrorMessage` (`src/modules/auth/context/AuthContext.tsx:78`) returns `payload.error ?? payload.message ?? fallback`. Its return type says `string`, but `ApiErrorPayload` declares `error?: string` and `parseJsonSafely` only casts the parsed body — nothing checks it at runtime, so the **object** is returned.

That object travels through `setError` / `{ success: false, error }` into `LoginForm.tsx:53` and reaches `AuthErrorAlert.tsx` as `<p>{errorMessage}</p>`. `errorMessage ||` does not help — an object is truthy. There is no error boundary above `<ProtectedRoute>` in `src/App.tsx`, so the throw unmounts the tree.

Every business failure of the auth service takes this path — invalid credentials, missing fields, password too short, username conflict, and any unexpected 500. The errors raised by `auth.middleware.ts` (missing/expired/invalid token) still use `{ error: 'string' }` and render correctly, which is probably why this has gone unnoticed: only some auth errors blank the screen.

**To Reproduce**

1. Run the OSS server with auth enabled and a registered user.
2. Open the app logged out, so the login form shows.
3. Enter the right username and a wrong password, and submit.
4. The form disappears and the page is blank.

**Expected behavior**

The login form stays on screen and shows `Invalid username or password`.

**Error message**

Production build:

```
Minified React error #31; ... args[]=object with keys {code, message}
```

Development build:

```
Objects are not valid as a React child (found: object with keys {code, message}).
If you meant to render a collection of children, use an array instead.
```

Measured in the same run, on a clean instance (`v1.37.x`, `main`): `form` count before submit `1`, after submit `0`, and the body text is empty. The server answered `401` with the structured envelope quoted above.

**Desktop**
 - OS: Linux (Debian 13)
 - Browser: Chromium 140 (headless), production build served by the app's own server
 - Version: `main` (verified with `git show origin/main:...` — the code in question is unchanged there)

**Additional context**

Possible fix, at the boundary that already knows about both shapes:

```ts
type ApiErrorPayload = {
  error?: string | { code?: string; message?: string; details?: unknown };
  message?: string;
};

function resolveApiErrorMessage(payload: ApiErrorPayload | null, fallback: string): string {
  const candidate = typeof payload?.error === 'object' && payload.error !== null
    ? payload.error.message
    : payload?.error;

  if (typeof candidate === 'string' && candidate.trim()) return candidate;
  if (typeof payload?.message === 'string' && payload.message.trim()) return payload.message;
  return fallback;
}
```

`SetupForm` reaches the same helper through `register`, so it is covered by the same change. Elsewhere the client already unwraps this envelope correctly — `src/modules/git-panel/hooks/useWorktreesController.ts:39` reads `payload.error?.message` — and `readApiJson` in `src/shared/api.ts` handles both shapes; auth predates it.

Two things that would make this class of bug harmless rather than fatal, if you want them:
- have `AuthErrorAlert` accept `unknown` and render only strings, so no caller can blank the login screen;
- put an error boundary above `<ProtectedRoute>` in `src/App.tsx`.

Related: #1109 reports the same envelope crashing the render in the plugin-install flow, which suggests the contract, not this one call site, is what is off.

I'm happy to send a PR for the `resolveApiErrorMessage` fix if that's useful.

L4XB:
## Summary

Fixes #1300.

Entering a wrong password blanked the login screen. The auth routes hand failures to the global error middleware, which serialises an `AppError` as `{ success: false, error: { code, message } }`, but `resolveApiErrorMessage` returned `payload.error` as-is. The envelope **object** travelled through `login()` into `LoginForm` and reached `AuthErrorAlert` as a React child, React threw while rendering `<p>{errorMessage}</p>`, and with no error boundary above `ProtectedRoute` the whole tree unmounted. The auth middleware's bare `{ error: 'string' }` responses rendered fine, which is why only some auth errors blanked the screen.

## Change

- `src/modules/auth/context/apiErrorMessage.ts` (new): `resolveApiErrorMessage` now understands both shapes: a string `error`, an `error.message` inside the envelope, or a top-level `message`, and always returns a string (empty strings and envelopes without a message fall back to the caller's default). `ApiErrorPayload` documents both shapes.
- `AuthContext.tsx` imports the helper instead of defining it; behaviour of `login` / `register` is otherwise unchanged. Moving the pure helper out of the provider file keeps the component file free of non-component exports (the fast-refresh lint rule) and makes it testable without React.

## Tests

`src/modules/auth/tests/resolveApiErrorMessage.test.ts` (vitest): the structured envelope (`AUTH_INVALID_CREDENTIALS`), the bare string shape, a top-level message, and the always-a-string guarantee (`null`, `{}`, empty strings, an envelope without `message`). Two cases fail on `main`.

`npx vitest run …/resolveApiErrorMessage.test.ts` → 4 passed; `oxlint src/modules/auth/` reports nothing for the changed files; `tsc --noEmit` clean for the auth module.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved authentication error-message handling for structured responses, plain-text errors, empty values, and incomplete responses.
  - Whitespace-only messages are now ignored, while valid messages are trimmed for clearer display.
  - Login and registration continue to show an appropriate fallback message when no usable error details are available.

- **Tests**
  - Added coverage for supported authentication error formats, whitespace handling, message trimming, and fallback scenarios.

### DIFF
--- a/src/modules/auth/context/AuthContext.tsx
+++ b/src/modules/auth/context/AuthContext.tsx
@@ -7,6 +7,7 @@ import { api } from '@/shared/api';
 import { AUTH_SESSION_EXPIRED_EVENT, AUTH_TOKEN_REFRESHED_EVENT, getAuthTokenRefreshDelay, isValidRefreshedToken, storeAuthToken } from '@/shared/authToken';
 import { hydrateChatDrafts, resetChatDrafts } from '@/shared/chatDrafts';
 import { hydrateUserPreferences, resetUserPreferences } from '@/shared/userSettings';
+import { resolveApiErrorMessage, type ApiErrorPayload } from '@/modules/auth/context/apiErrorMessage';
 /** The signed-in account held by AuthContext - a required `username` plus an optional id and any additional fields the auth API returns - and should be read through `useAuth()` rather than re-derived from raw auth responses. */
 type AuthUser = {
   id?: number | string;
@@ -45,10 +46,6 @@ type OnboardingStatusPayload = {
   hasCompletedOnboarding?: boolean;
 };
 
-type ApiErrorPayload = {
-  error?: string;
-  message?: string;
-};
 
 type AuthContextValue = {
   user: AuthUser | null;
@@ -73,14 +70,6 @@ async function parseJsonSafely<T>(response: Response): Promise<T | null> {
   } catch {
     return null;
   }
-}
-
-function resolveApiErrorMessage(payload: ApiErrorPayload | null, fallback: string): string {
-  if (!payload) {
-    return fallback;
-  }
-
-  return payload.error ?? payload.message ?? fallback;
 }
 
 const AuthContext = createContext<AuthContextValue | null>(null);
--- a/src/modules/auth/context/apiErrorMessage.ts
+++ b/src/modules/auth/context/apiErrorMessage.ts
@@ -1,0 +1,34 @@
+/**
+ * Error bodies the auth endpoints produce. The auth middleware answers with a
+ * bare `{ error: 'string' }`, while the global error middleware serialises an
+ * `AppError` as `{ success: false, error: { code, message, details } }`.
+ */
+export type ApiErrorPayload = {
+  error?: string | { code?: string; message?: string; details?: unknown };
+  message?: string;
+};
+
+/**
+ * Picks the human-readable message out of an auth error body, whatever its
+ * shape. Returning the structured `error` object itself sent it into
+ * `AuthErrorAlert` as a React child, which throws and unmounts the login
+ * screen.
+ */
+export function resolveApiErrorMessage(payload: ApiErrorPayload | null, fallback: string): string {
+  if (!payload) {
+    return fallback;
+  }
+
+  const { error, message } = payload;
+  if (typeof error === 'string' && error) {
+    return error;
+  }
+  if (error && typeof error === 'object' && typeof error.message === 'string' && error.message) {
+    return error.message;
+  }
+  if (typeof message === 'string' && message) {
+    return message;
+  }
+
+  return fallback;
+}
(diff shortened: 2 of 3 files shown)


src/modules/auth/context/AuthContext.tsx
```diff
@@ -7,6 +7,7 @@ import { api } from '@/shared/api';
 import { AUTH_SESSION_EXPIRED_EVENT, AUTH_TOKEN_REFRESHED_EVENT, getAuthTokenRefreshDelay, isValidRefreshedToken, storeAuthToken } from '@/shared/authToken';
 import { hydrateChatDrafts, resetChatDrafts } from '@/shared/chatDrafts';
 import { hydrateUserPreferences, resetUserPreferences } from '@/shared/userSettings';
+import { resolveApiErrorMessage, type ApiErrorPayload } from '@/modules/auth/context/apiErrorMessage';
```
blackmammoth: `ApiErrorPayload` is unused. Please make sure the setup is correct.

PR #1310, opened 2026-09-10
fix(auth): verify the session once per mount, not once per token refresh

TadMSTR:
Resubmission of #1177, as requested there.

## Why #1177 stopped applying

@blackmammoth was right that its diff no longer applies. The cause is a file move, not a fix: #1206 moved `src/components/**` to `src/modules/**`, so `src/components/auth/context/AuthContext.tsx` — the file #1177 patched — no longer exists at that path. It is now `src/modules/auth/context/AuthContext.tsx`.

The code shape it fixes survived the move unchanged. Verified against `main` at `5e73a49b`:

| Link in the loop | Location on current `main` |
|---|---|
| Server sets `X-Refreshed-Token` on any authenticated response past half-life | `server/modules/auth/auth.middleware.ts:78` |
| Client reads the header and pipes it into `storeAuthToken()` | `src/shared/api.ts:40-42` |
| `storeAuthToken()` dispatches `AUTH_TOKEN_REFRESHED_EVENT` | `src/shared/authToken.ts:97` |
| The listener calls `setToken(nextToken)` | `src/modules/auth/context/AuthContext.tsx:192` |
| **`checkAuthStatus` still lists `token` in its dependency array** | `src/modules/auth/context/AuthContext.tsx:247` |
| **The mount effect still depends on `checkAuthStatus`** | `src/modules/auth/context/AuthContext.tsx:260` |

Every refresh rebuilds `checkAuthStatus`, which re-runs the mount effect, which re-verifies the session — issuing three more authenticated requests, which can themselves refresh the token.

## Why it is normally invisible

`iat` has one-second resolution, so same-second reissues are byte-identical and React's `useState` bails on an `Object.is` match without re-rendering. The loop only runs once something makes the token *alternate* between two values — an HTTP-cached `X-Refreshed-Token` replaying an older value does exactly that.

The user-visible symptom is not the request flood. Each pass calls `setUser()` with a freshly parsed object and toggles `isLoading`, which tears down and reopens the chat WebSocket — so it surfaces as a reconnect storm.

## The change

Two edits, both in `AuthContext`:

- the `if (!token)` guard reads `readStoredToken()` instead. `authenticatedFetch` sends whatever storage holds, so the guard now asks exactly the question the request will answer — and `token` leaves the callback's dependencies.
- `token` is dropped from `checkAuthStatus`'s dependency array.

## Test

`src/modules/auth/tests/authVerificationReentrancy.test.tsx`, a vitest test matching the existing client suite. It mounts the provider against a stub server whose `/api/auth/user` alternates the refreshed token it returns.

Measured on this branch, not carried over from the original PR:

| `checkAuthStatus` deps | `/api/auth/user` requests per mount |
|---|---|
| with the fix | **1** |
| with `token` restored to the array | **11** |

11 rather than unbounded because the stub only offers a refreshed token 10 times and then settles — so a regression fails with a count instead of hanging CI.

Full client suite: 57 files / 398 tests pass. `typecheck` and `oxlint` clean (no new warnings on the touched file).

## Both review notes from #1177 are addressed

- *"the test-only renderer should remain out of production dependencies"* — the original PR predated `test:client`/vitest and reached for `react-test-renderer` + `node:test`. Rewriting it as a vitest test removes **both** `package.json` additions. This PR touches no dependency at all.
- *"the test setup should unmount the provider before restoring mocked browser globals"* — it now calls `cleanup()` explicitly before `vi.unstubAllGlobals()`. Vitest's default hook order is `stack`, so the `afterEach(cleanup)` in `vitest.setup.ts` runs *after* a test file's own `afterEach`; a provider left mounted would keep firing effects against a restored `fetch`.

The `src/shared/utils.ts` hunk from #1177 is also dropped — #1174 landed the same `import.meta.env?.VITE_IS_PLATFORM` guard.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented repeated authentication checks when session tokens refresh.
  - Improved session verification reliability during token updates.

- **Tests**
  - Added regression coverage to verify authentication checks run once per mount and are not retriggered by token refreshes.

### DIFF
--- a/src/modules/auth/context/AuthContext.tsx
+++ b/src/modules/auth/context/AuthContext.tsx
@@ -220,7 +220,11 @@ export function AuthProvider({ children }: AuthProviderProps) {
 
       setNeedsSetup(false);
 
-      if (!token) {
+      // Read the token from storage rather than from state. `authenticatedFetch`
+      // sends whatever storage holds, so this guard now asks exactly the question
+      // the request will answer — and, critically, keeps `token` out of this
+      // callback's dependencies. See the effect below for why that matters.
+      if (!readStoredToken()) {
         return;
       }
 
@@ -244,8 +248,26 @@ export function AuthProvider({ children }: AuthProviderProps) {
     } finally {
       setIsLoading(false);
     }
-  }, [checkOnboardingStatus, clearSession, t, token]);
+  }, [checkOnboardingStatus, clearSession, t]);
 
+  // Verifies the stored session once per mount.
+  //
+  // `checkAuthStatus` must not depend on `token`. `authenticateToken()` re-issues
+  // a token through `X-Refreshed-Token` on *every* authenticated request once the
+  // current one is past half-life, and `authenticatedFetch` feeds that straight
+  // into `storeAuthToken()`, which dispatches `AUTH_TOKEN_REFRESHED_EVENT` and so
+  // reaches `setToken()` in the listener above. If this callback were rebuilt on
+  // each token change, this effect would re-run and re-verify the session, issuing
+  // three more authenticated requests that can themselves refresh the token —
+  // a self-sustaining loop.
+  //
+  // It stays hidden while the token only moves forward: `iat` has one-second
+  // resolution, so same-second reissues are byte-identical and React's `useState`
+  // bails on an `Object.is` match without re-rendering. Once anything makes the
+  // token alternate between two values, the loop runs at request speed. Each pass
+  // also calls `setUser()` with a freshly parsed object and toggles `isLoading`,
+  // which tears down and reopens the chat WebSocket, so the visible symptom is a
+  // reconnect storm rather than the request flood underneath it.
   useEffect(() => {
     if (IS_PLATFORM) {
       setUser({ username: 'platform-user' });
--- a/src/modules/auth/tests/authVerificationReentrancy.test.tsx
+++ b/src/modules/auth/tests/authVerificationReentrancy.test.tsx
@@ -1,0 +1,146 @@
+import assert from 'node:assert/strict';
+
+import { act, cleanup, render } from '@testing-library/react';
+import { afterEach, beforeEach, test, vi } from 'vitest';
+
+import { AuthProvider } from '@/modules/auth/context/AuthContext';
+import { storeAuthToken } from '@/shared/authToken';
+
+/**
+ * Regression coverage for the auth re-entrancy loop.
+ *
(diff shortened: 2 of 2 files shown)


TadMSTR:
Field data, in case it's useful — this reproduced on a self-hosted instance today, which speaks to the "not reproducible on latest main" question from #1177.

Two episodes in a single day's server log, either side of a version bump:

| Window | Cycles | Rate | Version |
|---|---|---|---|
| 04:54–05:05 | ~5,300 | ~800/min | v1.37.2 |
| 10:31–10:46 | **11,542** | peak 1,167/min, **48 within one second** | v1.37.3 |

Each cycle is the same four lines repeating:

```
WebSocket connection attempt to: /ws?token=REDACTED
[OK] WebSocket authenticated for user: <user>
[INFO] Chat WebSocket connected
[INFO] Chat client disconnected
```

Both start immediately after a service restart and decay without ever stopping — the second ended only because the service was restarted again. Since then, 2 connect attempts total. The server-side error log is clean throughout; nothing fails, the socket is just torn down and reopened.

**Disclosure so the numbers are read correctly:** this instance runs v1.37.3 plus five local patches. All five are server-side (a CLI-argv secret scrub, plugin env passthrough, the plugin WS proxy's `Origin` header, and two build scripts). I checked before posting — `AuthContext.tsx`, `WebSocketContext.tsx`, `api.ts`, `authToken.ts` and `auth.middleware.ts` are all **byte-identical to the v1.37.3 tag**, so the code path above is stock.

**On the rate, which I think is the interesting part.** My first assumption was the `onclose` reconnect, but that path retries on a fixed 3000ms timer (`WebSocketContext.tsx:113-116`) and caps at 0.33 cycles/sec — 11,542 cycles that way would take about 9.6 hours, not 15 minutes.

The amplifier looks like `WebSocketContext.tsx:126`, where `connect`'s dependency array is `[dispatch, isAuthLoading, token, user]`, and the effect at `:159` depends on `connect`. So a change to `token` **or to `user`'s object identity** reopens the socket immediately, with no delay at all. The loop in this PR supplies both on every pass — `setToken()` from the refreshed header, and `setUser(userPayload.user)` with a freshly parsed object, which is a new identity every time. That's render speed rather than timer speed, which matches 48 connects inside one second.

I've deliberately kept this PR to removing the *trigger* and left that dependency array alone — narrowing it is a separate judgement call and I didn't want to widen the diff. Happy to open it as a follow-up if you think it's worth doing.

Worth noting the earlier episode predates the version bump, so this isn't a v1.37.3 regression — it's just that a restart is what makes it visible.

PR #1314, opened 2026-09-11
fix(server): allow workspace paths under a Windows drive root WORKSPACES_ROOT

### LINKED ISSUE #746: 在 Windows 上，当设置环境变量 `WORKSPACES_ROOT=D:\` 或 `D:` 时，尝试添加该驱动器下的项目路径（如 `D:\Claude code`）会报错：  Workspace path must be within the allowed workspace root: D:\
## 问题描述

  在 Windows 上，当设置环境变量 `WORKSPACES_ROOT=D:\` 或 `D:` 时，尝试添加该驱动器下的项目路径（如 `D:\Claude code`）会报错：

  Workspace path must be within the allowed workspace root: D:\

  ## 原因分析

  问题出在 `server/shared/utils.js` 的 `validateWorkspacePath` 函数中：

  ```javascript
  const resolvedWorkspaceRoot = normalizeProjectPath(await realpath(WORKSPACES_ROOT));
  if (!resolvedPath.startsWith(`${resolvedWorkspaceRoot}${path.sep}`)
      && resolvedPath !== resolvedWorkspaceRoot) {
      // 报错
  }

  在 Windows 上，realpath("D:") 或 realpath("D:\") 可能会失败或返回意外的结果，导致后续的路径比较逻辑出错。

  修复建议

  在 dist-server/server/shared/utils.js 第 201-208 行，增加对 Windows 驱动器根目录的特殊处理：

  // Fix for Windows drive root paths (D: or D:\)
  let workspaceRootForCheck = WORKSPACES_ROOT;
  try {
      const resolvedWorkspaceRoot = normalizeProjectPath(await realpath(WORKSPACES_ROOT));
      workspaceRootForCheck = resolvedWorkspaceRoot;
  } catch (e) {
      // realpath failed, use WORKSPACES_ROOT directly
  }

  // Windows drive root check: if workspace root is like "D:" or "D:\"
  const isWindowsDriveRoot = /^[A-Za-z]:\\?$/.test(workspaceRootForCheck);
  const pathStartsWithRoot = isWindowsDriveRoot
      ? (resolvedPath.match(/^[A-Za-z]:/)?.[0]?.toLowerCase() === workspaceRootForCheck.match(/^[A-Za-z]:/)?.[0]?.toLowerCase())
      : (resolvedPath.startsWith(`${workspaceRootForCheck}${path.sep}`) || resolvedPath === workspaceRootForCheck);

  if (!pathStartsWithRoot) {
      return {
          valid: false,
          error: `Workspace path must be within the allowed workspace root: ${WORKSPACES_ROOT}`,
      };
  }

  环境信息

  - OS: Windows 11
  - Node.js: v20.x
  - cloudcli 版本: 最新版

  ---

leehom0123:
## Problem

Fixes #746

On Windows, setting `WORKSPACES_ROOT=D:` (or `D:`) made `validateWorkspacePath` reject every path on that drive with:

> Workspace path must be within the allowed workspace root: D:
Two root causes:

1. `normalizeProjectPath` keeps the trailing separator for filesystem roots (`normalizeProjectPath('D:\') === 'D:\'`), so the containment check built the prefix `"D:\\\\"` (double separator) and `resolvedPath.startsWith(root + path.sep)` never matched paths like `D:\Claude code`.
2. `await realpath(WORKSPACES_ROOT)` was unguarded — `realpath` on a drive root can fail, aborting validation entirely.

## Fix

Added `isPathWithinRoot(candidatePath, normalizedRoot)` and used it for both the workspace-root containment check and the symlink containment check:

- Recognizes Windows drive roots (`D:` / `D:.` / `D:\`) and compares drive letters instead of doing a string prefix check.
- Compares Windows paths case-insensitively (drive letters and path segments).
- Keeps the original POSIX containment semantics unchanged (no partial-segment matches, e.g. `/home/user` vs `/home/username`).
- `realpath` on a not-yet-existing `WORKSPACES_ROOT` now falls back to the normalized absolute path instead of throwing.

## Verification

- New unit tests in `server/shared/tests/workspace-path.test.ts` (drive-root accept, cross-drive reject, case-insensitivity, partial-segment, POSIX behavior): 7/7 pass, full `server/shared` suite green.
- `tsc --noEmit` passes.
- End-to-end on Windows with `WORKSPACES_ROOT=D:`: `D:\Claude code`, `D:\Claude code\sub`, `D:\` all validate; with `WORKSPACES_ROOT=D:\workspaces`, `D:\Claude code` and `C:\Users` are still correctly rejected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workspace path validation across Windows and POSIX environments.
  * Correctly handles drive roots, drive-letter casing, path boundaries, symlink targets, and workspaces that do not yet exist.
  * Prevents incorrectly accepting paths from different drives or with only partially matching directory names.

* **Tests**
  * Added coverage for Windows drive-root behavior, case-insensitive comparisons, path containment, and POSIX compatibility.

### DIFF
--- a/server/shared/tests/workspace-path.test.ts
+++ b/server/shared/tests/workspace-path.test.ts
@@ -1,0 +1,44 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+
+import { isPathWithinRoot, normalizeProjectPath } from '@/shared/utils.js';
+
+test('isPathWithinRoot accepts paths under a Windows drive root (issue #746)', () => {
+  const root = normalizeProjectPath('D:\\');
+  assert.equal(root, 'D:\\');
+  assert.equal(isPathWithinRoot(normalizeProjectPath('D:\\Claude code'), root), true);
+  assert.equal(isPathWithinRoot(normalizeProjectPath('D:\\work\\project'), root), true);
+});
+
+test('isPathWithinRoot also accepts the bare drive form for a drive root', () => {
+  const root = normalizeProjectPath('D:');
+  assert.equal(isPathWithinRoot(normalizeProjectPath('D:\\Claude code'), root), true);
+});
+
+test('isPathWithinRoot matches the drive root itself', () => {
+  const root = normalizeProjectPath('D:\\');
+  assert.equal(isPathWithinRoot('D:\\', root), true);
+});
+
+test('isPathWithinRoot rejects paths on a different drive', () => {
+  const root = normalizeProjectPath('D:\\');
+  assert.equal(isPathWithinRoot(normalizeProjectPath('C:\\Users\\someone'), root), false);
+});
+
+test('isPathWithinRoot compares Windows paths case-insensitively', () => {
+  const root = normalizeProjectPath('C:\\Users\\lihong');
+  assert.equal(isPathWithinRoot(normalizeProjectPath('c:\\users\\LIHONG\\proj'), root), true);
+});
+
+test('isPathWithinRoot does not match on partial path segments', () => {
+  const root = normalizeProjectPath('C:\\Users');
+  assert.equal(isPathWithinRoot(normalizeProjectPath('C:\\UsersOther'), root), false);
+});
+
+test('isPathWithinRoot keeps POSIX containment behavior', () => {
+  const root = normalizeProjectPath('/home/user');
+  assert.equal(isPathWithinRoot(normalizeProjectPath('/home/user/project'), root), true);
+  assert.equal(isPathWithinRoot(normalizeProjectPath('/home/user'), root), true);
+  assert.equal(isPathWithinRoot(normalizeProjectPath('/etc/passwd'), root), false);
+  assert.equal(isPathWithinRoot(normalizeProjectPath('/home/username/project'), root), false);
+});
--- a/server/shared/utils.ts
+++ b/server/shared/utils.ts
@@ -213,6 +213,44 @@ export function normalizeProjectPath(inputPath: string): string {
 }
 
 /**
+ * Checks whether an already-normalized path equals or is nested inside a
+ * normalized workspace root.
+ *
+ * The naive `startsWith(root + sep)` comparison breaks for Windows drive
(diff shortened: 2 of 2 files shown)


MacLeod92:
This appears to be a duplicate of #1111 (opened 2026-08-05), which fixes the same root-path containment bug with `path.relative` and includes regression tests.

Thanks for linking #746. I had not seen it, and I have added it to the #1111 description.

leehom0123:
@MacLeod92 agreed — closing this in favor of yours.

I verified your `path.relative` approach against the scenario from #746 before closing: with `win32` rules a `D:\` root accepts its descendants, rejects other drives, matches case-insensitively, and doesn't match partial segments (`C:\UsersOther` under a `C:\Users` root) — all correct. Your approach is also cleaner than our prefix comparison and additionally covers the POSIX filesystem root `/`, which ours did not, so containerized `WORKSPACES_ROOT=/` setups come out working too. Thanks for picking that up and for linking #746 into the description.

Two edges we had covered here that yours doesn't, in case they're worth folding in — take them or leave them:

1. **Missing `WORKSPACES_ROOT`.** If the configured root itself does not exist yet (e.g. a freshly created drive root), `await realpath(WORKSPACES_ROOT)` throws `ENOENT` and the validation request fails with a 500 instead of a clean `{ valid: false }`. We wrapped it and fell back to the normalized `path.resolve(WORKSPACES_ROOT)` form.
2. **Bare drive-letter root form.** If `WORKSPACES_ROOT` is ever configured as `D:` (no separator), `path.win32.relative` resolves it drive-relative against the cwd on that drive and the containment check wrongly rejects valid paths (`relative('D:', 'D:\\Claude code')` → `'..\\..\\..\\..\\Claude code'`). Not reachable via the `realpath` output, but a `D:` → `D:\` normalization would make the check robust to that config shape.

Our branch (`fix/windows-drive-root-workspace-root`) also carries 8 `node:test` cases for the Windows rules if any are useful. Good luck getting #1111 merged.

PR #1316, opened 2026-09-11
feat(providers): source claude, codex, and opencode model catalogs from the CLIs

leehom0123:
## Problem

The model pickers never showed models from the user's own custom providers, and each provider shipped a hardcoded catalog that drifted from what its CLI can actually run:

- **OpenCode**: the adapter never asked the CLI for a catalog. A curated 54-model list covered only the built-in gateways (`opencode`, `opencode-go`, `anthropic`, `openai`), narrowed by probing `auth.json` / user config / env keys - subtraction only, so a user-defined provider could never contribute a model. Worse, when the CLI was unreachable the picker silently served that curated list, which looked identical to "the fix not working".
- **Claude / Codex**: same pattern - static catalogs instead of what `claude` / `codex` actually support.

## Fix

All three provider model adapters now read their catalog from the provider CLI - the same resolver `opencode run --model`, `claude`, and `codex` use at runtime:

- **OpenCode** parses `opencode models --verbose`: `providerID/modelID` + per-model JSON records give real ids, display names, and reasoning variants. CLI `variants` keys map onto the effort picker and are forwarded to the runtime as `--variant`. Entries whose `status` is not `active` are dropped. The hardcoded catalog is **gone**: when the CLI cannot answer (not installed, spawn error, timeout, unparseable), the picker is empty and the server logs `[OpenCode] `opencode models --verbose` unusable (...)` with the reason - no silently-stale list masquerading as a catalog.
- **Claude / Codex** source listings from the live CLI catalog, including supported effort levels and defaults, excluding hidden/retired models.
- Spawned via `cross-spawn`, matching the OpenCode runtime adapter's Windows `.cmd` handling.
- Each live catalog is cached briefly (60s) with a shared in-flight run, so picker + active-model lookup + effort validation cost one CLI invocation per page load; failed runs are not cached so the next caller retries.

## Tests

- OpenCode: 6 tests via a `runModelsCli` injection seam - user-defined providers reach the picker, CLI labels/variants win, retired entries drop out, default migrates to a live model, concurrent lookups share one CLI run, and an unusable CLI yields an empty catalog.
- Claude / Codex suites cover the CLI-sourced catalogs likewise.
- `tsc --noEmit` and oxlint: clean.

### DIFF
--- a/server/modules/providers/list/claude/claude-models.provider.ts
+++ b/server/modules/providers/list/claude/claude-models.provider.ts
@@ -1,5 +1,7 @@
 import { readFile } from 'node:fs/promises';
 
+import { query } from '@anthropic-ai/claude-agent-sdk';
+
 import { sessionsDb } from '@/modules/database/index.js';
 import type { IProviderModels } from '@/shared/interfaces.js';
 import type {
@@ -7,7 +9,7 @@ import type {
   ProviderModelOption,
   ProviderModelsDefinition,
 } from '@/shared/types.js';
-import { buildDefaultProviderCurrentActiveModel } from '@/shared/utils.js';
+import { AppError, buildDefaultProviderCurrentActiveModel } from '@/shared/utils.js';
 
 /**
  * Ultracode is not one of the SDK's reasoning-effort levels. Selecting it runs the turn at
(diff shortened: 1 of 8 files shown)


blackmammoth:
@leehom0123, please resolve the merge conflicts.

leehom0123:
Conflicts resolved — merged the latest main into the branch.

Two files conflicted:
- `claude-models.provider.ts`: import-level only; kept both sides (this PR's `AppError` plus main's `stripAnsiSequences`).
- `opencode-models.test.ts`: the real one. Main's two predefined-catalog tests (`exposes only the curated predefined catalog`, `offers only models the install can route to`) assert the hardcoded `OPENCODE_PREDEFINED_MODELS` catalog that this PR intentionally removes — the CLI is the single source now, so they can't survive the merge; the six CLI-sourced catalog tests cover the new behavior. Main's session model-id composition test from #1333 tests behavior that survives this PR, so I kept it and adapted it to this branch's `withOpenCodeHome` helper.

All provider model tests, typecheck, and lint pass locally.

PR #1317, opened 2026-09-11
feat(providers): restore Codex fork anchors and add OpenCode session forking

leehom0123:
## Problem

"Fork session from a specific message" did not work for two of the providers that advertised support for it:

- **Codex** — recent CLI releases (0.14x+) no longer write user prompts as `event_msg/user_message` rows; prompts now arrive as `event_msg/item_completed` with `item.type === 'UserMessage'`. The history reader had no branch for the new shape, so reopened Codex sessions rendered **no user bubbles at all**, and without the per-message `transcriptAnchorId` the edit/fork buttons never appeared.
- **OpenCode** — there was no fork adapter at all (`supportsSessionForking: false`).

## Approach

**Codex**: teach the reader the new rollout shape. The new branch shares the "one anchor per turn" rule with the legacy `user_message` path, so anchoring stays consistent; rolled-back turns keep their old rows-but-no-anchor behavior. No changes to the fork adapter / app-server client — `thread/fork`'s inclusive `lastTurnId` semantics are format-independent.

**OpenCode**: fork via the CLI's own `opencode serve` HTTP API (`POST /session/{id}/fork`) rather than writing `opencode.db` directly (the CLI owns its WAL; this app only ever opens it read-only). A short-lived, randomly-passworded server subprocess performs the copy — same arrangement as the existing Codex app-server client. The endpoint's cut is **exclusive** of the message id it names (verified against a live v1.18 server), while the fork contract's anchor is inclusive, so the adapter translates the anchor into the following message's id (read from `opencode.db` with the same `ORDER BY` history reads use). A whole-session fork from the sidebar works by simply omitting the cut point.

The `IProviderFork` contract gains an optional `transcriptIsSharedDatabase` flag so the sessions service accepts a source row whose `jsonl_path` is legitimately `NULL`, and `createForkedSession` persists a `NULL` path — which is what keeps deleting a forked app row from touching `opencode.db`.

**Drive-by fix** (separate commit): a failed login crashed the entire login screen. The server's `AppError` responses carry `error` as an object (`{ code, message }`) while `resolveApiErrorMessage` assumed a string; React threw *"Objects are not valid as a React child"* and unmounted everything — one mistyped password left a blank page. The resolver now only passes strings through.

## Testing

- New `opencode-session-fork.test.ts`: anchor → exclusive cut-point translation, whole-session fork, missing-anchor rejection, shared-store return shape.
- `session-fork.test.ts`: shared-database provider forks a `NULL`-path session; file-based providers are still refused without a transcript.
- `codex-message-editing.test.ts`: new-format fixtures parameterized alongside legacy ones (anchoring, one-anchor-per-turn, rollback behavior).
- `opencode-sessions.test.ts`: user prompts carry their native `msg_*` row id as the anchor.
- `npm test` (430 tests) and `npm run typecheck` pass; the 9 pre-existing failures in `claude-auth` / `projects.db.integration` reproduce on a clean `main` checkout (environment-dependent, unrelated).
- Manual: forked Codex and OpenCode sessions resume correctly; deleting a forked OpenCode row leaves `opencode.db` untouched.

No frontend changes — the buttons were always capability-driven.

## Summary by CodeRabbit

- **New Features**
  - Added session forking support for OpenCode sessions, including forks from a selected message or the complete session.
  - OpenCode sessions can now be forked when transcripts are stored in a shared database.

- **Improvements**
  - Improved Codex support for newer transcript formats, including text and image prompts and inherited fork history.
  - Forked conversations now exclude the selected message, with clearer localized labels.
  - Preserved reliable message anchors for editing and branching.

- **Bug Fixes**
  - Structured API errors now display readable messages instead of causing rendering failures.

### DIFF
--- a/server/modules/database/repositories/sessions.db.ts
+++ b/server/modules/database/repositories/sessions.db.ts
@@ -200,7 +200,10 @@ export const sessionsDb = {
    * Unlike `createAppSession` this writes `provider_session_id` and
    * `jsonl_path` immediately, because a fork's transcript file exists before
    * the row does — and the filesystem watcher would otherwise index it as an
-   * unrelated session under its own id.
+   * unrelated session under its own id. `jsonlPath` is null for providers that
+   * keep transcripts in a shared database (OpenCode), where the copy exists
+   * without any per-session file; keeping it null there is also what stops
+   * deleting one app session from deleting everybody's store.
    */
   createForkedSession(input: {
     sessionId: string;
@@ -208,7 +211,7 @@ export const sessionsDb = {
     projectPath: string;
     customName: string | null;
     providerSessionId: string;
-    jsonlPath: string;
+    jsonlPath: string | null;
     forkedFromSessionId: string;
     model: string | null;
     effort: string | null;
--- a/server/modules/providers/list/claude/claude-fork.provider.ts
+++ b/server/modules/providers/list/claude/claude-fork.provider.ts
@@ -1,4 +1,6 @@
 import { stat } from 'node:fs/promises';
+import { createReadStream } from 'node:fs';
+import * as readline from 'node:readline';
 import path from 'node:path';
 
 import { forkSession as forkClaudeSession } from '@anthropic-ai/claude-agent-sdk';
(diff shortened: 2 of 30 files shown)


blackmammoth:
hey @leehom0123, I really like the PR...please fix the merge conflicts and will review again.

leehom0123:
Thanks @blackmammoth! Merged the latest main into the branch and resolved the conflict — main's new "Codex history restores user prompts from typed item_completed rows" test and this PR's paginated-fork tests had both landed at the same spot at the end of `codex-sessions.test.ts`, so I kept both blocks. All server tests, typecheck, and lint pass locally. Ready for a re-review.

PR #1318, opened 2026-09-11
fix(codex): stop spawning console windows for Codex child processes

leehom0123:
## Problem

On Windows, Codex conversations flash a black console window on every turn (most visible when the server runs without a console, e.g. Electron or double-click launch; invisible when started from a terminal, which is why it reads as "random").

Node on Windows allocates a fresh console for spawned console-subsystem children unless `windowsHide: true` is passed. Two spawns were exposed:

1. **`codex app-server`** (`codex-app-server.client.ts`) — spawned by this app for `thread/fork` (editing/branching Codex conversations). Every fork flashed a cmd window.
2. **`codex.exe`** — `@openai/codex-sdk` spawns one `codex exec` per turn and passes no spawn flags through, so the SDK provides no way to set `windowsHide`. Verified upstream `@openai/codex-sdk@0.154.0` still omits it.

## Fix

1. Added `windowsHide: true` to the app-server spawn, matching what this repo already does for its other spawns (`claude-cli-path.ts`, `session-conversations-search.service.ts`).
2. Added `scripts/fix-codex-sdk-windows-hide.js`, chained into `postinstall` alongside the existing `fix-node-pty.js`, following the same in-place-patch pattern:
   - only runs on Windows;
   - anchors on the bundled spawn call in `@openai/codex-sdk/dist/index.js` and inserts `windowsHide: true`;
   - idempotent; skips with a console warning when the anchor disappears (SDK restructured, or upstream fixes it — the warning tells maintainers to remove the script);
   - failures warn but never break `npm install`.

## Verification

- Script patches the current install; second run is a no-op.
- After patching, `@openai/codex-sdk` spawns `codex.exe` with `windowsHide: true`; Codex turns no longer flash a console window.
- Codex provider tests: 36 tests, 35 pass / 1 pre-existing skip / 0 fail; `tsc --noEmit -p server/tsconfig.json` clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented an unwanted console window from briefly appearing when launching Codex on Windows.
  - Applied the same Windows display improvement during installation setup to ensure consistent behavior across environments.

- **Chores**
  - Updated installation steps to automatically apply the Windows compatibility adjustment when needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "start": "npm run build && npm run server",
     "release": "./release.sh",
     "prepublishOnly": "npm run build",
-    "postinstall": "node scripts/fix-node-pty.js",
+    "postinstall": "node scripts/fix-node-pty.js && node scripts/fix-codex-sdk-windows-hide.js",
     "prepare": "husky",
     "update:platform": "./update-platform.sh"
   },
--- a/scripts/fix-codex-sdk-windows-hide.js
+++ b/scripts/fix-codex-sdk-windows-hide.js
@@ -1,0 +1,78 @@
+#!/usr/bin/env node
+/**
+ * Hide the console window @openai/codex-sdk spawns codex.exe with
+ *
+ * The SDK runs one `codex exec` child process per turn through
+ * child_process.spawn without passing windowsHide: true. On Windows that
+ * makes the OS allocate a fresh console for the console-subsystem codex.exe,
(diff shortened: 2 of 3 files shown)


leehom0123:
Closing: reproduced the console window under PM2 — the cause is how PM2 launches the server without a console, not the missing windowsHide flags. Retracting this PR.

PR #1321, opened 2026-09-12
feat(chat): stream provider thinking token-by-token across claude, codex, and opencode

leehom0123:
## What

Turn-by-turn "one lump" thinking/text output becomes true token-by-token streaming for all three providers.

### Root causes fixed

- **Claude**: the SDK was never asked for partial messages, so thinking and text each arrived as one lump per turn. Now `includePartialMessages` is enabled and `stream_event` frames are normalized (`text_delta` -> `stream_delta`, `thinking_delta` -> `thinking_delta`).
- **OpenCode**: the live normalizer read the wrong field shape (payloads live nested under `part`) and spawned the CLI without `--thinking`, so every reasoning/text/tool event was silently dropped until the end-of-run history reload. Fixed against a live v1.18.30 capture.
- **Codex**: live reasoning events lost their stable `itemId` in normalization and were mislabeled as ordinary text, stacking one row per SDK tick. Now `item.updated` ticks are forwarded with the `itemId` kept as the row id, and the nested `message.isReasoning` flag is honored so reasoning normalizes as thinking instead of prose.

### Client

- Accumulated thinking stream (`thinking_delta` + well-known row) alongside the existing text stream.
- Same-id in-place realtime updates; thinking dedupe against the persisted transcript.
- A streaming Reasoning bubble.

## Tests

- `thinkingStreaming.test.tsx` — 5 cases (accumulated thinking row, dedupe, bubble rendering).
- `claude-sessions.test.ts` / `codex-sessions.test.ts` / `opencode-sessions.test.ts` / `opencode-runtime.provider.test.js` — 36 cases covering the new normalization paths.
- All 41 green on this branch (rebased onto current `main`).

## Summary by CodeRabbit

* **New Features**
  * Reasoning content now streams live while responses are generated.
  * The active Reasoning panel opens automatically during streaming and collapses after completion.
  * Live reasoning updates remain separate from answer text and avoid duplicate entries.
  * Claude, Codex, and OpenCode sessions now display streaming reasoning, text, and tool activity more reliably.

* **Bug Fixes**
  * Improved restoration and display of reasoning content from conversation history.
  * Prevented duplicate reasoning and tool entries during live updates.

### DIFF
--- a/server/modules/providers/list/claude/claude-runtime.provider.js
+++ b/server/modules/providers/list/claude/claude-runtime.provider.js
@@ -286,6 +286,13 @@ function mapCliOptionsToSDK(options = {}) {
 
   sdkOptions.settingSources = ['project', 'user', 'local'];
 
+  // Token-level streaming: without this the SDK only yields whole assistant
+  // messages, so reply text and thinking appear one lump per turn. The
+  // partial `stream_event` frames feed the client's accumulated stream_delta /
+  // thinking_delta rendering; the complete assistant messages still arrive
+  // alongside and remain the persisted record.
+  sdkOptions.includePartialMessages = true;
+
   // The SDK resumes with the provider-native session id, never the app id.
   // `resumeFromScratch` is set when the very first prompt of a conversation was
   // edited: there is nothing before it to resume through, so the turn has to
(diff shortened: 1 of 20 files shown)


blackmammoth:
The streaming view doesn't look good UI/UX wise...so closing this for now. If the UI can be improved a lot, we can introduce it.

PR #1322, opened 2026-09-12
fix(chat): keep streaming follow-bottom honest, and the transcript inside the viewport

leehom0123:
## What

Two transcript layout fixes, plus a follow-bottom scroll fix found while testing them:

**1. Narrow screens clip the transcript (missing `min-w-0`)**
The chat column was a `flex-1` item without `min-w-0`, so its automatic minimum size let one wide message (a long code line, a wide table) inflate the whole transcript column past the viewport; the pane's overflow-x then clipped the right-hand content out of view. With the minimum width released, the column compresses to the viewport and wide content scrolls inside its own overflow box.

**2. The prompt navigator rail was invisible at narrow desktop widths**
The rail hung outside the message column (`left: 100%`), which falls past the pane's clipped edge once the pane shrinks below the column's max width plus the rail's gutter (e.g. a 1000px window with the sidebar open). It now docks inside the gutter the message column gives up, decided from the measured pane width — not the viewport — so split views and resizable panels track correctly too.

On desktop the classic scrollbar owns ~15px at the pane's right edge, so the message column, the export menu, and the rail all step back there to stay clear of it.

**3. Streaming follow-bottom snapped back over the user**
The follow-bottom timer reads the scrolled-up flag through a ref at fire time, but the ref was only synced from a state effect; during heavy streaming renders the state→effect round-trip lags past the 50ms timer and yanked the view back down right after the user dragged up. The ref is now mirrored where the flag is computed, and the timer's effect keys on the array identity instead of `.length` so it re-runs while a streaming row grows without changing the message count.

## Tested

Via Chrome DevTools at 390px mobile, 1000px and 1280px desktop viewports: column/pane/scroll widths all collapse to the viewport with no overflowing descendants; the rail is visible and clear of the scrollbar at every width; `tsc --noEmit` and oxlint (via lint-staged) are clean.

## Summary by CodeRabbit

* **New Features**
  * Added a session information panel with context usage, turn statistics, tasks, subagents, context sources, and MCP server controls.
  * Added prompt navigation, subagent conversation viewing, and continuation support.
  * Model pickers now reflect live Claude, Codex, and OpenCode availability.
  * Added streamed reasoning output and tool-result image previews with zoomable lightboxes.
  * Added session forking support for OpenCode and improved anchor handling across providers.

* **Bug Fixes**
  * Improved API error messages, transcript rendering, image preservation, and large-transcript performance.
  * Added support for disabling individual MCP servers per user.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -147,3 +147,4 @@ tasks/
 cloudcli-sidebar-app-source.tar.gz
 cloudcli-sidebar.html
 electron/*.tar.gz
+.ref/
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -98,6 +98,7 @@
           "src/shared/voiceConfig.ts",
           "src/shared/codeEditorSettings.ts",
           "src/shared/userSettings.ts",
+          "src/shared/sessionInfoPanelPrefs.ts",
           "src/shared/chatDrafts.ts",
           "src/shared/syntaxHighlighter.ts"
         ],
--- a/CHECKLIST.md
+++ b/CHECKLIST.md
@@ -11,7 +11,7 @@
 - [ ] 新增抽象：`server/modules/providers/shared/session-insights/session-insights.provider.ts`
       定义 `SessionInsightsProvider` 基类（方法：`listSubagents`、`fetchSubagentHistory`、
       `getContextSnapshot`），各 CLI 目录各建一个实现文件；无能力的 CLI 返回空数组/`AppError`。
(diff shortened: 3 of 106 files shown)


blackmammoth:
hey @leehom0123, I like many of the changes but the PR is too huge to review. Please break it up into a lot of different pieces with a focused functionality for each and resubmit the PR. Make sure to resolve the coderabbit comments as well.

PR #1332, opened 2026-09-15
fix(chat): Enter that confirms an IME candidate must not send on Safari

isamu:
## Summary

Pressing Enter to confirm an IME candidate (Japanese, Chinese, Korean) sends the message on Safari, half-composed.

`src/modules/chat/hooks/useChatComposerState.ts` guards on `event.nativeEvent.isComposing`, which is the right idea but not sufficient: **Safari fires `compositionend` BEFORE the confirming Enter's `keydown`**, so the flag is already `false` by the time the handler runs. Chrome and Firefox fire it after — which is why the naive guard looks correct on those two and the bug reads as Safari-only.

This is a known WebKit ordering difference, not specific to this project:

- [Understanding Composition Browser Events](https://developer.squareup.com/blog/understanding-composition-browser-events/) — Square, on the ordering difference
- [makeplane/plane#8893](https://github.com/makeplane/plane/pull/8893) — the same fix in Plane
- [tiann/hapi#1481](https://github.com/tiann/hapi/issues/1481) — the same bug reported with a Safari repro, since fixed

## What the fix does

Tracks composition on `window` in the capture phase, and refuses an Enter that arrives within 30 ms of `compositionend`:

```ts
const endedJustNow =
  performance.now() - lastCompositionEndAtRef.current < SAFARI_IME_RACE_WINDOW_MS;
if (event.nativeEvent.isComposing || isComposingRef.current || endedJustNow) {
  return;
}
```

The `compositionend` → `keydown` sequence is synchronous (microseconds). A human pressing Enter a second time takes 100 ms or more, so a deliberate send never lands inside the window. The guard only ever becomes *more* conservative — on Chrome and Firefox `isComposing` is already `true` and nothing changes.

Listeners are on `window` in the capture phase rather than on the textarea, so a composition that ends after focus has moved still closes; `blur` does not bubble but a capture-phase listener still sees it.

## Items to confirm / review

Only the chat composer is changed. If other inputs in the app submit on Enter, they have the same issue and would want the same guard — I left them alone rather than widening the diff.

Disclosure: I maintain a tool in the same space and hit this bug there first; the fix is the same one.


viper151:
@isamu  are you able to provide screenshots or a video of the before and after perhaps?

isamu:
Here you go — Safari 26, macOS, Japanese IME. Typing こんにちは and pressing Enter
  to confirm the candidate. 

  Before:
  
https://github.com/user-attachments/assets/41c28e82-627e-41ca-849f-0b3c1137d8d5

  Pressing Enter to confirm the candidate sends the message, half-composed.
  
  After:
 
https://github.com/user-attachments/assets/5f6f1a66-71ee-43f1-b1be-41168e8c303e
 
  Confirming does nothing. A second Enter sends, as it should.
 
  The same keystrokes behave correctly in Chrome either way, which is why this only
  shows up on Safari and WKWebView: Chrome fires compositionend after the confirming
  keydown, so isComposing is still true there. Safari fires it before.

viper151 (approved):

PR #1337, opened 2026-09-16
fix(uploads): decode multipart filenames as UTF-8

notetook:
Uploading a file whose name contains non-ASCII characters (Korean, Japanese,
Chinese, accented Latin, ...) shows a garbled attachment name in the chat and
persists that garbled name in the session history.

Browsers send the multipart `filename` parameter as raw UTF-8 bytes, but
multer (via busboy) decodes it as latin1 unless `defParamCharset` is set, so
`file.originalname` arrives mojibaked:

```
expected: '개인정보 처리방침 안내서.pdf'
actual:   'ê°\x9Cì\x9D¸ì \x95ë³´ ì²\x98ë¦¬ë°©ì¹¨ ì\x95\x88ë\x82´ì\x84\x9C.pdf'
```

Set `defParamCharset: 'utf8'` on the chat image/file attachment uploaders
(`assets.routes.ts`) and on the file-tree uploader (`file-tree.module.ts`).
The on-disk filename is unaffected: it was already sanitised to ASCII.

## Reproduce

1. Attach a file named e.g. `안내서.pdf` (or `スクリーンショット.png`) in the chat composer and send.
2. Before: the attachment card in the transcript shows `ì\x95\x88ë\x82´ì\x84\x9c.pdf`.
   After: it shows `안내서.pdf`.

## Tests

Added `server/modules/assets/tests/assets.routes.test.ts`, which spins up the
assets router against a temporary home directory and pushes a Korean and a
Japanese filename through the real multer pipeline. Both cases fail on `main`
with the output above and pass with this change. `npm test`, `npm run
typecheck` and `npm run build` pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed uploaded file and image names containing non-ASCII characters so they are preserved correctly.
  - Improved filename handling for uploads through the assets and file tree features.
  - Preserved international characters in filenames when uploading images, files, and file tree content through multipart forms.


blackmammoth:
@notetook, tried reproducing this a couple times...both with the PR and not, I see this...so the issue is not addressed. 
image

PR #1338, opened 2026-09-16
fix(chat): dispatch a queued message when the run it waited on settles

### LINKED ISSUE #1331: Queued messages wait up to 30s to start — the dispatcher only polls, nothing fires on run completion
## What happens

Queue a message while a turn is running. The running turn finishes, the session
goes idle, and then nothing happens for up to 30 seconds before the queued turn
starts. The reply also arrives in one lump rather than streaming, because a
queued turn runs detached.

## Why

`server/modules/scheduled-messages/services/scheduled-message-dispatcher.service.ts`
is the only sender, and its only trigger is a timer:

```ts
const POLL_INTERVAL_MS = 30_000;
...
pollTimer = setInterval(poll, POLL_INTERVAL_MS);
```

`dispatchQueuedMessages()` skips any session where
`chatRunRegistry.isProcessing(sessionId)` is true, so a message queued during a
turn is necessarily skipped by every tick until the turn ends — and then waits
for the *next* tick, 0–30s later.

The frontend cannot compensate, by design. `useChatComposerState.ts`:

```ts
// The VPS dispatcher owns sending. While the card is visible, periodically
// reconcile only its removal so the UI notices when the server claims it.
```

so the client polls every 5s purely to notice removal, and never sends.

Nothing subscribes to run completion. `chat-run-registry.service.ts` flips
`run.status = 'completed'` in one place and emits no event, so the information
"this session just went idle" exists and is discarded.

## Suggested fix

Emit a run-settled event from the registry where the status flips, and have the
dispatcher subscribe, keeping the poll as the safety net for messages queued
while nothing was running and for restarts. Small diff, no behaviour change for
scheduled messages. Happy to open a PR — I have one running locally.

## Notes

- Related but different: #1225 (open) — queuing a second message overwrites the
  first, so only the last survives.
- Related: #1142 — steering was never implemented; queueing sends a **new run
  after** the current turn, which is also why the reply does not stream.

happy-proger:
## What

A message queued while a turn is running waits up to 30s after the session goes idle before it is sent.

## Why

`scheduled-message-dispatcher.service.ts` is the only sender, and its only trigger is a timer:

```ts
const POLL_INTERVAL_MS = 30_000;
...
pollTimer = setInterval(poll, POLL_INTERVAL_MS);
```

`dispatchQueuedMessages()` skips any session where `chatRunRegistry.isProcessing(sessionId)` is true, so a message queued during a turn is necessarily skipped by every tick until the turn ends -- and then waits for the *next* tick, 0-30s later.

The frontend cannot compensate, by design (`useChatComposerState.ts`):

```ts
// The VPS dispatcher owns sending. While the card is visible, periodically
// reconcile only its removal so the UI notices when the server claims it.
```

so the client polls every 5s purely to notice removal, and never sends.

Nothing subscribed to run completion: `chat-run-registry.service.ts` flips `run.status = 'completed'` in one place and emits no event, so "this session just went idle" existed and was discarded.

## The change

- `chat-run-registry` exposes `onRunSettled()` and fires it where the status flips, on a microtask so a throwing subscriber cannot break the event that settled the run.
- The dispatcher subscribes at init and unsubscribes on close.
- The 30s poll stays as the safety net for messages queued while nothing was running, and for server restarts.
- `dispatchQueuedMessages()` already re-checks `isProcessing()` and claims transactionally, so the event racing the poll is harmless.

Scheduled-message behaviour is unchanged.

## Measured

Before: 0-30s of dead air after the turn ended. After, from `session_drafts` on a live session:

```
18:04:27  turn ends (session goes idle)
18:04:28  queued_message -> NULL   (claimed and dispatched)
```

~1s, the rest of the time to first token being process spin-up.

## Notes

- Test suite: 415/417 pass; the one failure (`chat-edit-send.test.ts > an edit resumes through the turn before the one being replaced`) also fails on an untouched checkout and passes in isolation.
- The issue body also mentions the reply arriving in one lump rather than streaming. That is a separate and wider defect -- `includePartialMessages` is never set when building SDK options, so the Claude provider emits no `content_block_delta` events for *any* turn, queued or typed, and the `stream_delta` handling in `claude-sessions.provider.ts` is unreachable. Happy to open that as its own PR.

Fixes #1331

## Summary by CodeRabbit

- **New Features**
  - Queued scheduled messages are now dispatched immediately when an active chat run finishes, reducing delays between turns.
  - Added event-driven notifications when chat runs reach completion.
  - The existing polling process remains available as a fallback to ensure queued messages are still delivered.


blackmammoth:
hey @happy-proger, please address the coderabbit feedback if they are relevant

PR #1436, opened 2026-09-24
fix(windows): resolve the claude CLI by walking PATH instead of where.exe

xingcheng1061:
## Problem

On Windows the CLI lookup shelled out to `where.exe` and decoded its output as UTF-8. `where.exe` writes paths in the console code page, so on a host with a non-ASCII user directory (for example `C:\Users\星辰` on a GBK system) the result decoded into mojibake that no longer existed on disk. The SDK then failed with:

```
Claude Code native binary not found at <mangled path>
```

...even though the CLI was installed and on PATH. Reproduced live on a GBK host: `where.exe` emits `D0C7 B3BD` for the username, which UTF-8 decoding turns into replacement characters.

## Fix

Replace the subprocess with a pure filesystem walk that mirrors Windows lookup semantics:

- PATH **directory order** first, **PATHEXT extension order** within a directory
- duplicate PATH entries dropped (case-insensitive) so a repeated entry cannot shadow a later install
- every hit verified with `existsSync` so a stale PATH entry can never produce a phantom path the SDK cannot spawn
- a native `.exe` hit still wins over npm shim scripts, and shim scripts (`claude`, `claude.cmd`) are still resolved to their wrapped `claude.exe` via the existing wrapper parser

No process is spawned and no code page is consulted, so non-ASCII install directories resolve correctly. This also removes one subprocess spawn per lookup.

`PATHEXT` falls back to the cmd.exe default list when the host does not export it, keeping the lookup equivalent to what `where.exe` reported.

## Testing

- `server/shared/tests/claude-cli-path.test.ts` — 11 tests, including a regression case with a non-ASCII PATH install directory and cases for stale entries, shim-vs-exe preference, and explicit-path passthrough — all pass
- `npx tsc --noEmit -p server/tsconfig.json` — clean
- `oxlint` — 0 warnings, 0 errors

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Claude Code detection on Windows when installed in PATH directories, including directories with non-ASCII characters.
  * The app now skips PATH locations where the executable is missing and prefers a native Windows executable over an npm wrapper when both are available.
  * Windows executable lookup now respects configured extension settings and continues to support the system’s default extensions when none are configured.


xingcheng1061:
Good catch — pushed `4ed211cf`.

The search now prepends the current working directory (the first place Windows command resolution looks, and behavior the previous `where.exe` lookup inherited), deduplicated against PATH entries. Two new tests cover it: a `claude.exe` that only exists in the working directory is found, and a working directory that also appears on PATH is not scanned twice. A deleted working directory now degrades to a PATH-only search instead of throwing from `process.cwd()`.

13/13 tests, typecheck and lint clean.

